PR #24, opened 2014-05-06
Moving Average Filter for better vi timing

ricrpi:
On embedded systems it is likely that the time to do a VI update will exceed the 'AdjustedLimit' time.

The use of frame skipping for video plug-ins can counter this but if l_MainSpeedLimit is true then mupen64plus will pause during a VI where graphics are not being updated and so emulation cannot 'catch up'. 

I have found that implementing a 6 point Moving Average filter gets around this nicely so that emulation will not over-run when the processing load is low but will also not waste CPU cycles when the load is high.

### DIFF
--- a/src/r4300/interupt.c
+++ b/src/r4300/interupt.c
@@ -108,6 +108,7 @@ static void queue_free(interupt_queue *qToFree)
 
 static void clear_queue(void)
 {
+    int i;
     while(q != NULL)
     {
         interupt_queue *aux = q->next;


littleguy77:
@richard42 Did this pull request end up getting merged anyhow?  See PR #26.

Android testers have discovered an audio delay that was introduced some time in 2014.  @Gillou68310  mentioned he tracked it down to a change in the framelimiter implementation.  He can probably elaborate further but at first blush it looks like something may have been merged accidentally between 24 and 26.

PR #57, opened 2015-01-07
Allow building the core with the new dynarec under MSVC2010

Gillou68310:
Here's some instructions In order to build with the new_dynarec:
-Open mupen64plus-ui-console solution property page
-Select "Configuration Properties"
-Change mupen64plus-core project configuration to New_Dynarec_Debug or New_Dynarec_Release depending if you're building in Debug or Release mode.

Dependencies with other projects:
https://github.com/mupen64plus/mupen64plus-ui-console/pull/11
https://github.com/mupen64plus/mupen64plus-video-glide64mk2/pull/43
https://github.com/mupen64plus/mupen64plus-video-rice/pull/39
https://github.com/mupen64plus/mupen64plus-rsp-hle/pull/31
https://github.com/mupen64plus/mupen64plus-input-sdl/pull/14
https://github.com/mupen64plus/mupen64plus-audio-sdl/pull/7
https://github.com/mupen64plus/mupen64plus-win32-deps/pull/1

### DIFF
--- a/projects/msvc10/mupen64plus-core.vcxproj
+++ b/projects/msvc10/mupen64plus-core.vcxproj
@@ -1,0 +1,357 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Debug|Win32">
+      <Configuration>Debug</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="New_Dynarec_Debug|Win32">
+      <Configuration>New_Dynarec_Debug</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="New_Dynarec_Release|Win32">
+      <Configuration>New_Dynarec_Release</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|Win32">
+      <Configuration>Release</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>{92D3FEB9-2129-41C5-8577-BCD7D961EF41}</ProjectGuid>
+    <RootNamespace>mupen64pluscore</RootNamespace>
+    <Keyword>Win32Proj</Keyword>
+  </PropertyGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="Configuration">
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
+    <CharacterSet>MultiByte</CharacterSet>
+    <WholeProgramOptimization>true</WholeProgramOptimization>
+    <PlatformToolset>v100</PlatformToolset>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Release|Win32'" Label="Configuration">
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
+    <CharacterSet>MultiByte</CharacterSet>
+    <WholeProgramOptimization>true</WholeProgramOptimization>
+    <PlatformToolset>v100</PlatformToolset>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
+    <CharacterSet>MultiByte</CharacterSet>
+    <PlatformToolset>v100</PlatformToolset>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Debug|Win32'" Label="Configuration">
+    <ConfigurationType>DynamicLibrary</ConfigurationType>
+    <CharacterSet>MultiByte</CharacterSet>
+    <PlatformToolset>v100</PlatformToolset>
+  </PropertyGroup>
+  <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
+  <ImportGroup Label="ExtensionSettings">
+  </ImportGroup>
+  <ImportGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'" Label="PropertySheets">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <ImportGroup Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Release|Win32'" Label="PropertySheets">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <ImportGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="PropertySheets">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <ImportGroup Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Debug|Win32'" Label="PropertySheets">
+    <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
+  </ImportGroup>
+  <PropertyGroup Label="UserMacros" />
+  <PropertyGroup>
+    <_ProjectFileVersion>10.0.40219.1</_ProjectFileVersion>
+    <OutDir Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">$(SolutionDir)$(Configuration)\</OutDir>
+    <OutDir Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Debug|Win32'">$(SolutionDir)Debug\</OutDir>
+    <IntDir Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">$(Configuration)\</IntDir>
+    <IntDir Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Debug|Win32'">Debug\</IntDir>
+    <LinkIncremental Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">true</LinkIncremental>
+    <LinkIncremental Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Debug|Win32'">true</LinkIncremental>
+    <OutDir Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">$(SolutionDir)$(Configuration)\</OutDir>
+    <OutDir Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Release|Win32'">$(SolutionDir)Release\</OutDir>
+    <IntDir Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">$(Configuration)\</IntDir>
+    <IntDir Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Release|Win32'">Release\</IntDir>
+    <LinkIncremental Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">false</LinkIncremental>
+    <LinkIncremental Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Release|Win32'">false</LinkIncremental>
+    <CodeAnalysisRuleSet Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">AllRules.ruleset</CodeAnalysisRuleSet>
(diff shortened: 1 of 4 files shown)


richard42:
@Gillou68310, can you set the MSVC project file to build the new_dynarec code in C++ mode instead of moving all the variable declarations to the top?  It is a much better coding style to declare variables where they are needed, and really ridiculous that Microsoft didn't support C99 until VC2013.

Gillou68310:
> @Gillou68310, can you set the MSVC project file to build the new_dynarec code in C++ mode instead of moving all the variable declarations to the top?

Sure this is definitely a much better solution. I will close this PR and reopen it when I'm done, 

> It is a much better coding style to declare variables where they are needed, and really ridiculous that Microsoft didn't support C99 until VC2013.

Couldn't agree more ;-)

bsmiles32:
@Gillou68310 : I'm not favorable to your first commit (ie manually porting to Intel syntax) because of the maintenance nightmare it involves (error-prone duplication of code).
I think it would be better to use NASM or YASM to assemble the original linkage_x86.S file and make it output a Windows compatible object file instead.
This way, the project can be built with Visual Studio, and GCC without having to maintain manually different versions of the same thing.

[Better, don't use ASM altogether and prefer portable C code instead :) ]

Gillou68310:
> I think it would be better to use NASM or YASM to assemble the original linkage_x86.S file and make it output a Windows compatible object file instead.

NASM doesn't support AT&T format unfortunately.
I managed to build the core using YASM to compile linkage_x86.S (and after adding an underscore at the beginning of every symboles, damn you microsoft!!!!)

Also such instruction are not allowed by YASM: ```push	$0xa4000040```
So you have to move the value to a register then push the register on stack.

And at the end the new dynarec is looping forever without ever triggering an interrupt :-(

Honestly I don't trust YASM and IMO I don't think we should rely on it for this.

Gillou68310:
Here's my work so far, don't hesitate to comment: 
https://github.com/Gillou68310/mupen64plus-core/commit/c45cedcd78c8845c2cb554ec56845d2236ad586e

I'll re-open the PR once we find an agreement on the "linkage_x86.S" thing ;-)

bsmiles32:
> (and after adding an underscore at the beginning of every symboles, damn you microsoft!!!!)
Maybe declaring the functions as extern "C" could help in the C file.

> Also such instruction are not allowed by YASM: push $0xa4000040
> And at the end the new dynarec is looping forever without ever triggering an interrupt :-(
Fill a bug report for YASM, so everyone can benefit from that ?

It is just that I think that the new_dynarec is already a complicated beast, so I want to avoid any additional maintenance burden.

littleguy77:
> It is just that I think that the new_dynarec is already a complicated beast, so I want to avoid any additional maintenance burden.

If the past is any predictor of the future, new_dynarec probably won't experience much maintenance other than to fix build errors.  @Gillou68310 is making the first real fixes to it since it was pulled upstream more than two years ago.  It's by no means a knock against any devs (I exalt the ground asm devs walk on), but it's just the reality of our current situation.

bsmiles32:
The problem though with the new_dynarec is that it slows down other refactorings currently done in the core. For instance, my pull request on the memory refactoring has been waiting for more than 1 month just to be compatible with it, and still isn't ready... Time spent for the new_dynarec is not time spent improving other areas of the core...

littleguy77:
Ah, I see.  Thanks for clarifying.

Gillou68310:
> Maybe declaring the functions as extern "C" in the C file could help.

I'm already using extern "C" for functions used in asm ;-)

> It is just that I think that the new_dynarec is already a complicated beast, so I want to avoid any additional maintenance burden.

I agree with you, it's just that there's no good way to build AT&T assembly on windows. Maybe we should do the opposite I think GCC has no problem building intel format assembly ;-)

littleguy77:
Not to complicate matters further, but could the choice of MSVC version make life any easier?  If so, it would be a good excuse to standardize mupen64plus on one concrete version of MSVC.  Crazy that we maintain project files for MSVC 2008, 2010, 2012, etc.

wareya:
MSVC project files are not interportable and different versions of MSVC are sometimes not- or only-available.

richard42:
I don't mind assembly code, especially if it's necessary for weird dynarec stuff (jumping into/out of dynamic code, making functions with no stack storage) or SIMD (mmx/sse/avx for speed).  But I do agree that having separate AT&T and Intel-style code increases the maintenance work.  I would be okay with standardizing on intel-style, if Paul and the other downstream project users don't mind.  You can get around the underscore function name issue by defining a macro named something like DefineGlobalFunc which adds an underscore to the function name for Win32 builds.  Use the macro for each function definition.

littleguy77:
> I would be okay with standardizing on intel-style, if Paul and the other downstream project users don't mind

I can't even weigh in on that, but @Gillou68310 has a good handle on the Android needs (which I'm sure he's already considered) due to his involvement downstream.

bsmiles32:
From reading various sources, Intel syntax vs AT&T syntax is mostly a matter of preference.
GCC/GAS supports both syntaxes either with a switch (-masm=intel / -masm=att) or inside the assembly file via an assembler directive (.intel_syntax  / .att_syntax).
Same story with Clang thanks to its compatibility with GCC.

On another note, you added building support for MSVC 2010. While I find this good, it increases the maintenance burden when adding/deleting a file to the project : When I refactor I currently have to update the unix Makefile (just one line because only .c file is required), and 2 msvc project files which both requires me to add 1/3 lines describing the new files locations (C + header).
With your new msvc project addition, this brings to 3 the number of msvc projects file to update for the SAME plateform (ie Windows), whereas the Makefile works for a variety of plateform (Linux, BSD, ...). Can we agree on a single MSVC version ? Or switch to CMake ?
[Yeah I'm sorry to sound like the boring guy, but really, we must make it as painless as possible to read/modify/maintain the source if we want to make further progress and attract more talented devs].

littleguy77:
I've always seen the Unix makefiles as the official standard and the MSVC files as gravy.  It's much easier to infer the build structure from the former than the latter.  I have never expected the same level of support in MSVC as the makefiles.  I don't see any issue with standardizing on a single MSVC version, since you can still download and install older versions.

Here are the official support lifecycle links for 2010 (msvc10), 2012 (msvc11), and 2013 (msvc12)

http://support2.microsoft.com/lifecycle/default.aspx?LN=en-us&x=8&y=13&p1=14048
http://support2.microsoft.com/lifecycle/default.aspx?LN=en-us&x=11&y=6&p1=16823
http://support2.microsoft.com/lifecycle/default.aspx?LN=en-us&x=13&y=14&p1=17502

@Gillou68310 Is there a particular feature/compatibility issue that pushed you to 2010?

Gillou68310:
> @Gillou68310 Is there a particular feature/compatibility issue that pushed you to 2010?

I have a licence for MSVC2010 professional ;-)

Gillou68310:
@richard42 "word" and "dword" are part of NASM keywords do you mind if I rename them like this:

```
unsigned int cpu_word;
unsigned char cpu_byte;
unsigned short cpu_hword;
unsigned long long int cpu_dword;
```

bsmiles32:
I don't mind the rename. I might have to update my "memory" pull request to accommodate this change however.

Gillou68310:
Ok I made some progress on this, here's my work:
https://github.com/Gillou68310/mupen64plus-core/commits/new_dynarec_msvc

I didn't remove linkage_x86.S for the moment in case of regression.

It's building and running fine on both windows and android x86 (with ndk-r10 using GCC-4.8 as default toolchain).

@littleguy77 if this gets merged you only need to change the extension from ".S" to ".asm" this will also tell GAS to use intel format.
```
else ifeq ($(TARGET_ARCH_ABI), x86)
    # Use for x86:
    LOCAL_SRC_FILES += $(SRCDIR)/r4300/new_dynarec/linkage_x86.asm
    LOCAL_CFLAGS += -DDYNAREC
    LOCAL_CFLAGS += -DNEW_DYNAREC=1
```

@bsmiles32 would you mind trying to build this branch on linux ;-)
https://github.com/Gillou68310/mupen64plus-core/tree/new_dynarec_linux

> With your new msvc project addition, this brings to 3 the number of msvc projects file to update for the SAME plateform (ie Windows), whereas the Makefile works for a variety of plateform (Linux, BSD, ...). Can we agree on a single MSVC version ? Or switch to CMake ?

@bsmiles32 again I agree with you but is that ok for everyone if we deal with this in another PR?

bsmiles32:
@Gillou68310 : Sure :) I'll test when I get back home.
I also agree to deal with the multiple build systems in another PR.

littleguy77:
@Gillou68310 Perfect, thanks!  For upstream devs wondering what he's talking about, that's a snippet from Android's unique makefile, maintained downstream.

Gillou68310:
Ok I just finished installing ubuntu in virtualbox.

The core is building just fine using gcc 4.6.4 but when I try to run mupen64plus I get this :-(

```
 __  __                         __   _  _   ____  _             
|  \/  |_   _ _ __   ___ _ __  / /_ | || | |  _ \| |_   _ ___ 
| |\/| | | | | '_ \ / _ \ '_ \| '_ \| || |_| |_) | | | | / __|  
| |  | | |_| | |_) |  __/ | | | (_) |__   _|  __/| | |_| \__ \  
|_|  |_|\__,_| .__/ \___|_| |_|\___/   |_| |_|   |_|\__,_|___/  
             |_|         http://code.google.com/p/mupen64plus/  
Mupen64Plus Console User-Interface Version 2.0.0

UI-Console Error: dlopen('./libmupen64plus.so.2') failed: ./libmupen64plus.so.2: undefined symbol: verify_code_vm
UI-Console Error: AttachCoreLib() Error: failed to find Mupen64Plus Core library
```

Seems like something went wrong during link time.
Just for information "verify_code_vm" is an assembly function defined in linkage_x86

Old dynarec is working just fine BTW

Gillou68310:
Ok nevermind I just made an objdump on libmupen64plus.so and it seems like linkage_x86 has been excluded at compile time.

conchurnavid:
@Gillou68310 Have you changed following lines?

https://github.com/mupen64plus/mupen64plus-core/blob/e91fa610fe35ffd0319f320aadf662d059ad7826/projects/unix/Makefile#L469
https://github.com/mupen64plus/mupen64plus-core/blob/e91fa610fe35ffd0319f320aadf662d059ad7826/projects/unix/Makefile#L522
https://github.com/mupen64plus/mupen64plus-core/blob/e91fa610fe35ffd0319f320aadf662d059ad7826/projects/unix/Makefile#L594

richard42:
I'm fine with the hword/word/dword renaming.  Those variable names were too generic.

richard42:
My attitude to the MSVC project files is similar to littleguy77.  The primary build system is the unix makefile..  And it's easy to make Win32 builds with MXE, so if the MSVC project files are broken from time to time as things change, it doesn't bother me.  They can be fixed whenever someone uses them.

bsmiles32:
@Gillou68310 : Tested on my machine, pull request sent.
@conchurnavid : Thanks, you pretty much solved it

Gillou68310:
Thanks for your help @conchurnavid ;-)
@bsmiles32 I merged and tested your PR but for some reason the compiler is still using linkage_x86.S instead of .asm. I tried removing linkage_x86.S but the compiler fails when trying to build the .asm.
Also I tried building for android x86 using GCC4.6 (instead of the default 4.8) and it also fails, so I guess this is a GCC4.8 only feature :-(

Any ideas?

bsmiles32:
What's the error message when GCC fails ?

Gillou68310:
```
    CXX _obj/osd/osd.o
    CC  _obj/r4300/new_dynarec/linkage_x86.o
gcc: warning: ../../src/r4300/new_dynarec/linkage_x86.asm: linker input file unused because linking not done
    LD  libmupen64plus.so.2.0.0
g++: error: _obj/r4300/new_dynarec/linkage_x86.o: No such file or directory
make: *** [libmupen64plus.so.2.0.0] Error 1

```

bsmiles32:
Maybe there is an ambiguity in the makefile for which assembly file has to be assembled.
Can you try renaming your linkage_x86.asm to linkage_x86.S and assemble it with -masm=intel parameter ?

Gillou68310:
> Can you try renaming your linkage_x86.asm to linkage_x86.S and assemble it with -masm=intel parameter ?

I tried renaming and setting -masm=intel like this:
```
$(OBJDIR)/%.o: $(SRCDIR)/%.S
	$(COMPILE.c) -masm=intel -o $@ $<
```

Seems like the intel format isn't recognized.

Just an example of the last 3 errors:
```
../../src/r4300/new_dynarec/linkage_x86.S:869: Error: junk `[last_count]' after expression
../../src/r4300/new_dynarec/linkage_x86.S:869: Error: too many memory references for `mov'
../../src/r4300/new_dynarec/linkage_x86.S:870: Error: too many memory references for `sub'

```

Gillou68310:
> @littleguy77 if this gets merged you only need to change the extension from ".S" to ".asm" this will also tell GAS to use intel format.

Ok this has nothing to do with GAS. YASM has been included into the NDK since r9d, here's the change log:
```
Added the yasm assembler, as well as LOCAL_ASMFLAGS and EXPORT_ASMFLAGS flags for x86 targets. The ndk-build script uses prebuilts/*/bin/yasm* to build LOCAL_SRC_FILES that have the .asm extension.
```
This also explains why I can't build linkage_x86.asm on linux. GCC has support for intel format but it has no support for NASM syntax.

IMO the only way to share this assembly file between on all platforms is to build with NASM on every platforms.

I updated the makefile to build with NASM here:
https://github.com/Gillou68310/mupen64plus-core/commit/4343e1d7b3bd942df0a4f36b07284c96c841c2c7

It's working fine on linux but it still needs to be tested on OSX + MINGW

Gillou68310:
> IMO the only way to share this assembly file between all platforms is to build with NASM on every platforms.

@richard42, @bsmiles32 is that okay for you?

According to NASM documentation (see 9.1.1 External Symbol Names)
http://www.nasm.us/doc/nasmdoc9.html
```
Most 32-bit C compilers share the convention used by 16-bit compilers, that the names of all global symbols (functions or data) they define are formed by prefixing an underscore to the name as it appears in the C program. However, not all of them do: the ELF specification states that C symbols do not have a leading underscore on their assembly-language names. 
```

I made the changes accordingly:
https://github.com/Gillou68310/mupen64plus-core/commit/a3a54f22a4cda46fae547c8422c766bb302f1540

@littleguy77 you will need to define ELF_TYPE in Android.mk:
```
else ifeq ($(TARGET_ARCH_ABI), x86)
    # Use for x86:
    LOCAL_ASMFLAGS = -d ELF_TYPE
    LOCAL_SRC_FILES += $(SRCDIR)/r4300/new_dynarec/linkage_x86.asm
    LOCAL_CFLAGS += -DDYNAREC
    LOCAL_CFLAGS += -DNEW_DYNAREC=1
```

Also NDK-r9d will be the minimum requirement in order to build mupen64plus-ae is that okay for you?

littleguy77:
No problem, I have always used the latest NDK without issue.  I'm running NDK-r10d now, and I think Paul is too.

richard42:
I'm fine with requiring NASM to build this file on all platforms.

PR #61, opened 2015-01-21
Makefile: Add VFP hardfloat as an option

gizmo98:
https://github.com/ricrpi/mupen64plus-core/commit/1cf54788f636c43de57f6c2f8c8dfc7928332f18


Narann:
Thanks, but you forgot to update the doc at the end of the Makefile. I suggest under the:

```
NEON=1 == (ARM only) build for hard floating point environments"
```
Something like this:
```
VFP_HARD=1 == (ARM only) Full hardware floating point ABI"
```
As you can see, I also suggest to use the name `VFP_HARD` because `VFP` has no real meaning as what we change is floating ABI.

gizmo98:
All right. I added your suggestions.

Narann:
Thanks!

PR #62, opened 2015-01-24
Cenification part1

bsmiles32:
This PR provides better modularization of n64 core components.
It is largely inspired by the work of MarathonMan for his cycle accurate emulator CEN64.

With the new modularization, several new subsystems have been introduced:
-AI (Audio Interface) : it contains for now only the audio interface controller.
-PI (Parallel Interface): it contains the parallel interface controller, the cart rom, the flashram and sram modules.
-RI (RDRAM Interface): it contains the RDRAM interface controller, and the RDRAM module.
-SI (Serial Interface): it contains the serial interface controller, the PIF, the game controllers, mempaks, the cart CIC, cart EEPROM and cart AF-RTC.
-VI (Video Interface): it contains for now only the video interface controller.
-RSP: it contains only the RSP core module.
-RDP: it contains the RDP core module and the FrameBuffers monitoring informations.
-R4300: it contains the R4300 core module, the MIPS interface controller, and everything that was already present before.

All new modules avoid heavy usage of global variables except when no other option was possible (plugin, stuff for instance).

With this PR, I tried not to refactor too much the individual modules in order to prevent accidental behavior changes. Individual refactorings will come later.
I have also avoided heavy refactorings within the R4300 subsystem because it would very likely break the NEW_DYNAREC. This will also be addressed in a subsequent PR.

PR #65, opened 2015-02-01
Si refactorings

bsmiles32:
This PR should be applied after PR #62 .

Main features of this PR includes:
- separation of save chips emulation (mempak,eeprom,sram,flashram) from their file-based storage.
- no more file accesses everytime a save chip is accessed. Files are read/written only at the beginning/end of emulation.
- separation of game controller emulation from the input plugin. For now we still call the input plugin to read controllers buttons, but in the future, we should let the frontend decide what to do (input via touchscreen, via keypress file (TAS), ...)
- integration of rumblepak emulation inside the core. For now we still call the input plugin to rumble, but in the future, we should let the frontend decide what to do (shake screen, trigger sound, rumble controller, ...)
- allow to specify alternate time source for RTC emulation. For now we only provide a C localtime time source, but other possibilities could be implemented (for instance a user defined date whoch could be helpful for TAS).
- various refactorings

All in all, this PR provides more flexibility "under the hood", and should be seen as a first step toward moving more responsibilities to the frontend, and removing dependencies in the core.

EDIT: due to concerns about saved data loss, we now flush saved data to disk immediately.

### DIFF
--- a/projects/msvc11/mupen64plus-core.vcxproj
+++ b/projects/msvc11/mupen64plus-core.vcxproj
@@ -136,6 +136,7 @@
     <ClCompile Include="..\..\src\r4300\exception.c" />
     <ClCompile Include="..\..\src\rdp\fb.c" />
     <ClCompile Include="..\..\src\osal\files_win32.c" />
+    <ClCompile Include="..\..\src\main\fla_file.c" />
     <ClCompile Include="..\..\src\pi\flashram.c" />
     <ClCompile Include="..\..\src\api\frontend.c" />
     <ClCompile Include="..\..\src\si\game_controller.c" />
@@ -228,6 +229,7 @@
     <ClInclude Include="..\..\src\r4300\exception.h" />
     <ClInclude Include="..\..\src\rdp\fb.h" />
     <ClInclude Include="..\..\src\osal\files.h" />
+    <ClInclude Include="..\..\src\main\fla_file.h" />
     <ClInclude Include="..\..\src\pi\flashram.h" />
     <ClInclude Include="..\..\src\si\game_controller.h" />
     <ClInclude Include="..\..\src\plugin\get_time_using_C_localtime.h" />
--- a/projects/msvc11/mupen64plus-core.vcxproj
+++ b/projects/msvc11/mupen64plus-core.vcxproj
@@ -144,6 +144,7 @@
     <ClCompile Include="..\..\src\r4300\x86\gcop1_l.c" />
     <ClCompile Include="..\..\src\r4300\x86\gcop1_s.c" />
     <ClCompile Include="..\..\src\r4300\x86\gcop1_w.c" />
+    <ClCompile Include="..\..\src\plugin\get_time_using_C_localtime.c" />
     <ClCompile Include="..\..\src\r4300\x86\gr4300.c" />
     <ClCompile Include="..\..\src\r4300\x86\gregimm.c" />
     <ClCompile Include="..\..\src\r4300\x86\gspecial.c" />
@@ -221,6 +222,7 @@
     <ClInclude Include="..\..\src\osal\files.h" />
     <ClInclude Include="..\..\src\pi\flashram.h" />
     <ClInclude Include="..\..\src\si\game_controller.h" />
+    <ClInclude Include="..\..\src\plugin\get_time_using_C_localtime.h" />
     <ClInclude Include="..\..\src\r4300\instr_counters.h" />
     <ClInclude Include="..\..\src\r4300\x86\interpret.h" />
     <ClInclude Include="..\..\src\r4300\interupt.h" />
--- a/projects/msvc11/mupen64plus-core.vcxproj
+++ b/projects/msvc11/mupen64plus-core.vcxproj
@@ -130,6 +130,7 @@
     <ClCompile Include="..\..\src\plugin\dummy_video.c" />
     <ClCompile Include="..\..\src\osal\dynamiclib_win32.c" />
     <ClCompile Include="..\..\src\si\eeprom.c" />
+    <ClCompile Include="..\..\src\plugin\emulate_game_controller_via_input_plugin.c" />
     <ClCompile Include="..\..\src\main\eventloop.c" />
     <ClCompile Include="..\..\src\r4300\exception.c" />
     <ClCompile Include="..\..\src\rdp\fb.c" />
@@ -178,6 +179,8 @@
     <ClCompile Include="..\..\src\r4300\x86\rjump.c" />
     <ClCompile Include="..\..\src\main\rom.c" />
     <ClCompile Include="..\..\src\rsp\rsp_core.c" />
+    <ClCompile Include="..\..\src\plugin\rumble_via_input_plugin.c" />
+    <ClCompile Include="..\..\src\si\rumblepak.c" />
     <ClCompile Include="..\..\src\main\savestates.c" />
     <ClCompile Include="..\..\src\main\sdl_key_converter.c" />
     <ClCompile Include="..\..\src\osd\screenshot.cpp" />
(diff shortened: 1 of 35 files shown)


littleguy77:
Wow, another beautiful set of commits!  :smile: 

I can't really absorb the details of what's going on here, but I just had a few questions/concerns based on your PR comment:

> - no more file accesses everytime a save chip is accessed. Files are read/written only at the beginning/end of emulation.

What happens if the app crashes (or pc loses power, battery dies, etc.)?  Would the player lose all their in-game save progress for the session?

> - separation of game controller emulation from the input plugin. For now we still call the input plugin to read controllers buttons, but in the future, we should let the frontend decide what to do (input via touchscreen, via keypress file (TAS), ...)

I'm not sure I understand.  I thought the whole purpose of the input plugin is to abstract the input stream?  mupen64plus-ae has its own custom input plugin that marshals all the touchscreen/controller inputs from the Java Android APIs to the native core: https://github.com/mupen64plus-ae/mupen64plus-ae/tree/master/jni/mupen64plus-input-android.  I like this decoupling and it keeps all the android junk out of the core.  Am I misunderstanding what you are trying to say?

> - integration of rumblepak emulation inside the core. For now we still call the input plugin to rumble, but in the future, we should let the frontend decide what to do (shake screen, trigger sound, rumble controller, ...)

Again, I might be missing something.  Why would we want to re-couple rumble behavior with the core?  Or are you suggesting we split the input plugin into
 - controls plugin (buttons, sticks, triggers, touchscreen, i.e. communication from user to core)
 - rumble plugin (i.e. communication from core to user)

bsmiles32:
> What happens if the app crashes (or pc loses power, battery dies, etc.)? Would the player lose all
> their in-game save progress for the session?

Progress is lost because save chips content is flushed to file only after emulation is done.
I understand that it might not be optimal (especially in case of app crash).
I can eventually modify the code to flush save chips content :
- after each write (and we get the original behavior)
- or have a separate thread checking periodically if things need to be flushed (?)

> I'm not sure I understand. I thought the whole purpose of the input plugin is to abstract the input stream? 

Yeah the input plugin "abstracts" in theory the input stream.
But in practice, the input plugin does much more that just providing "input streams" because it  has to interpret PIF commands. That is not desirable as it forces a plugin to know about the emulated console, and lead to akward duplication of effort (see PLUGIN_RAW for instance).
With what I propose, all the PIF command interpretation is done in the core and user callbacks take
care of provinding the necessary input/output. You can look at these callbacks as very focused "plugins".

Currently you can specify independently :
- how the game controller is emulated via the is_connected/get_input callbacks.
- how the rumble pak is emulated via the rumble callback.
- how the AF RTC is emulated via the get_time callback.

This allow to better isolate the "core" from its inputs and outputs.
Furthermore, it could allow for some cool stuff if they get implemented :
- using an input file to generate game controllers pressed buttons sequences
- using a touch screen to generate game controllers inputs
- making the screen shake, or play a user defined sound when rumble is activated.
- allow to specify a user defined date for RTC starting date

All of that without having to interpret PIF commands, because this is done in the core.
This will greatly simplify the writing of such callbacks :)

For now, I've connected all these callbacks to functions which reproduce the old behavior (ie
call the input plugin for game_controllers and rumble and use C localtime for RTC).
But ultimately, I'd like to get rid of the input plugin all together and let the frontend the callbacks it wants, because the frontend knows better what kind of input/output combination is best adapted.
Furthermore, by moving this responsibilty to the frontend we might get rid of SDL inside the core,
and let the frontend do the right initialization.

littleguy77:
Thanks much for the clarification, I think I see what you mean now -- better encapsulation, not less encapsulation.  I don't mind if I have to refactor some android stuff, and I'll just go along with whatever everyone thinks.

My lingering concern is only the persisting of SRAM data.  I strongly feel that the data should be persisted immediately, and it should be done on-demand (rather than by a separate polling thread).  This matches the logical behavior of the original N64 console and maximizes data integrity in the event of crash or power loss.  I can't tell you how many times an angry Android user has complained about loss of progress (in most cases due to their own error).  I'd hate to have yet another source of the issue.  Just last night the app crashed on me, and thankfully I had the SRAM data to recover with.

wareya:
Agreed dramatically w/r/t persistence. It needs to be such that when someone saves ingame it's written to disk as soon as is reasonably possible.

bsmiles32:
Okay. I will update later this PR to flush immediately saves to disk.

Question: In this PR all commits from PR #62 appears (because I used it as a starting point).
Is it possible to "hide" all these redundant commits (up to 4d29372 included) from this PR since it obfuscate the real changes ?

littleguy77:
Thanks @bsmiles32.

I don't think there's a way to show just the increment to the previous pull request.  Is there a reason you didn't just extend the previous pull request rather than make a new one?

bsmiles32:
Okay, saved data are now flush immediately.

@littleguy77 : I didn't extended the other PR because it was already big enough, and topics are slightly differents : separation into several subsystems for the previous PR (without too much per subsystem refactorings); and this one exclusively with SI refactorings.

littleguy77:
Excellent, thanks much :)

You might want to rename the functions `write_***_file` rather than `touch_***_file`, for clarity

Narann:
I really love your work!

However, I would suggest to open a ticket to discuss such points before provide a PR. This make things clear for everybody!

About the save _problem_, I think everybody didn't understood well: We only talk about the save chip, not the save states. Save state are still written as soon as F5 is pressed. I think having the save chip cached for read (no file access) is fine and only write to file when the save chip memory is written (I guess it's what you do).

I also think that only save at the end of emulation was a bad idea, thanks for the fix!

I like your vision about the core bsmile! I don't like to have plugins with the same (buggy) parsing code and like the callback idea (camera shake idea is quite fun haha!). I like to have this principle spreaded for the whole mupen64plus eco system (so break the API... mupen64plus 3.0? :D).

But, I would really love to discuss because there is a lot of peoples involves that would maybe have a different perspective (And to be honest I'm not the best programmer ever so there is maybe con to this we should eard). Starting with input plugins (between the official one and the AE one) is an interesting beggening.

So, I suggest to create an issue with RFC tag stating what you will do. If nobody answer, it's there problem and go ahead but having a RFC protect you from any futur critic if you break something. I know this "management" seem heavy but I don't want any war between devs about something integrated without letting others discuss about it. And during this discuss, good ideas could appear! What do you think bsmile32?

And of course: What do you think of all of this @richard42? :)

PS: Travis [fail to build](https://travis-ci.org/mupen64plus/mupen64plus-core/jobs/49210878) but I'm not sure it's related to your work.

littleguy77:
>  About the save problem, I think everybody didn't understood well: We only talk about the save chip, not the save states. ... I think having the save chip cached for read (no file access) is fine and only write to file when the save chip memory is written (I guess it's what you do).

I did understand the distinction, and bsmiles' latest commit addresses my concerns perfectly :)  Writing is what needs to be done immediately for data integrity.  Reading can be cached.

> So, I suggest to create an issue with RFC tag stating what you will do. If nobody answer, it's there problem and go ahead but having a RFC protect you from any futur critic if you break something. I know this "management" seem heavy but I don't want any war between devs about something integrated without letting others discuss about it. And during this discuss, good ideas could appear!

Couldn't agree more.  For major API changes we should cast a wide net on the google list serv as well, perhaps even contact key devs directly.

bsmiles32:
> However, I would suggest to open a ticket to discuss such points before provide a PR. This make
> things clear for everybody!

For this PR, I thought it wouldn't need much more discussion as all my changes here are fully retro-compatible, and only provide some more flexibility "under the hood".
And I find easier to discuss with code to look at :)
But point taken, will do next time I want to discuss things, especially things that impact downstream.
Which leads me to....

> I like to have this principle spreaded for the whole mupen64plus eco system (so break the API...
> mupen64plus 3.0? :D).

We have to be REALLY carefull here as plugin removal is not something trivial, and require high coordination with other projects involved (frontends, and ports mostly).
This need to be discussed in a separate ticket and plublic mailing list.

> PS: Travis fail to build but I'm not sure it's related to your work.
Yeah, it fails while setting up the dependencies, so not related to my work.
Maybe asking travis to PS: Travis fail to build but I'm not sure it's related to your work.rebuild (how ?) could change that.

> You might want to rename the functions write_***_file rather than touch_***_file, for clarity

Will do later.

Narann:
> We have to be REALLY carefull here as plugin removal is not something trivial, and require high coordination with other projects involved (frontends, and ports mostly).
> This need to be discussed in a separate ticket and plublic mailing list.

I agree, discussing input plugins could be interesting as it should not require huge refactoring.

bsmiles32:
@richard42: do you want me to rebase this PR now that PR #62 is merged ?

PR #66, opened 2015-02-03
Fix VFP_HARD=1 compilation error,...

gizmo98:
...add Raspberry PIs Videocore IV GPU as an option and fix non reachable HAVE_GLES/OSD path.

### DIFF
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -100,7 +100,7 @@ ifneq ("$(filter arm%,$(HOST_CPU))","")
         CFLAGS += -mfpu=neon -mfloat-abi=hard -DNEON
     else
       ifeq ($(VFP_HARD), 1)
-        CFLAGS += -mfpu=vfp -mfloat-abi=hard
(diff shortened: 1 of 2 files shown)


gizmo98:
It's possible to use gcc macros instead of -DNEON -DVFP_HARD and so on. So NEON and VFP_HARD options could be certainly removed:
https://github.com/gizmo98/mupen64plus-core/commit/b2b13851ee0bb25c39d6e7b5ce1b2f9168e3ef18

VC wouldn't be necessary if GL_LDLIBS and GL_CFLAGS could be used as input variables. GL_LDLIBS = -lGLESv2 overrides any 'GL_LDLIBS="-L/opt/vc/lib" make' stuff. This could be fixed with GL_LDLIBS += -lGLESv2.

Narann:
I'm not sure I get everything. You want to create another PR?

gizmo98:
I am just testing a cleaner way to set or check neon, vfp_hard or vc lib path. If it is working i will create a PR.

PR #67, opened 2015-02-06
Raspberry Pi fix and ARM NEON/VFP_HARD cleanup

gizmo98:
-ARMv5 and ARMv6 (like Raspberry PIs CPU) need -DARMv5_ONLY (assem_arm.c). Check CPU and set define if ARMv5 or ARMv6 is present.
-Use gcc macros instead of -DNEON and -DVFP_HARD (linkage_arm.S and Makefile)
-Raspberry PI: Mupen64plus-core is running now with "make VFP_HARD=1 VC=1 OSD=0". Default rice plugin still fails with a core error (SDL_SetVideoMode failed). mupen64plus-video-gles2n64 plugin from ricrpi is working. GLES stuff needs some investigation.

### DIFF
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -97,10 +97,10 @@ ifneq ("$(filter arm%,$(HOST_CPU))","")
     NEW_DYNAREC := 1
     CFLAGS += -marm
     ifeq ($(NEON), 1)
-        CFLAGS += -mfpu=neon -mfloat-abi=hard -DNEON
+        CFLAGS += -mfpu=neon -mfloat-abi=hard
     else
       ifeq ($(VFP_HARD), 1)
-        CFLAGS += -mfpu=vfp -mfloat-abi=hard -DVFP_HARD
+        CFLAGS += -mfpu=vfp -mfloat-abi=hard
       else
         CFLAGS += -mfpu=vfp -mfloat-abi=softfp
       endif
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -96,6 +96,14 @@ ifneq ("$(filter arm%,$(HOST_CPU))","")
     PIC ?= 1
     NEW_DYNAREC := 1
     CFLAGS += -marm
+    ifneq ("$(filter armv5%,$(HOST_CPU))","")
+    	CFLAGS += -DARMv5_ONLY
+    	$(warning Using ARMv5_ONLY)
+    endif
+    ifneq ("$(filter armv6%,$(HOST_CPU))","")
+    	CFLAGS += -DARMv5_ONLY
+    	$(warning Using ARMv5_ONLY)
+    endif
     ifeq ($(NEON), 1)
         CFLAGS += -mfpu=neon -mfloat-abi=hard
     else
--- a/src/r4300/new_dynarec/linkage_arm.S
+++ b/src/r4300/new_dynarec/linkage_arm.S
@@ -18,25 +18,27 @@
  *   51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.          *
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 	.cpu arm9tdmi
-#ifndef NEON
-#ifndef VFP_HARD
-	.fpu softvfp
+#ifndef __ARM_NEON__
+#if (defined(__VFP_FP__) && !defined(__SOFTFP__))
+ 	.fpu vfp
 #else
-	.fpu vfp
+	.fpu softvfp
 #endif
 #else
     	.fpu neon
 #endif
 	.eabi_attribute 20, 1
 	.eabi_attribute 21, 1
-#ifndef NEON
+#ifndef __ARM_NEON__
 	.eabi_attribute 23, 3
 #endif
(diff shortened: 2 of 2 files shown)


Narann:
Thanks for this interesting updates. All this ifdef stuff make me sigh because they make the code very hard to read... I will let Richard the opportunity to review this (is this macros supported for gcc < 4.7?). I really don't know what a good cross platform way would be. :(

I also have the `SDL_SetVideoMode failed` on Raspberry Pi. I'm not an expert and my time to n64 dev is quite limited (unfortunately) so if anyone have any workaround to fix this it would be really appreciate and maybe allow me to finally fix a long standing but with the "refactored" rice plugin and so do the final merge. You can test using [my branch](https://github.com/Narann/mupen64plus-video-rice).

gizmo98:
I get the error message "SDL_SetVideoMode failed:Could not create GLES window surface". The error message "Could not create GLES window surface" can be found in SDL2s RPI driver function RPI_CreateWindow: http://www.libsdl.org/tmp/SDL/src/video/raspberry/SDL_rpivideo.c

Seems to be there is something wrong with SDL_CreateWindow() (parameters).

Narann:
Arf, I didn't get the same. I get `SDL_SetVideoMode failed:` without _any_ message...

Anyway...

gizmo98:
Do you have SDL2 with OpenGLES support? I use retropies SDL2.0.3 module. https://github.com/petrockblog/RetroPie-Setup/blob/master/scriptmodules/supplementary/sdl2.sh

Narann:
Interesting! I will test all of this.

gizmo98:
If i add a strange fix from ricrpis fork ricrpis gles2rice and gles2n64 are working:https://github.com/gizmo98/mupen64plus-core/commit/20a4a035b53586bff26db7b831e21771394bd5e6

But in the end it disables VidExt_GL_SetAttribute()?! This does not seem right. With this fix your rice plugin produces a segmentation fault instead of a SDL_SetVideoError message.

littleguy77:
We noticed in Android that context creation fails on some devices due to the EGL_BUFFER_SIZE attribute.  Try removing that attribute and repeating if SDL_CreateWindow fails.

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/master/src/paulscode/android/mupen64plusae/jni/NativeSDL.java#L59

gizmo98:
Could be also that the video plugins set false GL attributes or default settings are wrong. Ricrpis gles2rice plugin has the same error if VidExt_GL_SetAttribute uses SDL_GL_SetAttribute. With his strange fix rice plugins cannot set anything with CoreVideo_GL_SetAttribute but gles2rice is running.

gizmo98:
A new observation. With ricrpis fix there is no SDL_SetVideo Error. You can see a "video: context created" message and so on. But there is a segmentation fault after "init dynarec".

gizmo98:
Ricrpis gles2rice plugin is now working without this strange fix. Gles2rice default Z-Buffer setting is 32Bit. With a 16Bit Z-Buffer a context can be created. Naranns video-rice plugin produces an error message after "context creation" and "init dynarec". I will post the whole error message of Naranns rice plugin later.

Narann:
Yeah thanks!

joolswills:
Just a related question - in retropie we are currently building for rpi2 with

joolswills:
Just a quick question. In retropie we are currenyl building for the rpi2 with 

```
-mcpu=cortex-a7 -mfpu=neon-vfpv4 -mfloat-abi=hard
```

However mupen64plus-core will add the -mfpu=neon - One of the problems with makefiles that try and do everything is they can often override users parameters.

http://community.arm.com/groups/tools/blog/2013/04/15/arm-cortex-a-processors-and-gcc-command-lines

I'm not expert on arm architecture, but looking at the table, we don't want to use -mfpu=neon for the cortex-a7 but -mfpu=neon-vfpv4 ?

gizmo98:
The most emulators use -mfpu=neon. I haven't seen any -mfpu=neon-vfpv4 implementation. Have you already tested if PI 2 is working with -mfpu=neon/NEON=1? The cortex-A7 should be backward compatible.

joolswills:
Just got my pi so in process of testing etc now. Cheers.

Maybe we don't need -mfpu=neon-vfpv4 then - but I need to do more reading up on it.

gizmo98:
Hi Narann,
There is a segmentation fault after "init dynarec". I tested with Mario64. The screen is black. I can here Mario (It's a me mario) and then there is the segmentation fault. Any idea how to debug your rice plugin?

Narann:
YEAH! I think you've finally been able to reproduce the same bug as @littleGuy77! I miserably fail trying to (and lost a huge amount of time for this).

There is defenetely no easy way to debug OpenGL on rpi. You can try with [apitrace](https://apitrace.github.io/) but if the problem appear on context creation I'm even not sure you will have any valid OpenGL call.

If I succeed to compile all this stuff on my pi, my main debug way will certainly be massive printing, from SDL loading to the first batch of command, adding OPENGL_ERROR macro everywhere.

Could you contact me privately and explain me by a step by step how I can build all my stuff for rpi? I have a RPI B+.

Thanks in advance! :)

PR #69, opened 2015-02-08
Decouple the Pure Interpreter from the Cached Interpreter

Nebuleon:
*This pull request arose from [RFC: Deleting the Hacktarux JIT ("Old Dynarec")](https://groups.google.com/forum/#!topic/mupen64plus/4rOO7cX0toY).*

As of commit 44f1424898a2bdcb23bc11861b35b0e9f9855aca, the Pure Interpreter in mupen64plus simply forwards all opcodes to the Cached Interpreter and asks it to fill `precomp_instr` structures for the opcode at the Program Counter and the next one, then executes the first one with `PC->ops();`.

`PC->ops();` is an [indirect call which mispredicts 100% of the time](http://www.emulators.com/docs/nx25_nostradamus.htm), and the Cached Interpreter uses even more indirectly called functions drawn from arrays, killing performance on chips with large unpredicted branch penalties. In addition to those performance penalties, this also creates unnecessary coupling, which makes the Pure and Cached interpreters harder to properly isolate, pluginise, test, and so on.

This pull request proposes a Pure Interpreter that has its own opcode dispatcher and repeatedly calls a single function, `InterpretOpcode();`, which does not consult the Cached Interpreter.

It supports _IDLE branches and jumps with [a common piece of code](https://github.com/Nebuleon/mupen64plus-core/blob/851c130/src/r4300/pure_interp.c#L112-L126); it correctly kills all writes to MIPS integer register 0, whose value must stay 0 at all times; and it updates the debugger and the core comparison feature at every opcode except the delay slots of jumps, exactly as before.

However, if `fast_mem_access` returns `NULL`, it will crash with a `SIGSEGV` instead of writing "prefetch() execute address : ..." to standard output and exiting with `stop = 1;`.

### DIFF
--- a/src/r4300/interpreter_cop0.def
+++ b/src/r4300/interpreter_cop0.def
@@ -21,7 +21,7 @@
 
 DECLARE_INSTRUCTION(MFC0)
 {
-   switch(PC->f.r.nrd)
+   switch(rfs)
    {
       case CP0_RANDOM_REG:
         DebugMessage(M64MSG_ERROR, "MFC0 instruction reading un-implemented Random register");
@@ -29,7 +29,7 @@ DECLARE_INSTRUCTION(MFC0)
       case CP0_COUNT_REG:
         update_count();
       default:
-        rrt32 = g_cp0_regs[PC->f.r.nrd];
+        rrt32 = g_cp0_regs[rfs];
         sign_extended(rrt);
    }
    ADD_TO_PC(1);
@@ -37,7 +37,7 @@ DECLARE_INSTRUCTION(MFC0)
 
 DECLARE_INSTRUCTION(MTC0)
 {
-  switch(PC->f.r.nrd)
+  switch(rfs)
   {
     case CP0_INDEX_REG:
       g_cp0_regs[CP0_INDEX_REG] = (unsigned int) rrt & 0x8000003F;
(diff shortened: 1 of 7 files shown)


bsmiles32:
I just tested this PR and I am very happy with the outcome :
- less coupling
- clean code
- nice commits
- improved performance (I benchmarked the time necessary to emulate 1000 VI and saw a consistant decrease in time with this PR on my system [x86_64]).

Well done !

Nebuleon:
I had actually not benchmarked the PI in any way before or after making the changes, but would appreciate seeing numbers from your testing since you have them, @bsmiles32.

Clean code with as few dependencies as possible was indeed what I was aiming for, though. Forwarding to the CI is just way too complicated for nothing. I'm glad you like it. :)

bsmiles32:
I used the following command (replace $core$ and $rom$ with approriate values) :
time mupen64plus --emumode 0 --nospeedlimit --audio dummy --testshots 1000 --corelib $core$ $rom$

Before / After times (in seconds) to render 1000 VI on an Intel Broadwell 5Y10 cpu :
Majora's Mask : 20.5 / 19.3
Ocarina of Time : 21.8 / 19.4
Super Mario : 23.1 / 19.4
Perfect Dark : 20.4 / 19.6
Banjo Kazooie : 27.2 / 25.0

So yeah, nice improvement all over the place.

Issue #72, opened 2015-02-09
Build issue related to recent VFP changes


littleguy77:
@gizmo98

I am having difficulty with [this line](https://github.com/mupen64plus/mupen64plus-core/blob/683127a1ae51998cfa22d8abc52fe5c827c90b40/src/r4300/new_dynarec/linkage_arm.S#L40) when I try to compile for Android.  In our [makefile](https://github.com/mupen64plus-ae/mupen64plus-ae/blob/288c29d5a05e31d2cdd7e0de021bdc490cc072fa/jni/Android.mk#L140), we set the following CFLAGS for ARM-v7a:
```
    LOCAL_CFLAGS += -mfloat-abi=softfp
    LOCAL_CFLAGS += -mfpu=vfp
```
The GCC 4.8 linker throws an error ```... new_dynarec/linkage_arm.o uses VFP register arguments, output does not```.

If I comment out the line [marked here](https://github.com/mupen64plus/mupen64plus-core/blob/683127a1ae51998cfa22d8abc52fe5c827c90b40/src/r4300/new_dynarec/linkage_arm.S#L40), everything builds again, so it seems to me that either
 - ```__VFP_FP__``` or ```__SOFTFP__``` is not being set properly in Android NDK 10d, or
 - the ```#ifdefs``` here need to be tweaked, or
 - we need to tweak something in the Android makefile

Any thoughts?

@Gillou68310, @paulscode, @xperia64 Here is the commit on my branch where you can test, if you have any ideas: https://github.com/littleguy77/mupen64plus-ae/commit/f75c70ee974adf73b76a9550e2d22cc5d975c707

littleguy77:
The following will work if all else fails.  But it's just a band-aid and doesn't address the underlying issue.
```
#if (defined(__VFP_FP__) && !defined(__SOFTFP__) && !defined(__ANDROID__))
```

littleguy77:
[This page](https://wiki.debian.org/ArmEabiPort) suggests that the problem is with the ```#ifdefs```.  They need to be tweaked to properly detect ```mfloat-abi=softfp```, and skip that line if so.

> Paradoxically, the -mfloat-abi=softfp does not set the __SOFTFP___ macro, since it selects real floating point instructions using the soft-float ABI at function-call interfaces.

littleguy77:
I think this is the proper ifdef clause:
```
[...]

PR #73, opened 2015-02-10
linkage_arm.S: Fix VFP SOFTFP target

gizmo98:
https://github.com/mupen64plus/mupen64plus-core/issues/72

### DIFF
--- a/src/r4300/new_dynarec/linkage_arm.S
+++ b/src/r4300/new_dynarec/linkage_arm.S
@@ -19,7 +19,7 @@
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 	.cpu arm9tdmi
 #ifndef __ARM_NEON__
-#if (defined(__VFP_FP__) && !defined(__SOFTFP__))
+#if (defined(__VFP_FP__) && !defined(__SOFTFP__) && defined(__ARM_PCS_VFP))
  	.fpu vfp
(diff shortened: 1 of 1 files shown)


src/r4300/new_dynarec/linkage_arm.S
```diff
@@ -19,7 +19,7 @@
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 	.cpu arm9tdmi
 #ifndef __ARM_NEON__
-#if (defined(__VFP_FP__) && !defined(__SOFTFP__))
+#if (defined(__VFP_FP__) && !defined(__SOFTFP__) && defined(__ARM_PCS_VFP))
```
littleguy77: Thanks for catching this one too :)

littleguy77:
Thanks @gizmo98 !

@richard42 @Narann This fixes #72.

Narann:
gcc specific macro? Mmmh, I will let @richard42 review this. It's typically the kind of thing which make code maintenance complex.

littleguy77:
If you take issue with the use of GCC macros, then PR #67 should have been rejected.  This PR simply corrects the usage of the GCC macros introduced in PR #67.

This is a regression fix.  Without it, mupen64plus-ae cannot be built.

littleguy77:
Thanks much :)

PR #74, opened 2015-02-11
Fix vi->field update when in interlaced mode.

bsmiles32:
vi->field value is assumed to be either 0 or 1 for VI_CURRENT_REG
update. The previous vi->field update formula did not guarantee that,
and therefore could lead to corrupted VI_CURRENT_REG values.

This regression was introduced in commit f9308ad.


Narann:
Look ok to me.

PR #75, opened 2015-02-13
Implemented recompiled DMULT instruction on ARM

Gillou68310:


Narann:
I like all those `assert()`. I have no opinion as it's a quite complex code. Anyone could test?

Gillou68310:
I launched a test campaign 1 month ago on paulscode forum. Everything seems to be fine there. Might be worth testing this on Raspberry pi armv6.

PR #77, opened 2015-02-13
DIV/DIVU instructions use hardware integer divide if available

Gillou68310:


bsmiles32:
I can't really judge the DIV/DIVU part (I trust what you've done), but I have suggestions for the ARM cpu feature detection :
- why in C++ instead of C ? (I don't think it really matters but most of the codebase is in C)
- I think arm_cpu_features could actually go in the osal directory as it is plateform dependant and not emulation related
- Furthermore it seems to assume a linux plateform by parsing the /proc/cpuinfo file so I would suggest putting compilation time checks for the linux plateform and #error with unsupported plateform otherwise.

What do you think ?

Gillou68310:
> why in C++ instead of C ? (I don't think it really matters but most of the codebase is in C)

Just a matter of habit I guess, strings are easier to manipulate in C++ IMO

> I think arm_cpu_features could actually go in the osal directory as it is plateform dependant and not emulation related

Totaly agree ;-)

> Furthermore it seems to assume a linux plateform by parsing the /proc/cpuinfo file so I would suggest putting compilation time checks for the linux plateform and #error with unsupported plateform otherwise.

The new_dynarec on ARM is already linux dependant that's why I didn't make the check.

richard42:
I agree that it makes sense to put this in the OSAL directory.  Otherwise I think this is okay.  I would prefer if it were C instead of C++ but it's not really a dealbreaker.

littleguy77:
In all honesty I think it would be a lot of work for nothing (at least right now) to put this in the OSAL directory.  Do these CPU features even make sense for non-ARM devices?  Could they be useful to the old dynarec?  If the answer is no to both, they seem like implementation-specific needs for the ARM new dynarec, which people are only compiling with GCC right now AFAIK.

If the focus is to decouple old and new dynarec in the immediate future, it would seem that putting this in OSAL right now would just add more potential coupling.  If other pieces of code eventually want to use the cpuinfo stuff, let them extract it and encapsulate it at that point.  But they've made it this fare without needing it.

richard42:
littleguy77, you're right.  OSAL is for Operating System abstraction, whereas this code is for handling all the ARM cpu variants.  Old dynarec will never build on ARM, so it doesn't need it.  And the ARM build of the new dynarec will probably never be built under anything but Linux.  So it makes sense to leave it in place.  However this PR now has conflicts with the head.  Can someone rebase this?

twinaphex:
> Just a matter of habit I guess, strings are easier to manipulate in C++ IMO

I do consider it a dealbreaker. The core code needs to be absolutely 100% C code. C++ code is not acceptable at all.

I'll be forced to rewrite this in case it gets left in this state.

Narann:
I don't want to create any flamewar here but yeah:
* The whole core is in C, I know C++ is _compatible_ but formally, it's not the same language anymore (integrating a lot of different paradigm than C).
* Integrating C++ code for a tiny feature here (file reading) will legitimate C++ use later for a bigger feature.
* From my perspective, if a C++ project does not have standards and patterns (implying a design philosophy) it quickly became a mess because C++ is used in an inconsistent way (C++ is the wider language ever created). In my situation Rice was (is actually still) totally like this. Write in C then evolve in C++.
* Reading the code, it's not something complex to write in C, It would be sad to don't write this parsing code in C.

Don't get me wrong, I don't blame your work @Gillou68310, I just argue for consistency and will try to help you to change this. The core has similar code in [ConfigInit](https://github.com/mupen64plus/mupen64plus-core/blob/master/src/api/config.c#L421) and [write_configlist_file](https://github.com/mupen64plus/mupen64plus-core/blob/master/src/api/config.c#L358).

src/r4300/new_dynarec/arm_cpu_features.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    std::string line;
+    std::ifstream file(procfile);
+
+    if (!file)
+        return 0;
+
+    while (std::getline(file, line))
+    {
+        if (line.find(marker) != std::string::npos)
```
Narann: Find a string into another string: [strstr](http://www.cplusplus.com/reference/cstring/strstr/)

Narann:
Maybe we could put [this](https://github.com/mupen64plus/mupen64plus-core/blob/master/src/api/config.c#L473) input [util.c](https://github.com/mupen64plus/mupen64plus-core/blob/58022568ee95da01d40141bfbc70e8ee8bf4911d/src/main/util.c) under `read_whole_file(char* filepath)`. And also a `find_next(char* str, char c)` returning the number of char before the given pointer and the next char _c_.

This way we could do the feature check like this:

```c
const char *buf = read_whole_file("/proc/cpuinfo");
char *pch = strstr(buf, "Feature\t:")

if(pch){
    token = strtok(pch," ");
    while(token != NULL)
        if(token == feature)
            return 1
        token = strtok(NULL," ");
    return 0;
}
```
I'm not a C expert but I'm sure it's not so complicate.

src/r4300/new_dynarec/arm_cpu_features.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include <string>
+#include "arm_cpu_features.h"
+
+extern "C" {
+    #include "api/callbacks.h"
+}
+
+arm_cpu_features_t arm_cpu_features;
+
+const char procfile[] = "/proc/cpuinfo";
```
Narann: lol, I just realize: How this will work for Windows? (Because Windows 10 supports ARM).

Nebuleon:
@TwinAphex - What is your main issue with having C++ code in the Core? Is it that C++ code requires linking with the C++ standard library, or code maintenance concerns (slippery slope, "everyone will use nested template<>s everywhere"), or something else?

richard42:
Actually there is already C++ code in the core: the OGLFT module used for the on-screen display is C++.  That's why I didn't complain too much.  It would be nice if the core library had no C++ in it though, just to limit the dependencies.  I've found that string parsing sucks in many different languages; I prefer C string parsing over C++ and Objective C.

Gillou68310:
Update: Parse cpuinfo in C instead of C++

Gillou68310:
Update: rebase against master

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+static unsigned char check_arm_cpu_feature(const char* feature)
+{
+    unsigned char status = 0;
+    FILE *pFile = fopen (procfile, "r");
+    
+    if (pFile != NULL)
+    {
+        char line[1024];
+        
+        while (fgets (line , sizeof(line) , pFile) != NULL)
```
Narann: Remove space between `fgets` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+        {
+            if (strncmp (line, "Features\t: ", 11))
```
Narann: Remove space between `strncmp` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+                continue;
+
+            if (strstr (line + 11, feature) != NULL)
```
Narann: Remove space between `strstr` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+                status = 1;
+            
+            break;
+        }
+        fclose (pFile);
```
Narann: Remove space between `fclose` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+#include "api/callbacks.h"
+
+arm_cpu_features_t arm_cpu_features;
+
+const char procfile[] = "/proc/cpuinfo";
+
+static unsigned char check_arm_cpu_feature(const char* feature)
+{
+    unsigned char status = 0;
+    FILE *pFile = fopen (procfile, "r");
```
Narann: Remove space between `fopen` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+        }
+        fclose (pFile);
+    }
+    return status;
+}
+
+static unsigned char get_arm_cpu_implementer(void)
+{
+    int implementer = 0;
+    FILE *pFile = fopen (procfile, "r");
```
Narann: Remove space between `fopen` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+    
+    if (pFile != NULL)
+    {
+        char line[1024];
+        
+        while (fgets (line , sizeof(line) , pFile) != NULL)
```
Narann: Remove space between `fgets` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+        {
+            if (strncmp (line, "CPU implementer\t: ", 18))
```
Narann: Remove space between `strncmp` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+                continue;
+            
+            sscanf(line+18, "0x%02hhx", &implementer);
+            
+            break;
+        }
+        fclose (pFile);
```
Narann: Remove space between `fclose` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+    }
+    return (unsigned char)implementer;
+}
+
+static unsigned short get_arm_cpu_part(void)
+{
+    int part = 0;
+    FILE *pFile = fopen (procfile, "r");
```
Narann: Remove space between `fopen` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+    
+    if (pFile != NULL)
+    {
+        char line[1024];
+        
+        while (fgets (line , sizeof(line) , pFile) != NULL)
```
Narann: Remove space between `fgets` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+        {
+            if (strncmp (line, "CPU part\t: ", 11))
```
Narann: Remove space between `strncmp` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+                continue;
+            
+            sscanf(line+11, "0x%03hx", &part);
+            
+            break;
+        }
+        fclose (pFile);
```
Narann: Remove space between `fclose` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+    }
+    return (unsigned short)part;
+}
+
+void detect_arm_cpu_features(void)
+{
+    arm_cpu_features.SWP      = check_arm_cpu_feature("swp");
```
Narann: So the file is open for each arm cpu feature? Meaning opening the file each time?

What about passing an array of char*? I just ask.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+        arm_cpu_features.IDIVa = arm_cpu_features.IDIVt = 1;
+}
+
+void print_arm_cpu_features(void)
+{
+    char buffer[1024];
+
+    strcpy (buffer, "ARM CPU Features:");
+
+    if (arm_cpu_features.SWP)      strcat (buffer, " SWP");
```
Narann: Remove space between `strcat` and `(`.

src/r4300/new_dynarec/arm_cpu_features.c
```diff
@@ -0,0 +1,146 @@
...
+    // Qualcomm Krait supports IDIVa but it doesn't report it. Check for krait.
+    if (get_arm_cpu_implementer() == 0x51 && get_arm_cpu_part() == 0x6F)
+        arm_cpu_features.IDIVa = arm_cpu_features.IDIVt = 1;
+}
+
+void print_arm_cpu_features(void)
+{
+    char buffer[1024];
+
+    strcpy (buffer, "ARM CPU Features:");
```
Narann: Remove space between `strcpy` and `(`.

src/r4300/new_dynarec/assem_arm.c
```diff
@@ -2053,14 +2053,11 @@ static void emit_writebyte(int rt, int addr)
...
 }
 
-/*
-static void emit_mul(int rs)
+static void emit_mul(u_int rs1,u_int rs2,u_int rt)
 {
-  assem_debug("mul %%%s",regname[rs]);
-  assert(0);
+  assem_debug("mul %s,%s,%s",regname[rt],regname[rs1],regname[rs2]);
+  output_w32(0xe0000090|(rt<<16)|(rs2<<8)|rs1);
```
Narann: Simple question for my own culture: What `0xe0000090` is supposed to be?

Narann:
Good job, the code is very nice (except space everywhere in function calls).

src/r4300/new_dynarec/assem_arm.c
Gillou68310: It's the opcode of the MUL instruction ;-)

If you write this in assembly:
```mul r1,r2,r3 //r1 = r2*r3``` 

The compiler will generate the following binary code:
```0xe0010392```

Gillou68310:
Removed spaces between ```function_name``` and ```(```

src/r4300/new_dynarec/arm_cpu_features.c
Gillou68310: I know it's not quite optimized, but considering this is done only one time at startup I think it's not worth the effort ;-)

Narann: Ok. :)

src/r4300/new_dynarec/assem_arm.c
Narann: So it's not supposed to change.

Do you mean no one have still think to use something like `#define ARMI_MUL 0xe0000090` or `X86I_MUL` for instruction opcode? Any cons about using define for such thing?

That's easy to do and would make the code even cleaner. Is this relevant to create a ticket for this? :)

C experts, any advise?

Narann: I realize all those "magic numbers" are obvious when you read the assem_debug comment.

Gillou68310: Yes it is ;-) that's why I think using defines for opcodes won't really help making the code cleaner.

Narann: Yes, my bad. :)

PR #78, opened 2015-02-15
Copy and byte-swap incoming N64 ROMs in one step

Nebuleon:
main/rom.c's open_rom function does memcpy from the source buffer into the Core's buffer, and then swap_rom on the Core's buffer. With .z64 and .n64 input ROMs, the second pass has to read each word back after the memcpy, and they were likely kicked out of the data cache.

This pull request proposes a one-step operation that copies and swaps bytes from the source buffer into the Core's buffer.

### DIFF
--- a/src/main/rom.c
+++ b/src/main/rom.c
@@ -21,6 +21,7 @@
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
 #include <stdio.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
 #include <ctype.h>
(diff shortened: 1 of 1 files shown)


src/main/rom.c
```diff
@@ -81,42 +82,51 @@ static int is_valid_rom(const unsigned char *buffer)
...
+        /* .v64 images have byte-swapped half-words (16-bit). */
+        size_t i;
+        const uint16_t* src16 = (const uint16_t*) src;
+        uint16_t* dst16 = (uint16_t*) dst;
+        for (i = 0; i < len; i += 2)
         {
+            *dst16++ = m64p_swap16(*src16++);
+        }
+    }
+    else if (*(const uint8_t*) src == 0x40)
```
Narann: Please, let the comment that explain what `0x40` is supposed to check: `/* Wordswap if .n64 image. */`

src/main/rom.c
```diff
@@ -81,42 +82,51 @@ static int is_valid_rom(const unsigned char *buffer)
...
-static void swap_rom(unsigned char* localrom, unsigned char* imagetype, int loadlength)
+static void swap_copy_rom(void* dst, const void* src, size_t len, unsigned char* imagetype)
 {
-    unsigned char temp;
-    int i;
-
-    /* Btyeswap if .v64 image. */
-    if(localrom[0]==0x37)
-        {
+    if (*(const uint8_t*) src == 0x37)
```
Narann: Same here, let the comment to explain what `0x37` is supposed to check: `/* Btyeswap if .v64 image. */`

src/main/rom.c
```diff
@@ -81,42 +82,51 @@ static int is_valid_rom(const unsigned char *buffer)
...
-        for (i = 0; i < loadlength; i+=2)
-            {
-            temp=localrom[i];
-            localrom[i]=localrom[i+1];
-            localrom[i+1]=temp;
-            }
-        }
-    /* Wordswap if .n64 image. */
-    else if(localrom[0]==0x40)
+        /* .v64 images have byte-swapped half-words (16-bit). */
```
Nebuleon: The .v64 comment is now here.

src/main/rom.c
```diff
@@ -81,42 +82,51 @@ static int is_valid_rom(const unsigned char *buffer)
...
-        for (i = 0; i < loadlength; i+=4)
-            {
-            temp=localrom[i];
-            localrom[i]=localrom[i+3];
-            localrom[i+3]=temp;
-            temp=localrom[i+1];
-            localrom[i+1]=localrom[i+2];
-            localrom[i+2]=temp;
-            }
+        /* .n64 images have byte-swapped words (32-bit). */
```
Nebuleon: The .n64 comment is now here.

src/main/rom.c
Narann: As you wish. It's just I'm not a fan of floating magic numbers without clear explainations of their hint.

But it's still ok in your case. :)

Nebuleon:
Then what about changing the code to use something like `*(const uint8_t*) src == V64_SIGNATURE[0]`? And changing [is_valid_rom](https://github.com/mupen64plus/mupen64plus-core/blob/44f1424/src/main/rom.c#L68-L82) to use `V64_SIGNATURE`, `N64_SIGNATURE` and `Z64_SIGNATURE`?

littleguy77:
That makes too much sense ;P

Nebuleon:
But isn't that the point? ;)

I'll cancel this pull request and make another with better code later.

Narann:
I like this. What about you @bsmiles32?

PR #80, opened 2015-02-17
Copy and byte-swap incoming N64 ROMs in one step (version 2)

Nebuleon:
*This pull request is a new version of the defunct pull request #78, with more changes as suggested by @Narann to remove magic constants from the code. I don't think I could switch the branch on the old pull request once it was made...*

main/rom.c's open_rom function does memcpy from the source buffer into the Core's buffer, and then swap_rom on the Core's buffer. With .z64 and .n64 input ROMs, the second pass has to read each word back after the memcpy, and they were likely kicked out of the data cache.

This pull request proposes a one-step operation that copies and swaps bytes from the source buffer into the Core's buffer. It also uses named 4-byte signature arrays to detect an image's endianness.

### DIFF
--- a/src/main/rom.c
+++ b/src/main/rom.c
@@ -20,6 +20,7 @@
  *   51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.          *
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


src/main/rom.c
```diff
@@ -65,58 +66,66 @@ static m64p_system_type rom_country_code_to_system_type(unsigned short country_c
...
-    if((buffer[0]==0x80)&&(buffer[1]==0x37)&&(buffer[2]==0x12)&&(buffer[3]==0x40))
-        return 1;
-    /* Test if rom is a byteswapped .v64 image with header 0x37804012. [BADC] */
-    else if((buffer[0]==0x37)&&(buffer[1]==0x80)&&(buffer[2]==0x40)&&(buffer[3]==0x12))
-        return 1;
-    /* Test if rom is a wordswapped .n64 image with header  0x40123780. [DCBA] */
-    else if((buffer[0]==0x40)&&(buffer[1]==0x12)&&(buffer[2]==0x37)&&(buffer[3]==0x80))
+    if (memcmp(buffer, Z64_SIGNATURE, sizeof(Z64_SIGNATURE)) == 0
+     || memcmp(buffer, V64_SIGNATURE, sizeof(V64_SIGNATURE)) == 0
+     || memcmp(buffer, N64_SIGNATURE, sizeof(N64_SIGNATURE)) == 0)
```
Narann: Nice!

Narann:
The code look definitely more clear!

PR #82, opened 2015-02-21
Debugger enhancements

Nebuleon:
I propose two commits to enhance the consistency of the debugger's MIPS disassembler.

First, I add a `$` before the names of all integer registers, not just `$zero`, `$at`, `$gp`, `$sp` and `$ra`.

Second, I split opcodes from their operands by tabs as well as spaces in `r4k_disassemble_split`, because for some reason, the disassembler writes Coprocessor 1 (FPU) opcodes separated from their operands with `'\t'`. So instead of `"mtc1\t\t$a0,$f0"` being parsed as the opcode `"mtc1\t\t$a0,$f0"` with no operands, it now gets correctly parsed as an `mtc1` with `$a0,$f0` as its operands.

Small, but (IMHO) useful.

### DIFF
--- a/src/debugger/dbg_decoder.c
+++ b/src/debugger/dbg_decoder.c
@@ -149,10 +149,10 @@ static const char * const r4k_str_fmt_name[16] =
 
 static const char * const r4k_str_reg_name[32] = 
 {
-	"$zero", "$at",	"v0",	"v1",	"a0",	"a1",	"a2",	"a3",
-	"t0",	"t1",	"t2",	"t3",	"t4",	"t5",	"t6",	"t7",
-	"s0",	"s1",	"s2",	"s3",	"s4",	"s5",	"s6",	"s7",
-	"t8",	"t9",	"k0",	"k1",	"$gp",	"$sp",	"s8",	"$ra"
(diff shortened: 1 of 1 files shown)


Narann:
That seems legit. I don't really know how the core debugger works. I'm just concern about potential side effect for other peoples using debugger (even if I'm not sure a lot of peoples do so.

I will let others merge this is they thinks it's ok but I don't see any problem with this. :+1:

wareya:
Can you make it so that booting up the emulator in debugger mode doesn't involve calling a function, the reverse of that function, and then unpausing the emulator, for no apparent reason at all? I have to CoreDoCommand(M64CMD_RESUME, 0, NULL) and then DebugSetRunState(2) and then DebugStep(). It's insane.

Nebuleon:
That does seem insane, yes, but it's not in the scope of this pull request at all - it seems more fit for a discussion on its own about the API for the debugger on the Google Group for mupen64plus.

I needed the disassembler for something, and the Coprocessor 1 opcodes being separated by tabs and the rest by spaces was making a mess of my disassembly output. I don't know how the rest of the debugger works, either, both in the implementation (which code does what?) and activation (how do I use it in a frontend? in ui-console, what command-line arguments do I pass?) senses.

PR #83, opened 2015-02-22
R4300 isolation

bsmiles32:
Following the discussion from [1], I propose some changes to decouple (at least partially) the r4300 subsystem from the rest of the core.
The idea behind these changes is to allow later to fully decouple new_dynarec / hacktarux r4300 implementations, and maybe allow for a 3rd "from scratch" r4300 implementation if necessary.
Please test especially on 32bit and new_dynarec setups.

[1] https://groups.google.com/forum/#!topic/mupen64plus/4rOO7cX0toY


src/r4300/r4300_core.h
```diff
@@ -31,4 +37,30 @@ struct r4300_core
...
+int64_t* r4300_regs(void);
+int64_t* r4300_mult_hi(void);
+int64_t* r4300_mult_lo(void);
+unsigned int* r4300_llbit(void);
+uint32_t* r4300_pc(void);
+
+uint32_t* r4300_last_addr(void);
+unsigned int* r4300_next_interrupt(void);
+
+unsigned int get_r4300_emumode(void);
```
Narann: Why not `r4300_get_emumode()` to keep consistency in the function names? From what I've seen in the code every `r4300` related function start with `r4300`

Also this can be useful for modern IDE to find any `r4300` related function just typing them.

src/r4300/r4300_core.h
```diff
@@ -31,4 +37,30 @@ struct r4300_core
...
+
+/* Allow cached/dynarec r4300 implementations to invalidate
+ * their cached code at [address, address+size]
+ *
+ * If size == 0, r4300 implementation should invalidate
+ * all cached code.
+ */
+void invalidate_r4300_cached_code(uint32_t address, size_t size);
```
Narann: Same here, why not use `r4300_invalidate_cached_code()`?

src/rdp/fb.c
```diff
@@ -137,8 +137,7 @@ void protect_framebuffers(struct rdp_core* dp)
                 {
                     fb->once = 0;
                     fast_memory = 0;
-                    for (j=0; j<0x100000; j++)
-                        invalid_code[j] = 1;
+                    invalidate_r4300_cached_code(0, 0);
```
Narann: `r4300_invalidate_cached_code()`?

src/r4300/r4300_core.h
bsmiles32: The rule I used to name my functions here is a little bit more complicated than that :
- for "normal" functions I put the action verb first (-> get_r4300_emumode, write_ai_regs, invalidate_cached_code ...)
- for "thin" RW wrappers that just return a pointer to a variable I did not put any verb that's why you get the r4300 first (r4300_regs, r4300_mult_hi, ...). Since these are used to read AND write the pointed variables (the returned pointer is not const) I felt like putting a get_ verb prefix would be somewhat misleading (at least wrt the get/set pattern). And I don't really want to write both getter/setter functions for these : a single function which returns a pointer is good enough here.

That's the somewhat rationale behind the function names. I hope it makes sense.

Nebuleon:
I was going to do something like this after I was done with #84, because the memory accessors having duplicate versions of `invalid_code` checks would have prevented my progress with other code. But I guess that was in the scope of your proposal from the Google Group.

Speaking of #84, `src/r4300_core.c` now has code of the form ...

```C
int64_t* r4300_something()
{
    return (int64_t*)&something;
}
```

... because you wanted to do the right thing and return `int64_t*` from a new function, but #84 didn't exist when you made this code. It would be useful to remove the casts in another commit after both pull requests are merged.

The Hacktarux JIT/amd64 seems to work as well as before. :+1: I can only assume the Hacktarux JIT/x86 will also work, because it uses the same invalidation logic.

I still can't build the New Dynarec on amd64, even with `BITS=32`.

src/r4300/r4300_core.c
```diff
@@ -21,8 +21,89 @@
...
+
+void invalidate_r4300_cached_code(uint32_t address, size_t size)
+{
+    if (r4300emu != CORE_PURE_INTERPRETER)
+    {
+#ifdef NEW_DYNAREC
+        invalidate_cached_code_new_dynarec(address, size);
+#else
+        invalidate_cached_code_hacktarux(address, size);
+#endif
```
Nebuleon: Does this correctly deal with the case where the user is using the Cached Interpreter but the New Dynarec is compiled in?

It seems there's a missing check for `r4300emu` beyond the `#ifdef`s.

bsmiles32: Nice catch ! Will fix it after your PR gets merged.

bsmiles32:
> I still can't build the New Dynarec on amd64, even with BITS=32.

Is it because of my PR or because you don't have {32bit,ARM} cross toolchain ?

> It would be useful to remove the casts in another commit after both pull requests are merged.

I hope your PR gets merged first, so I can rebase mine and fix things directly in the right commits, so that every commit is buildable and bug-free.

> (0, 0) for the entire cache, otherwise it's (start_addr, len)?

Yes, that's the idea. Since each r4300 implementation {ie hacktarux / new_dynarec} could use different cache sizes I preferred to encode the "full range" with a specific value (size == 0). But if that's not good I can rework that eventually.

> What happens if a game does an uncached write and executes the code from normal cached addresses?

I guess there is a misunderstanding, probably because of the name I used for the function.
invalidate_r4300_cached_code is intended to let r4300 implementations be notified when a write occurs at a virtual address which could contain r4300 instructions. This allow dynarecs and the cached interpreter to invalidate whatever code they have "cached/compiled" for later execution. It is in no way related to the "cached"/"uncached" accesses that MIPS r4300 can do depending on the highest bits of a virtual address (0x80000000 vs 0xa000000 for instance). Furthermore, from my understanding m64p doesn't emulate cached vs uncached accesses.

Nebuleon:
> > I still can't build the New Dynarec on amd64, even with BITS=32.
> Is it because of my PR or because you don't have {32bit,ARM} cross toolchain ?

I don't have an ARM device to test with, and for some reason my 32-bit toolchain doesn't recognise `zconf.h` from zlib and fails building a few files in.

> > It would be useful to remove the casts in another commit after both pull requests are merged.
> I hope your PR gets merged first, so I can rebase mine and fix things directly in the right commits, so that every commit is buildable and bug-free.

Fair enough, I guess. Either order would yield correct and bug-free builds if you have the cast, but if you don't want the cast to be there at all, #84 should be merged first.

> > (0, 0) for the entire cache, otherwise it's (start_addr, len)?
> Yes, that's the idea. [...]

Alright. That's good, actually. :)

> > What happens if a game does an uncached write and executes the code from normal cached addresses?
> [...] It is in no way related to the "cached"/"uncached" accesses that MIPS r4300 can do depending on the highest bits of a virtual address (0x80000000 vs 0xa000000 for instance). Furthermore, from my understanding m64p doesn't emulate cached vs uncached accesses.

That's what I was asking about.

According to some `DebugMessage` calls I placed in `invalidate_r4300_cached_code`, the 8 and A (`0x8???_????` and `0xA???_????`) versions of RDRAM addresses are constantly invalidated in quick succession in Conker's Bad Fur Day. I don't know what causes this. But is the plan to ask R4300 emulators to invalidate the 8 and A versions separately, or to tell R4300 emulators about the accesses that actually occur (8 or A) and let them with the other version (A or 8)?

With current R4300 emulators this doesn't matter, because `jump_to_func` has pretty elaborate code to check on every jump if the target page or its cached/uncached/physical (if virtual) doppelganger is invalid, and then plaster all versions' pages with `NOTCOMPILED`. But future cores might not have the same logic to detect this.

The 8 and A versions of RAM currently have different code blocks (and Cached Interpreter pages) created for them, as much as the 8 and 1 versions have different code created for them in Conker's Bad Fur Day. In the case of virtual code addresses, it matters, because `0x1000_0000 | (addr & 0x0FFF_FFFF)` is not necessarily mapped at `0x8000_0000 | (addr & 0x0FFF_FFFF)`, so `J` and `JAL` instructions mean different things if executed from physical or virtual addresses. But in the case of 8 and A, the base address is the same, except for the special boot address, `0xA400_0040`.

Perhaps the native code cache should simply ignore the A version of physical addresses, and just make an 8 version? That is, if there is no plan for Mupen to **ever** make a distinction between cached and uncached accesses.

bsmiles32:
I think I finally understand what you mean :)

I think invalidate_r4300_cached_code should be called only with the virtual address that originated the access. Each r4300 implementation will be responsible to translate that address into meaningful invalidation strategy.

Basically it means that in pi_controller only the 0xA???-???? invalidation call will stay.
The other (0x8???-????) will have to be done at the discretion of each r4300 implementation if desired.

If it is to be part of this PR, I will do it in a separate commit.

Nebuleon:
Ah, so it is indeed the PI controller doing this! [I see the code now.](https://github.com/mupen64plus/mupen64plus-core/blob/e1639f91f96db8344c425226e0d4a3c6178b54b4/src/pi/pi_controller.c#L127-L128) Thanks! :)

All the emulators check the `invalid_code` array for the cached and uncached doppelgangers at the moment, so you could delete the separate invalidation for the `0x8???_????` version in the PI controller in a separate commit.

What's more complex though, and the reason I wouldn't put that in this pull request, is that `write_nomem` (and the byte, 16-bit and 64-bit access versions) invalidates the virtual version of the address it writes, then relies on the accessor for the resolved physical address to invalidate the physical version, and there's `CHECK_MEMORY` in [the `SW` opcode](https://github.com/mupen64plus/mupen64plus-core/blob/e1639f91f96db8344c425226e0d4a3c6178b54b4/src/r4300/interpreter_r4300.def#L443-L452) that could mix you up.

bsmiles32:
Update: rebased against master + fixed invalidate_r4300_cached_code.

Nebuleon:
Alright, now that's a much smaller diff, because the commits don't have to be R4300 isolation and intN_t changes all at once, and you don't have to use casts in `r4300_???()` :+1:

I do see that you changed the FPU control registers to be `uint32_t` instead of `int32_t`, which makes sense. [`CFC1` had sign-extensions around `FCR0` and `FCR31`](https://github.com/mupen64plus/mupen64plus-core/blob/e1639f91f96db8344c425226e0d4a3c6178b54b4/src/r4300/interpreter_cop1.def#L47-L56) to make sure the sign extension was applied when reading it back into a GPR, so I wonder what I was concerned about when using `int32_t`... Anyway, this is pretty awesome now!

littleguy77:
Sorry this slipped off my radar.  Do you guys still need this tested on new_dynarec and/or ARM?  I can probably get to it tonight.

Nebuleon:
Yes please, that would be nice. Do test using the Cached Interpreter on a build with the New Dynarec compiled in, just to see if there are other errors that I didn't catch, and the New Dynarec itself on ARM.

littleguy77:
No apparent issues in mupen64plus-ae.  :+1:   No build issues for ARM/new_dynarec nor x86/new_dynarec.  No apparent runtime issues with ARM/new_dynarec and ARM/cached-interpreter.  Did not test to test x86/new_dynarec (no test device).

PR #84, opened 2015-02-23
Use [u]intN_t for Nintendo 64 data

Nebuleon:
The Nintendo 64 uses various data with exact bit counts (8-bit, 16-bit, 32-bit, 64-bit). `char`, `unsigned char`, `short`, `unsigned short`, `int`, `unsigned int`, `long`, `unsigned long`, `long long` and `unsigned long long` may be the required number of bits, but this is not guaranteed. This patch series uses `<stdint.h>` to provide types with exact bit counts for variables holding various types of Nintendo 64 data instead of the types above:

* integer registers and Coprocessors 0 and 1 registers (`int64_t`, `int32_t`);
* addresses (`uint32_t`);
* MIPS III opcodes (`uint32_t`) and instruction fields (`int16_t`, `uint32_t`);
* memory accesses (`uint8_t`, `uint16_t`, `uint32_t`, `uint64_t`);
* saved states;
* the byte-swap utility functions (`uint16_t`, `uint32_t`, `uint64_t`);
* the ROM header (`uint8_t`, `uint16_t`, `uint32_t`).

This follows in the footsteps of recent code @bsmiles32 has worked on, which has been using these types. It should not modify mupen64plus's behavior on platforms where the size of an `int`-based type matches the new intN_t type for it, only further cement the bit widths for those variables that need them.

To print out these intN_t quantities, I have made a fixup commit to use `PRIX8`, `PRIX16` and `PRIX32` from `<inttypes.h>` where appropriate.

This patch series starts creating 32-bit and 64-bit constants with **`INT32_C()`** and **`INT64_C()`**. Not every constant is like that now, but the macros append the proper suffix for constants with exact bit counts on the architecture being compiled for. I mainly used the macros for bitwise operations and to initialise the values of registers with exact bit counts, except the value `0`.

The Hacktarux JIT/x86, Hacktarux JIT/amd64 and New Dynarec/ARM are not modified, due to `unsigned int` and `unsigned long long` having the appropriate widths for their platforms already. There may be a problem with 32-bit-on-amd64 builds, due to the size of pointers differing from the size of ints and longs; this could be fixed up by the use of `uintptr_t`. This is left unhandled in my intN_t branch.

I'd like some feedback particularly related to build successes or failures on Windows, 32-bit-on-amd64 and ARM, and the use of `INTn_C()`. If things break, I'll try my best to remake the patches.

### DIFF
--- a/src/main/cheat.c
+++ b/src/main/cheat.c
@@ -469,7 +469,7 @@ static int cheat_parse_hacks_code(char *code, m64p_cheat_code **hack)
     while ((token = strtok_compat(input, ",", &saveptr))) {
         input = NULL;
 
-        ret = sscanf(token, "%08" SCNX32 " %04X", &hackbuf[num_codes].address,
+        ret = sscanf(token, "%08" SCNx32 " %04X", &hackbuf[num_codes].address,
                      &hackbuf[num_codes].value);
         if (ret == 2)
             num_codes++;
--- a/src/main/cheat.c
+++ b/src/main/cheat.c
@@ -39,6 +39,9 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <stdint.h>
+#define __STDC_FORMAT_MACROS
+#include <inttypes.h>
 
 // local definitions
 #define CHEAT_CODE_MAGIC_VALUE 0xDEAD0000
@@ -466,7 +469,7 @@ static int cheat_parse_hacks_code(char *code, m64p_cheat_code **hack)
     while ((token = strtok_compat(input, ",", &saveptr))) {
         input = NULL;
 
-        ret = sscanf(token, "%08X %04X", &hackbuf[num_codes].address,
+        ret = sscanf(token, "%08" SCNX32 " %04X", &hackbuf[num_codes].address,
                      &hackbuf[num_codes].value);
         if (ret == 2)
             num_codes++;
--- a/src/main/rom.c
+++ b/src/main/rom.c
@@ -25,6 +25,9 @@
 #include <string.h>
 #include <ctype.h>
 #include <stddef.h>
+#include <stdint.h>
+#define __STDC_FORMAT_MACROS
+#include <inttypes.h>
 
 #define M64P_CORE_PROTOTYPES 1
 #include "api/m64p_types.h"
@@ -61,7 +64,7 @@ m64p_rom_header   ROM_HEADER;
 rom_params        ROM_PARAMS;
 m64p_rom_settings ROM_SETTINGS;
 
-static m64p_system_type rom_country_code_to_system_type(unsigned short country_code);
+static m64p_system_type rom_country_code_to_system_type(uint16_t country_code);
 static int rom_system_type_to_ai_dac_rate(m64p_system_type system_type);
 static int rom_system_type_to_vi_limit(m64p_system_type system_type);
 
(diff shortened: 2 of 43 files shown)


Nebuleon:
I wasn't aware of #83 when I finished this. This pull request will clash pretty hard with it:

* In `main/savestates.c`, the Coprocessor 0 commit will clash with bsmiles32/mupen64plus-core@6517df9
* In `main/savestates.c`, the fix for the New Dynarec's use of `last_addr` in the `uint32_t` jump address commit will clash with bsmiles32/mupen64plus-core@1e8843e
* In `main/savestates.c`, the Coprocessor 1 commit will clash with bsmiles32/mupen64plus-core@48a06de

Possibly more. I'll follow the status of #83.

bsmiles32:
Congratulations for working on such an important task ! I've always wanted to get rid of the non fixed sized integers but always got demotivated in the process. So this PR is very welcome.
I will test it later, but if it works for everybody, this PR should be merged before my r4300_isolation PR.
Once that's done I'll update my PR and fix the mistake in invalidate_cached_code :)

src/main/cheat.c
```diff
@@ -466,7 +469,7 @@ static int cheat_parse_hacks_code(char *code, m64p_cheat_code **hack)
     while ((token = strtok_compat(input, ",", &saveptr))) {
         input = NULL;
 
-        ret = sscanf(token, "%08X %04X", &hackbuf[num_codes].address,
+        ret = sscanf(token, "%08" SCNx32 " %04X", &hackbuf[num_codes].address,
```
Narann: No more `X`?

src/main/rom.c
```diff
@@ -202,19 +205,19 @@ m64p_error open_rom(const unsigned char* romimage, unsigned int size)
     DebugMessage(M64MSG_INFO, "Name: %s", ROM_HEADER.Name);
     imagestring(imagetype, buffer);
     DebugMessage(M64MSG_INFO, "MD5: %s", ROM_SETTINGS.MD5);
-    DebugMessage(M64MSG_INFO, "CRC: %x %x", sl(ROM_HEADER.CRC1), sl(ROM_HEADER.CRC2));
+    DebugMessage(M64MSG_INFO, "CRC: %" PRIX32 " %" PRIX32, sl(ROM_HEADER.CRC1), sl(ROM_HEADER.CRC2));
```
Narann: Only `%`? I didn't know this was legit.

src/main/cheat.c
wareya: I think that expands to "%08lx %04X".

sscanf is x vs X agnostic, yes?

Narann:
Wow, this job is such massive, thanks Nebulon!

I can't test by myself but if anyone confirm it work. I will merge it. I'm just concerned about all those _orphan_ `%`.

Nebuleon:
Yeah, I didn't know the formatting on GitHub would be so unfortunate, but this is how it's supposed to work with `<inttypes.h>`.

For example,
```C
#include <stdint.h>
#define __STDC_FORMAT_MACROS
#include <inttypes.h>

printf("%08" PRIX32 " %04" PRIX16 "\n", address, value);
```
is preprocessed into
```C
printf("%08" "X" " %04" "X" "\n", address, value);
```
which the compiler considers to be a single string.

src/main/cheat.c
Nebuleon: Yeah, a `scanf` format string considers lower-case and upper-case hex input to fit both the `"%x"` and the `"%X"` format. In addition, there's not even a `SCNX32` macro defined in `<inttypes.h>` in the standard, only `SCNx32`.

PR #47, opened 2014-11-23
[WIP] Refactor memory

bsmiles32:
This is the first part of the memory subsystem refactoring [1].

The goal of this set of commits is twofold:
-first remove the duplication that exists between byte,hword,word and dword accesses to memory
-second remove some global variables associated with memory accesses (address, rdword, cpu_byte, hword, word, dword)

Issues:
- break the new_dynarec (If someone more knowledgeable than me can look what to change to make it work again)
- use interpreted functions for memory accesses in the dynarec (I didn't know how to rewrite them, so I need help here).

I'd like to hear about any regression (behavior/performance), especially from downstream since I can't really test that.

cc: @richard42, @fayvel, @mupen64plus-ae, @twinaphex

[1] I have some more commits waiting to be cleaned up in my branch refactoring_experimental :)
Those will come later.

PR #88, opened 2015-02-25
Simplify the unaligned memory access opcodes

Nebuleon:
I attach a commit that modifies `LWL`, `LWR`, `LDL`, `LDR`, `SWL`, `SWR`, `SDL` and `SDR` to compute masks and bit shift amounts at run time to load and store parts of 32-bit and 64-bit values, instead of having a giant `switch` statement, 3 or 7 branches of which have similar code with different constants.

The main advantages of this are:

* reduced code duplication and source line count;
* reduced code size (when compiled);
* reduced branching or indirect jumping (depending on how the compiler implements a 4- or 8-option `switch`, a series of conditional branches or a jump table).

But there are some disadvantages:

* the code is much more difficult to wrap one's head around when first reading it;
* there is more code to execute unconditionally to compute the masks and bit shifts.

The main game that uses these unaligned memory access opcodes, among those I test regularly, is Conker's Bad Fur Day. But the amount of those opcodes relative to all the opcodes it executes (as well as the fact that it uses virtual memory addresses) is very low, so I don't know how I'd go about testing their performance.

I suggest to use the Split difference view on GitHub Web to read this commit, because it's very confusing in the Unified view.

### DIFF
--- a/src/r4300/interpreter_r4300.def
+++ b/src/r4300/interpreter_r4300.def
@@ -94,8 +94,28 @@ DECLARE_INSTRUCTION(DADDIU)
    ADD_TO_PC(1);
 }
 
-// TODOXXX refactor the following functions to remove the
-// lsaddr and lsrpt locals. this may lead to a small speedup too
+/* Assists unaligned memory accessors with making masks to preserve or apply
+ * bits in registers and memory.
+ *
+ * BITS_BELOW_MASK32 and BITS_BELOW_MASK64 make masks where bits 0 to (x - 1)
+ * are set.
+ *
+ * BITS_ABOVE_MASK32 makes masks where bits x to 31 are set.
+ * BITS_ABOVE_MASK64 makes masks where bits x to 63 are set.
+ *
+ * e.g. x = 8
+ * 0000 0000 0000 0000 0000 0000 1111 1111 <- BITS_BELOW_MASK32(8)
+ * 1111 1111 1111 1111 1111 1111 0000 0000 <- BITS_ABOVE_MASK32(8)
+ *
+ * Giving a negative value or one that is >= the bit count of the mask results
+ * in undefined behavior.
+ */
+
+#define BITS_BELOW_MASK32(x) ((UINT32_C(1) << (x)) - 1)
+#define BITS_ABOVE_MASK32(x) (~((UINT32_C(1) << (x)) - 1))
+
+#define BITS_BELOW_MASK64(x) ((UINT64_C(1) << (x)) - 1)
+#define BITS_ABOVE_MASK64(x) (~((UINT64_C(1) << (x)) - 1))
 
 DECLARE_INSTRUCTION(LDL)
 {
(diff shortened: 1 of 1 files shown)


src/r4300/interpreter_r4300.def
```diff
@@ -548,92 +402,30 @@ DECLARE_INSTRUCTION(SDR)
    int64_t *lsrtp = &irt;
    uint64_t old_word = 0;
    ADD_TO_PC(1);
-   switch ((lsaddr) & 7)
+   address = lsaddr & UINT32_C(0xFFFFFFF8);
+   if ((lsaddr & 7) == 7)
+   {
+     cpu_dword = *lsrtp;
```
bsmiles32: if (address != 0) test missing ?

src/r4300/interpreter_r4300.def
```diff
@@ -642,48 +434,30 @@ DECLARE_INSTRUCTION(SWR)
    int64_t *lsrtp = &irt;
    uint64_t old_word = 0;
    ADD_TO_PC(1);
-   switch ((lsaddr) & 3)
+   address = lsaddr & UINT32_C(0xFFFFFFFC);
+   if ((lsaddr & 3) == 3)
+   {
+     cpu_word = (uint32_t) *lsrtp;
```
bsmiles32: same here

bsmiles32:
These are comments by just reading the source. I will test more carefully later.
I annotated 2 places where the "if (address != 0)" is missing with respect to the original version.
From a refactoring point of view these should be corrected.
But I think that this "if (address != 0)" test should be done also at other places where it is not done currently, in fact before every memory access. Otherwise we end up corrupting the registers values whereas the expected behaviour is to trigger a TLB miss.
I had similar issue when I tried to refactor that part in the stalled PR #47 (I strongly encourage you to look inside the interpreter_r4300.def file there).

Note : the Travis failure doesn't seem to be related to your PR (it failed with "Unable to locate package libsdl2-dev").

Otherwise, it's very nice to see you tackle challenging and key areas of the core even if you are a relatively new contributor. So congratulations and keep the good work :)

Nebuleon:
> I annotated 2 places where the "if (address != 0)" is missing with respect to the original version.
From a refactoring point of view these should be corrected.

> But I think that this "if (address != 0)" test should be done also at other places where it is not done currently, in fact before every memory access. Otherwise we end up corrupting the registers values whereas the expected behaviour is to trigger a TLB miss.

Well that's odd... I received emails about the line comments, but they're not in the conversation comments as other people's line comments have been in the past, and neither are they in the "Files changed" tab.

Anyway, yep, I'm aware that `address` becoming `0` indicates that there's an impending TLB Refill exception. It is made that way by `read_nomem` and `write_nomem`, the virtual memory accessor functions. In `L*` opcodes, that indicates `*rdword` (which is often *rt*) *has not been* modified and *must not* be sign-extended by `L*`, and in `S*` opcodes, that indicates memory was not touched.

I examined the code paths a few months ago, and found only one thing that was marginally wrong: `CHECK_MEMORY` runs even if `address` has become `0`. That's not so bad, though, because the page at `0x0 .. 0xFFF` will almost never contain code, and if it does, the worst thing that happens is that it gets invalidated. On the other hand, I have never seen the *rt* register get sign-extended or rewritten in an `L*` opcode if, after `read_???_in_memory()`, a TLB Refill exception had been signalled.

Ddid I really mess that up in `LWL` and `LWR`?

> I had similar issue when I tried to refactor that part in the stalled PR #47 (I strongly encourage you to look inside the interpreter_r4300.def file there).

In #47 you seem to have bitten off way more than you could chew. You tried to change the mechanism by which TLB Refill exceptions were reported, the memory accessor helpers (a single read function instead of `readmem[addr >> 16] ();`), and the ABI of memory accessors themselves - which broke all JITs -, then tried to eliminate global variables while not knowing what to do with `CHECK_MEMORY`.

4 months later or so, with #83 and its `invalidate_r4300_cached_code` function, as well as some memory refactorings merged already, perhaps it'd be easier now to make `CHECK_MEMORY` be a call to that instead, and you can pass `address` or a local variable, whatever you wish. It would then be easier to kill the global variable while keeping the `void (*) (void)` functions that the JITs use, as simple wrappers to your redone 3-parameter functions:
```C
void read_???()
{
  address = new_read_???(address, rdword, 4);
}
```

> Note : the Travis failure doesn't seem to be related to your PR (it failed with "Unable to locate package libsdl2-dev").

Yeah, Travis has been wharrgarbling today, it went in and out of maintenance after taking upwards of 3 hours to build some projects.

> Otherwise, it's very nice to see you tackle challenging and key areas of the core even if you are a relatively new contributor. So congratulations and keep the good work :)

Thanks!

Nebuleon:
*(The previous comment I had written here became completely irrelevant after @bsmiles32's edit, and I moved half of the post to move it into #47.)*

bsmiles32:
> Well that's odd... I received emails about the line comments, but they're not in the conversation
> comments as other people's line comments have been in the past, and neither are they in the "Files
> changed" tab.

I'm sorry for the confusion. I deleted the original inline notes after realizing that they were not adequate and resulted from me not reading carefully the code. Then I edited the global comment to remove what resulted from my misunderstanding. So yes, your code seems perfectly valid and much easier to read than it was before.

> Ddid I really mess that up in LWL and LWR?

No. I misread your code. I'm sorry.

> In #47 you seem to have bitten off way more than you could chew.

Yeah, I know... The problem is that almost everything is interdependant.
I will retry this memory refactoring later, and if I can find a way to make it more incremental I will.
But this is hard as it is deeply integrated into the various r4300 implementations.
If you have suggestions, I'm all ears :)

bsmiles32:
Tested and it should be good enough to merge: +1

Narann:
Travis fail but it seems to be a travis problem. It's a shame there is no way to restart it. I merge now. Thanks for contributing! :)

conchurnavid:
@Narann  The repo admin can press the button at the right top over the log to restart the build

conchurnavid:
Here is a picture of this "button"

restart_travis

Narann:
Thanks @conchurnavid, I don't see the button as (I guess) I'm not an admin of the travis of mupen64plus.

PR #90, opened 2015-03-02
Updated MSVC2013 project file + reorganized/renamed/removed some files

Gillou68310:


src/main/rom.c
```diff
@@ -103,11 +103,12 @@ static void swap_copy_rom(void* dst, const void* src, size_t len, unsigned char*
 {
     if (memcmp(src, V64_SIGNATURE, sizeof(V64_SIGNATURE)) == 0)
     {
-        *imagetype = V64IMAGE;
```
Narann: Why change position of `*imagetype = V64IMAGE;`?

src/main/rom.c
```diff
@@ -115,11 +116,12 @@ static void swap_copy_rom(void* dst, const void* src, size_t len, unsigned char*
     }
     else if (memcmp(src, N64_SIGNATURE, sizeof(N64_SIGNATURE)) == 0)
     {
-        *imagetype = N64IMAGE;
```
Narann: Why change position of `*imagetype = N64IMAGE;`?

Gillou68310: To stay C89 compliant ;-)

Narann: You got me! :P

Gillou68310:
Ok I think the commit message here is inappropriate I'm not trying to be fully C89 compliant, the real purpose of this commit is to keep compatiblity with older MSVC compilers. The only pending issue is the missing inttypes.h header file.

@richard42 are you sure you want to drop compatiblity with older MSVC compiler?

richard42:
I'm sure that I don't want to support 4 different versions of MSVC, and I don't think anybody else does either.  Plus the lack of C99 support in pre-2k10 versions was a real annoyance for us, and a disgrace for Microsoft.  So I think that consolidating on the 2013 compiler is the right move.

Gillou68310:
> I'm sure that I don't want to support 4 different versions of MSVC, and I don't think anybody else does either.

And I totally agree with you here. Actually I noticed that all .vcxproj fles are backward compatible with older MSVC version by just changing the Platform toolset option (aka compiler version). 

I made a small tool to change the platform toolset for all mupen64plus .vcxproj automatically here:
https://github.com/Gillou68310/PlatformToolset

> Plus the lack of C99 support in pre-2k10 versions was a real annoyance for us

Yeah that's a shame! Maybe it's time for me to ask my job's IT a licence for MSVC2013 professional ;-)

richard42:
Or find a friend at Microsoft to buy one for you at the company store. :)  Don't they make an Express version of MSVC2013 that you could use?

littleguy77:
The express edition doesn't support add-ins (some of which are very useful) or advanced tools like profiling.  I have heard that the new "community edition" is halfway between the express and pro versions in terms of features and is free.  Might be worth a look @Gillou68310 ...

Gillou68310:
Wow thanks @littleguy77 I just checked out the community edition and it's exactly the same as the professional edition except that it's free!!!

Issue #91, opened 2015-03-04
Fallthrough in switches of interpreter code


conchurnavid:
There are three defects in the coverity report which are about switch "case" without break and also no /* fall through */ comment.

There are 38098, 38099 and 38100.Does anyone know if the break is actually missing or just the comment is missing to point out that this is intentionally?

Nebuleon:
Without further information, I assume the report refers to things like [lines 304..309 of `pure_interp.c` as of commit 16b2793](https://github.com/mupen64plus/mupen64plus-core/blob/16b2793c4ba3d09d6231735de557b598e5eb5589/src/r4300/pure_interp.c#L304-L309), where multiple `case` statements use the `NI` (Not Implemented) interpreter function. Those are intentional, and there's not even any code between the `case` statements that could confuse the reader.

If there are others with a legitimately missing `break;` statement, please provide line numbers or a new commit like in #92. :)

conchurnavid:
What information is missing? The line number can be found in the coverity reports I mentioned. I don't have more informations than that. Here is is a copy of the line number from the reports:

38098: interpreter_cop0.def 30
38099: interpreter_cop0.def 28
38100: cic.c 54 (the default part of the switch)

Nebuleon:
I can view nothing, including the mapping between Coverity's error numbers and line numbers. To view them, I presume I have to join Coverity, and then be added to the project. Thus, the opening message of the bug report has no context for pretty much anyone other than yourself and whoever has already joined Coverity and been added to the project. I've received 13 rapid-fire invitation emails, but none of the emails had any information about the source of the invitation, so I marked them all as spam and trashed them - I guess one of them is for this, though. Try looking here, while signed out, and see if you can find anything that I don't: https://scan.coverity.com/projects/4381

Anyway, the new information refers to [lines 28..30 of `src/r4300/interpreter_cop0.def`](https://github.com/mupen64plus/mupen64plus-core/blob/f648835273a2128cd467e95ef69765ac18a3e93a/src/r4300/interpreter_cop0.def#L28-L30) and [line 54 of `src/si/cic.c` as of commit f648835](https://github.com/mupen64plus/mupen64plus-core/blob/f648835273a2128cd467e95ef69765ac18a3e93a/src/si/cic.c#L54-L55).

`interpreter_cop0.def`:
[...]

PR #93, opened 2015-03-05
Revert "Check if the size calculation for config file failed"

conchurnavid:
This fix for CID 38108 was actually introducing dead code and not fixing the
actual problem. The check for the ftell_result == -1 was already some lines
before.

The CID report was also not about the ftell return value. Sorry about that.

This reverts commit 6c10db91a5f1a13431b9aefabbbc0918ee5ca4b7.

### DIFF
--- a/src/api/config.c
+++ b/src/api/config.c
@@ -492,10 +492,6 @@ m64p_error ConfigInit(const char *ConfigDirOverride, const char *DataDirOverride
         fclose(fPtr);
         return M64ERR_FILES;
     }
-    if (filelen < 0) {
-        fclose(fPtr);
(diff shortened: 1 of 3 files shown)


conchurnavid:
Also added the patches for CID 38098, 38099 and 38100 from #91

Nebuleon:
I'm implicitly OK with the 3 `switch` commits given what I said in #91 :+1: but am not too sure what the reason was for the file length commit that is being reverted by ea23750. What was it supposed to fix, in what circumstances?

Narann:
This PR seems legit but as it's a revert I prefer tiny words about the original hint.

PR #95, opened 2015-03-05
Fix New Dynarec build failures arising from commit bdb8a88

### LINKED ISSUE #94: new_dynarec arm is broken
Commit https://github.com/mupen64plus/mupen64plus-core/commit/bdb8a88ada81a08a928189bc6600b889350744c3 brakes new_dynarec arm target. 
../../src/r4300/new_dynarec/arm/assem_arm.c:22:28: fatal error: ../cp0_private.h: No such file or directory
compilation terminated.
Makefile:666: recipe for target '_obj/r4300/new_dynarec/new_dynarec.o' failed
make: *** [_obj/r4300/new_dynarec/new_dynarec.o] Error 1

Line 22 should be something like: #include "../../cp0_private.h"
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/r4300/new_dynarec/arm/assem_arm.c#L22

Nebuleon:
Hopefully fixes #94. I looked in the new subdirectories of `new_dynarec` and this should fix all inclusion issues.

### DIFF
--- a/src/r4300/new_dynarec/arm/assem_arm.c
+++ b/src/r4300/new_dynarec/arm/assem_arm.c
@@ -19,7 +19,7 @@
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
 #include "main/main.h"
-#include "../cp0_private.h"
+#include "../../cp0_private.h"
 
 extern int cycle_count;
 extern int last_count;
@@ -102,7 +102,7 @@ const u_int invalidate_addr_reg[16] = {
   0,
   0};
 
-#include "../fpu.h"
+#include "../../fpu.h"
 
 static u_int jump_table_symbols[] = {
   (int)invalidate_addr,
--- a/src/r4300/new_dynarec/x86/assem_x86.c
+++ b/src/r4300/new_dynarec/x86/assem_x86.c
@@ -85,7 +85,7 @@ static const u_short rounding_modes[4] = {
   0xB3F, // ceil
   0x73F};// floor
 
-#include "../fpu.h"
+#include "../../fpu.h"
 
 // We need these for cmovcc instructions on x86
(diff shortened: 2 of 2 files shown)


Narann:
@Gillou68310, can you confirm this fix #94? :)

gizmo98:
Checked. new_dynarec arm is working!

Gillou68310:
Thanks for taking care of this @Nebuleon  ;-) 
But I think that "cp0_private.h" and "main.h" could actually be removed from "assem_arm.c" as they are already included in "new_dynarec.c"

Narann:
Do you think @Gillou68310 is relevant @Nebuleon? If so, you can update this PR before I merge it.

Gillou68310:
@Narann is it possible to ask travis to build both old and new dynarec so we won't end up with those kind a errors?

Nebuleon:
`arm/assem_arm.c` and `new_dynarec.c` are different compilation units; including `cp0_private.h` in the latter would have no effect in the former. The ARM-specific file requires `g_cp0_regs` (for example, [line 1165 of `assem_arm.c` as of commit f648835](https://github.com/mupen64plus/mupen64plus-core/blob/f648835273a2128cd467e95ef69765ac18a3e93a/src/r4300/new_dynarec/arm/assem_arm.c#L1165)), so it should continue to include `cp0_private.h` (at the proper path).

Narann:
Good idea @Gillou68310!

But I have no idea how to do that. I've created a ticket #96.

Gillou68310:
> wow, including .c files in .c files... I really don't like this.

Yeah I agree that's not the best practice! 

Ok I think we can leave these include files for the moment, but the new dynarec will need a big cleanup at some point. 

> Good idea @Gillou68310!
But I have no idea how to do that. I've created a ticket #96.

Thanks @Narran I will check this!

PR #98, opened 2015-03-06
[WIP/RFC] Transfer Pak emulation support

bsmiles32:
Not intended for general user consumption, just an early preview for devs and hackers.

This implements basic emulation of the transfer pak and some game boy cartridges (No MBC, MBC3 no RTC and MBC5).
The transfer pak emulation is very likely inaccurate (based on NRage input plugin) but gets the jobs done on some popular games (Pokemon Stadium, Perfect Dark)
so I went with that. Furthermore, I don't have any information regarding the internal of the transfer pak so unless someone knows better or can record some communication
between a real console and a transfer pak it will be hard to improve on that front...

I tested that mostly with Pokemon Stadium 1 and I was able to browse my pokemons, items, ... in the prof. lab.
However GB Tower doesn't work :( It uses some unknown RSP ucode (Zoinkity made a thread about that on emutalk few years ago) and maybe other stuff we don't emulate quite right.
I also tested with Perfect Dark and could unlock several cheats thanks to that transfer pak so that's good :)
I tested Mario Tennis, but the Transfer Pak menu didn't appeared so no luck here.

With this PR I'm mostly looking for feedback on how to proceed with the following points (but feel free to add anything that you'd like to discuss) :
- how to make the frontend load the GB rom ? (for now I hardcoded that into the core, but maybe it would be better if the frontend was responsible for that)
- how to handle GB save files ? where to store them ?
- how would that additional data impact savestates ?

Please experiment, test and give feedback :)


tony971:
Pinging @mgba-emu and @shonumi because cross-platform gameboy (advance) emulation is a current project for both of them

endrift:
Sorry, I'm not a Game Boy person, I only know GBA. They're very different. Although I can tell you information about the cartridge bus, I'm not sure that would be useful here.

shonumi:
Oh wow, I actually wanted to do something like this myself, but never had the time. Great to see a PR for this at last.

Not that I'm familiar with how M64Plus handles its frontend, but loading the GB ROM via CLI args sounds like the best way forward. Making it another parameter shouldn't be too hard.

GB battery RAMs (save game files) would need to be loaded too. Where exactly I have no clue, unfortunately. At any rate, I'll try building this PR and test it out. Maybe I can help out, but I'm excited to see happen :)

bentley:
MBC3 RTC save format varies between emulators. The closest thing to a standard is the one supported by BGB and VBA, so when it’s implemented this is probably the format to use: http://bgb.bircd.org/rtcsave.html

pepijndevos:
Do you have a link to the emutalk thread? I was unable to find it.

I thought the cart was fairly simple, 16 bit address, 8 bit response, well documented bank switching. You seem to have this covered pretty well. The only other pins I remember are reset and audio.

bsmiles32:
@pepijndevos : The thread I was referring to is [1].
Gameboy bank switching logic is documented well enough on the internet, however, all the logic purely related to the transfer pak is not. The only things I've found are from the NRage source + doc.

I'm sorry I haven't taken the time to put that PR in shape for merge, but I haven't had much time to code lately. Hope to work on that after the next release.

[1] : http://www.emutalk.net/threads/54598-Pok%C3%A9mon-Stadium-%28Pocket-Monster-Stadium%29-requires-RSP-emulation

pepijndevos:
Fascinating! So the problem is not in the controller, but emulating the co-processor? It seems like an interesting thing to dive into, if I knew where to start. I'm thinking that the microcode could be dumped, there might be some docs on the processor, and the image format of Pokemon games is also documented. But yea, it's hard, and I'm an assembly noob.

Papermanzero:
Concerning the UI and the controller configuration.
I would go for an approach like glide64mk2.
Create profiles in an ini file based on game names.

For example a profiles ini could look like that:

```
[PokemonStadium]
Control1 pak = transfer pak, "C:/path/to/the/gb/rom", "C:/path/to/the/gb/save"

[MarioKart64]
Control1 pak = mem pak, "C:/path/to/mem/pak/"
Control2 pak = mem pak, "C:/path/to/mem/pak/"

[Lylatwars]
Control1 pak1 = rumble pak
Control1 pak2 = mem pak, "C:/path/to/mem/pak/folder"
Control1 pak3 = none

[Default]
Control1 pak1 = mem pak
Control1 pak2 = rumble pak
Control2 pak1 = mem pak
Control2 pak2 = rumble pak
```
So in the first case if the user starts Pokemon Stadium, a transfer pak with a certain gb game is connected to the first controller. If the user switches to MarioKart a certain memory pak is in controller 1 and controller 2.  Finally in Lylatwars the user has the ability to switch between pak 1 to 3 with a certain keypress. Last but not least the default case for all other games.

It would be the most convenient way to avoid contentious changes of gb games, paks and settings.

fraser125:
if bsmiles32 or someone else wants some controller protocol data collected.
I've got the following hardware setup and available:
N64 Console
N64 Flash Cart
Controller
Transfer PAK
Several Gameboy / GBC games (let me know if you have a specific title request).
Logic Analyzer
Modified Extension cable (for use with Logic Analyzer)

I've already started recording and documenting much of the controller protocol, many websites get close but make a lot of assumptions that work, but I want to test and verify. I've reviewed some Transfer PAK data so far, mostly just the detection and status data. I'm not sure the best way to really test the Transfer PAK.

I've also got the Voice Recognition Unit if you have thoughts on implementing that hardware let me know. 

Ideally someone could create a Homebrew ROM with the desired N64 code, or tell me the desired N64 cartridge to purchase and exact game repro steps to record.

Narann:
Thanks for your proposition @fraser125!

@bsmiles32 Would you have some time to investigate @fraser125 proposition? Look quite interesting! :)

Cen64 dev is slow for now but I also suggest you jump on Cen64 forum because marathonman would be interested too (I suppose).

tony971:
@fraser125 as far as the VRU goes, there are people who would very much like some progress on that.

http://forum.pj64-emu.com/showthread.php?t=3583

http://www.emutalk.net/threads/55279-Hey-You!-Pikachu-Possible-HLE-Implementation

bsmiles32:
Updated this PR a bit :)
- It has been rebased against master
- It contains some fixes notably the ram enable for some mbc
- It should support the mbc3 rtc with clock_backend

Still WIP, but might interest developers.
If you don't object, this might be merged as is to have stuff in place and refine it later. It shouldn't affect current users as they have to edit the source code to use this (in src/main.c there is an ifdef to toggle and some files are hard-coded).

PR #99, opened 2015-03-09
More decoupling

Nebuleon:
These commits are intended to follow up those in #69 and #83 with more decoupling. Each commit has a message that describes it.

I would like some comments on the two following things I saw while making the changes:

1. In [`r4300_reset_hard` of `r4300.c` as of commit ff1b04f](https://github.com/mupen64plus/mupen64plus-core/blob/ff1b04f43e97ee8d2d4203656a1f08aec4768f6b/src/r4300/r4300.c#L95-L154), I've made it so `rounding_mode` was updated by a new function, `update_x86_rounding_mode`. But I saw that the rest of the function is directly accessing variables of the R4300 implementation despite work done in #83 to prevent this. And, what's more, it even accesses the TLB variables directly. **Should I replace the direct accesses in that function as part of this PR? Should there be TLB pointer functions, like `r4300_tlb_entries()`?**

2. In commit d4282c4511037df835e8740f48f1aeecb7b783dc, I've made it so that writes of `1` to `invalid_code` after 32-bit writes to N64 memory in the Debugger were done by `invalidate_r4300_cached_code` instead. But I saw that the debugger heavily uses the `blocks` and `invalid_code` variables directly, which belong to the Cached Interpreter and Hacktarux JIT. **Should `blocks` and `invalid_code` be torn down in some way?**

I would like to page @bsmiles32 specifically as he has been involved with #83, but also @richard42, @TwinAphex and @Narann, who have been involved with the core.

### DIFF
--- a/src/debugger/dbg_memory.c
+++ b/src/debugger/dbg_memory.c
@@ -61,8 +61,7 @@ static void *opaddr_recompiled[564];
 static disassemble_info dis_info;
 
 #define CHECK_MEM(address) \
-   if (!invalid_code[(address) >> 12] && blocks[(address) >> 12]->block[((address) & 0xFFF) / 4].ops != current_instruction_table.NOTCOMPILED) \
-     invalid_code[(address) >> 12] = 1;
+   invalidate_r4300_cached_code(address, 4);
 
 static void process_opcode_out(void *strm, const char *fmt, ...){
   va_list ap;
--- a/src/main/savestates.c
+++ b/src/main/savestates.c
@@ -192,6 +192,7 @@ static int savestates_load_m64p(char *filepath)
     gzFile f;
     int version;
     int i;
+    uint32_t FCR31;
 
     size_t savestateSize;
     unsigned char *savestateData, *curr;
@@ -410,7 +411,9 @@ static int savestates_load_m64p(char *filepath)
     if ((cp0_regs[CP0_STATUS_REG] & UINT32_C(0x04000000)) == 0)  // 32-bit FPR mode requires data shuffling because 64-bit layout is always stored in savestate file
         shuffle_fpr_data(UINT32_C(0x04000000), 0);
     *r4300_cp1_fcr0()  = GETDATA(curr, uint32_t);
-    *r4300_cp1_fcr31() = GETDATA(curr, uint32_t);
+    FCR31 = GETDATA(curr, uint32_t);
+    *r4300_cp1_fcr31() = FCR31;
+    update_x86_rounding_mode(FCR31);
 
     for (i = 0; i < 32; i++)
     {
@@ -463,6 +466,7 @@ static int savestates_load_pj64(char *filepath, void *handle,
     char buffer[1024];
     unsigned int vi_timer, SaveRDRAMSize;
     int i;
+    uint32_t FCR31;
 
     unsigned char header[8];
     unsigned char RomHeader[0x40];
@@ -550,7 +554,9 @@ static int savestates_load_pj64(char *filepath, void *handle,
     // FPCR
     *r4300_cp1_fcr0() = GETDATA(curr, uint32_t);
     curr += 30 * 4; // FCR1...FCR30 not supported
-    *r4300_cp1_fcr31() = GETDATA(curr, uint32_t);
+    FCR31 = GETDATA(curr, uint32_t);
+    *r4300_cp1_fcr31() = FCR31;
+    update_x86_rounding_mode(FCR31);
 
     // hi / lo
     *r4300_mult_hi() = GETDATA(curr, int64_t);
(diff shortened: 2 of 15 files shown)


Narann:
I'm not an expert of this part of the code but you seems to have moved a lot of stuff and, yeah, this PR would request more comments.

twinaphex:
@Nebuleon Not directly related to this decoupling but more to do with mupen64plus-core in general and one that should deserve some attention at some point:

One of the big things that still separates Mupen64plus libretro from Mupen64plus upstream is that for Mupen64plus libretro, I took the effort of making sure main_run is nonblocking and can be called in a reentrant fashion. I need this not only for the purpose of being able to one-shot per frame in the main loop, but also so that I can do realtime rewind (since the savestate code also had to be made reentrant).

In that regard, things are still not really satisfactory right now in the Libretro fork since I still need libco to be able to skip between threads cooperatively (the main thread is for libretro, the other thread is for Mupen64plus). I'd like that to not be needed so that instead we will have a model for the Mupen64plus emulation core that is much more like a conventional emulator like SNES9x or FCEU, one where there are separate global functions for initing, deiniting, and iterating for one frame run.

Hopefully these things pointed out can help inform your future refactors. Also, I still have a lot of patches that need to be made to new_dynarec so that it is compatible with iOS' ABI - it is probably best we merge that as soon as possible in some manner that you guys find agreeable since other committers have been doing a lot of patches to new_dynarec as of late and it has become a struggle keeping that up-to-date. Also, @Ari64 had some patches that would make new_dynarec PIC-compatible - hopefully that can be looked at at some point too. It would be highly valuable to the libretro port since the entire core is in a dynamic library and therefore it requires code to be PIC-compliant (not to mention this is a good way of future-proofing things anyway as more and more app stores and sandboxed app models are requiring this).

Narann:
Thanks @twinaphex 

> One of the big things that still separates Mupen64plus libretro from Mupen64plus upstream is that for Mupen64plus libretro, I took the effort of making sure main_run is nonblocking and can be called in a reentrant fashion. I need this not only for the purpose of being able to one-shot per frame in the main loop, but also so that I can do realtime rewind (since the savestate code also had to be made reentrant).

Even if this is not started yet, anybody interested by a ticket to discuss about making `main_run` reentrant?


> In that regard, things are still not really satisfactory right now in the Libretro fork since I still need libco to be able to skip between threads cooperatively (the main thread is for libretro, the other thread is for Mupen64plus). I'd like that to not be needed so that instead we will have a model for the Mupen64plus emulation core that is much more like a conventional emulator like SNES9x or FCEU, one where there are separate global functions for initing, deiniting, and iterating for one frame run.

What do you suggest? Improve core API to with the functions you've pointed out? Would it be enough?

Others, do you think it could be interesting to have a ticket for this?

I guess all of this is not a priority right now (dynarec seems to be what requiere much love now) that's why I ask if it's relevant to create tickets. :)

Gillou68310:
> Also, @Ari64 had some patches that would make new_dynarec PIC-compatible - hopefully that can be looked at at some point too.

Which patches are you talking about? Maybe I can take a look at it.

littleguy77:
> I guess all of this is not a priority right now (dynarec seems to be what requiere much love now) that's why I ask if it's relevant to create tickets. :)

As a general rule, yes, always open a ticket even if the issue is obscure.  That way @twinaphex doesn't have to repeat the same requests over and over and devs are aware.

Nebuleon:
@twinaphex: What you describe is neither reentrant nor non-blocking, unless one frame of emulation can be requested in the middle of running another, e.g. in two separate threads ([reentrancy](http://en.wikipedia.org/wiki/Reentrancy_%28computing%29)), or one frame of emulation can be requested and it immediately returns (i.e. doesn't block) before it's actually done ([non-blocking](http://en.wikipedia.org/wiki/Non-blocking_I/O)).

Please create separate issues as appropriate to describe in detail the main differences between libretro mupen64plus and this repository and what could be done to minimise the differences, and nother one for the PIC patches, for wider discussion.

Nebuleon:
This pull request has been updated.

* Start from `master`, as of commit 084c54d0fa7d162802b4be8c685bb1e2d8215d53;
* Cherry-pick (5):
  * fpu.h: Use the MIPS FPU control register instead of 'rounding_mode';
  * interpreter_r4300.def: Convert some more types to uint32_t and uint64_t;
  * dbg_memory.c: Do not use invalid_code in CHECK_MEM;
  * Move *_mode, x86 rounding mode constants, out of cp1.c;
  * Factor out the code that updates the x86 rounding mode;
* Remake with comments from @bsmiles32 (1):
  * recomp.c: Remove duplication in code for alternate physical addresses;
* Delete (2):
  * interpreter_tlb.def: Do not use invalid_code in TLBWrite;
  * Delete unndeeded adler32 field from precomp_block; clean up headers.

Narann:
Thanks Nebuleon! Good code, goode comments, good commit message. All of this look great!

This PR start to be huge, I will merge it tomorrow so please core people: Test and report to avoid big problems.

bsmiles32:
Quickly tested. No regression on my side.

Very nice work !

Nebuleon:
Updated again with comments from Narann.

PR #103, opened 2015-03-13
Added more fields and fixed the Players field

BonzaiThePenguin:
Added fields for Developer, Publisher, Date, Genre, and ESRB, and fixed the value for the Players field in a few cases. This data was populated from the following tables on Wikipedia, so we may need to add attribution?

http://en.wikipedia.org/wiki/Chronology_of_Nintendo_64_games
http://en.wikipedia.org/wiki/List_of_Nintendo_64_games

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -12,6 +12,11 @@ SaveType=Controller Pack
 Status=1
 Rumble=Yes
 Players=4
+Developer=Eurocom
+Publisher=Electronic Arts
+Date=2000/12/08
+Genre=Action/First-Person Shooter
+ESRB=T
 
 [9D58996A8AA91263B5CD45C385F45FE4]
 GoodName=007 - The World is Not Enough (U) [!]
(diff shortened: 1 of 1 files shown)


BonzaiThePenguin:
Warning: just noticed that this currently conflicts with rom.c's sanity checks. It's hard-coded to only accept GoodName, CRC, RefMD5, SaveType, Status, Players, Rumble, CountPerOp, and Cheat.

Narann:
Great work! Thanks!

If you don't mind, I would let others discuss about those additions.

For example. IIRC, _developer_ is an information already write into the ROM's header. Same for sanity checks you just pointed!

BonzaiThePenguin:
I don't think the developer is included in the header. I checked beforehand to make sure, and "Rare" did not appear in any of the Rare games I have. It only has the name of the game.

(On a related note, the "release date" field from the N64 header documentation that's floating around online is probably mislabeled. I wrote down the values for the games in my library and three of the four bytes were identical across all of them, and the fourth one didn't seem to have any relation to a release date or a release order.)

rom.c would definitely have to be updated before any changes should be accepted, as right now it logs a ton of warnings to logcat about unrecognized fields.

PR #106, opened 2015-03-15
Fix regression introduced by commit 8ae91b5.

bsmiles32:


src/r4300/new_dynarec/new_dynarec.c
```diff
@@ -1149,9 +1149,12 @@ static void ll_kill_pointers(struct ll_entry *head,int addr,int shift)
        (((ptr-(u_int)base_addr-MAX_OUTPUT_BLOCK_SIZE)>>shift)==((addr-(u_int)base_addr)>>shift)))
     {
       inv_debug("EXP: Kill pointer at %x (%x)\n",(int)head->addr,head->vaddr);
+      u_int host_addr=(int)kill_pointer(head->addr);
       #if NEW_DYNAREC == NEW_DYNAREC_ARM
-        u_int host_addr=(int)kill_pointer(head->addr);
```
Narann: Why this wasn't working?

bsmiles32:
As explained by @Gillou68310 , kill_pointer call has to be done for all platforms not just ARM. The returned value however is only needed  by the ARM. Since it wasn't used by the x86 it generated warnings on x86 but not on ARM.

Narann:
I just realize you did a [PR and merge instantly](https://github.com/mupen64plus/mupen64plus-core/pull/105). I suggest to let one tiny day for PRs related to dynarec code to maybe let some others test it as we have no way yet to test multiple build options.

bsmiles32:
see the name of the PR :)

Narann:
What I mean is that if we had wait a tiny day before merge the previous PR we maybe wouldn't have to create another PR to fix it. :D

PR #110, opened 2015-03-21
Cleanup includes order and add missing includes

conchurnavid:


richard42:
this fails the travis build with DEBUGGER=1

conchurnavid:
Thanks. I have also added the missing include in this file.

src/ai/ai_controller.c
```diff
@@ -21,14 +21,18 @@
 
 #include "ai_controller.h"
 
+#include <string.h>
+
 #include "main/rom.h"
 #include "memory/memory.h"
+#include "r4300/cp0.h"
+#include "r4300/interupt.h"
+#include "r4300/mi_controller.h"
```
bsmiles32: These includes were omitted on purpose because they are provided by the r4300_core which acts as the "interface" of the r4300 subsystem for other subsystems (like here the AI). I used that logic during all my refactorings of the different subsystems, that's why there are so many "missing includes".

I can understand that "good style" would be to always include headers for functions that are used in a module, but here I consider these headers to be an implementation detail and therefore not required for inclusion. 

Should I annotate all occurrences of headers that I feel should not be re-added ? Basically {ai,pi,ri,si,vi}_controller.h, {r4300,rdp,rsp}_core.h act as the interface header for their respective subsystems and therefore any other header should not be included in external subsystems.

conchurnavid: ok, sounds fair. I can modify it later (most likely not in the next few days due to some other tasks). Just don't merge it until I've updated the source.

conchurnavid:
@bsmiles32 Something odd I've noticed:

 * force_detected_rdram_size_hack is not exported by ri_controller.h but still used by pi_controller.c
 * rompause is not exported by r4300_core but still used by main/main.c (same for COUNT_PER_OP, ...)
 * many, many more things were the internal headers of a subsystems are used outside the subsystem

I have now reverted all changes outside the subsystem when it included things from the subsystems you've mentioned.

bsmiles32:
@conchurnavid thanks for the hard work and the feedback.

> force_detected_rdram_size_hack is not exported by ri_controller.h but still used by pi_controller.c

force_detected_rdram_size_hack is a hack and I made explicit its usage. pi_controller includes therefore the rdram_detection_hack header. But I wouldn't mind if it was included through ri_controller.h.

> rompause is not exported by r4300_core but still used by main/main.c (same for COUNT_PER_OP, ...)
Those are part of the r4300 module and I didn't change that, so when they are used, r4300.h should be included. The problem here is that the r4300 subsystem refactoring is still not finished, so not everything is "properly" abstracted through an interface. This work is still ongoing (@Nebuleon did some more work here) but we are still limited by the new_dynarec.

bsmiles32:
@richard42 : I don't think it was ready for merge yet given the inline comments. The last message was more a general comment on the overall work. I should have made explicit that inline comments should have been taken into account before merge... But that's not really important because we can fix that with an additional commit.

richard42:
sorry about that; I thought it was all cleared up

PR #117, opened 2015-05-05
Squelch sscanf warning in arm_cpu_features.

littleguy77:
GCC 4.8 warns about a format/type mismatch in the sscanf statements.

@Gillou68310 Would you mind reviewing this, since you were the original author?  Is there a reason you were starting with int type and then casting to narrower types?


Gillou68310:
This can be merged, I don't even see a reason why I did this in the first place!!!

PR #123, opened 2015-06-16
Merge back "Avoid .text relocations in PIC shared library" fix + cleanup linkage_arm.S

Gillou68310:
The "Avoid .text relocations in PIC shared library" patch causes crash at start of Conker's BFD on android. But after cleaning up linkage_arm.S this crash disapears. I'm not sure what fixed the issue but IMO it's either a GCC issue or a code alignement issue.


Narann:
So here we go? :)

Gillou68310:
Yep ;-)

PR #125, opened 2015-06-30
Remove Zelda OoT Subscreen Delay Fix from mupen64plus.ini

Gillou68310:
The Subscreen Delay Fix for zelda OoT is now in conflict with GLideN64 own subscreen fix.
gonetz/GLideN64#589

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -6852,8 +6852,6 @@ SaveType=Flash RAM
 Status=4
 Rumble=Yes
 Players=1
-; Subscreen Delay Fix
-Cheat0=801DB78B 0002
 ; End Credits Fix
 Cheat1=D109A814 0320,8109A814 0000,D109A816 F809,8109A816 0000
 
@@ -6864,8 +6862,6 @@ SaveType=Flash RAM
 Status=4
 Rumble=Yes
 Players=1
-; Subscreen Delay Fix
-Cheat0=801DB78B 0002
 ; End Credits Fix
 Cheat1=D109A834 0320,8109A834 0000,D109A836 F809,8109A836 0000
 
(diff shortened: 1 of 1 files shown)


AmbientMalice:
This will cause other video plugins to suffer the delay. Project 64 solved this by making the hack conditional based on a whitelist of plugins - namely Jabo and Glide64.

Gillou68310:
> This will cause other video plugins to suffer the delay.

That's correct, however I originaly added this hack to the core because I was convinced that the problem was caused by a core issue.
Actually it has been proved to be a frame buffer to RDRAM issue. 
So IMO this hack has nothing to do in the core. Users still have 2 possible way to fix the issue, either use GLideN64 or manually activate Subscreen Delay Fix cheat code if using another video plugin. 

Also gonetz's hack might not be to difficult to integrate on other video plugins. I'll see what I can do.

Gillou68310:
Here's the subscreen delay fix for gliden64mk2:
https://github.com/Gillou68310/mupen64plus-video-glide64mk2/tree/oot_subscreen_delay

Laqueeshous:
does such a fix exist for majora's mask?

erjoalgo:
I had to manually add this cheat code to the end of `/usr/share/games/mupen64plus/mupen64plus.ini`

```
; f62dcec4-1169-4ba2-b47f-6048170a50ad-apply-all-possible-subscren-delay-fixes
[57A9719AD547C516342E1A15D5C28C3D]
GoodName=Legend of Zelda, The - Ocarina of Time (U) (V1.2) [!]
CRC=693BA2AE B7F14E9F
Players=1
SaveType=SRAM
Rumble=Yes
; Subscreen Delay Fixes
Cheat0=801D860B 0002
Cheat1=801D864B 0002
Cheat2=801D8F4B 0002
Cheat3=801D8F8B 0002
Cheat4=801DA5CB 0002
Cheat5=801DA78B 0002
Cheat6=801DAE8B 0002
Cheat7=801DB74B 0002
Cheat8=801DB78B 0002
; f62dcec4-1169-4ba2-b47f-6048170a50ad-apply-all-possible-subscren-delay-fixes
```

Since this was removed from core, it would be useful if this cheat could be at least added by default mupen64plus.ini file to fix this annoying subscreen delay for users.

PR #128, opened 2015-08-13
Rename new_dynarec.txt to new_dynarec.mediawiki

Narann:
This will simply allow this file to be read by github


Narann:
Here is the result: https://github.com/mupen64plus/mupen64plus-core/blob/rename-dynarec-wiki/doc/new_dynarec.mediawiki

PR #131, opened 2015-08-31
Allow building the old dynarec on x64 with VisualStudio2013

Gillou68310:


richard42:
Thanks for your work on this.  I have a few comments:
1. Are you sure that the "New Dynarec" works with the x64 build?  I'm pretty sure that this is only usable in 32-bit land on Linux.
2. Instead of using macros to differentiate between "long" and "long long" in gencallinterp(), can you use uintptr_t from stdint.h?  I'm surprised that this is the only place where we had a 64-bit "long" compatibility issue.
3. Instead of using a separate external ASM file for dyna_start (only for 64-bit windows), can you use inline ASM code?  I don't like separating the win/linux implementations so much, and with inline code you could keep the saved register values and return_address as static.

Gillou68310:
That's correct the current version of the "New Dynarec" won't work on x64. The problem is that I can't remove both New_Dynarec_Debug and New_Dynarec_Release configurations from the x64 config or they will be removed from the Win32 config too. In the meantime I already managed to get a partially working version of the x64 new_dynarec on windows.

Gillou68310:
UPDATE
https://github.com/Gillou68310/mupen64plus-core/commit/6d0a99f54d9841206c29a7f248cc6a940cbc7440

PR #132, opened 2015-10-05
new_dynarec: Avoid replacing an allocated temporary register

Gillou68310:
Fixed crash in castlevania

### DIFF
--- a/src/r4300/new_dynarec/new_dynarec.c
+++ b/src/r4300/new_dynarec/new_dynarec.c
@@ -10127,6 +10127,10 @@ int new_recompile_block(int addr)
   // to use, which can avoid a load-use penalty on certain CPUs.
   for(i=0;i<slen-1;i++)
   {
+    // Avoid replacing an allocated temporary register
+    if(count_free_regs(regs[i].regmap)<=minimum_free_regs[i])
+      continue;
+
     if(!i||(itype[i-1]!=UJUMP&&itype[i-1]!=CJUMP&&itype[i-1]!=SJUMP&&itype[i-1]!=RJUMP&&itype[i-1]!=FJUMP))
     {
       if(!bt[i+1])


Narann:
Anyone to test this? :)

src/r4300/new_dynarec/new_dynarec.c
```diff
@@ -10127,91 +10127,150 @@ int new_recompile_block(int addr)
   // to use, which can avoid a load-use penalty on certain CPUs.
   for(i=0;i<slen-1;i++)
   {
-    if(!i||(itype[i-1]!=UJUMP&&itype[i-1]!=CJUMP&&itype[i-1]!=SJUMP&&itype[i-1]!=RJUMP&&itype[i-1]!=FJUMP))
+    if(count_free_regs(regs[i].regmap)>minimum_free_regs[i])
```
Narann: What about:
```c
// Avoid replacing an allocated temporary register
if(count_free_regs(regs[i].regmap)<=minimum_free_regs[i])
    continue
```
And save a huge if scope?

Gillou68310: That's more aesthetic for sure ;-)
Both code are equivalent anyway

Narann: I'm not a big fan of nested if statement in general. It make the code harder to read/maintain. I like the "filter approach". The idea is to allow "if not,continue" at the top of a for loop (and only here). This technical "filter" the iterated content before the code really begin. So if you have a if statement taking the whole function block without any else statement it's a good candidate for "filtering".

Roughly:
```c
for loop

 # filter code block (find what could "skip" the loop)

 # loop execution code block
```
Do the separation at the beginning of the for loop make the loop execution code easier to read.

Gillou68310: You're right the new_dynarec already is a bit hard to read, no need to make it worse ;-)
I updated the pull request.

src/r4300/new_dynarec/new_dynarec.c
```diff
@@ -10127,6 +10127,9 @@ int new_recompile_block(int addr)
   // to use, which can avoid a load-use penalty on certain CPUs.
   for(i=0;i<slen-1;i++)
   {
+    if(count_free_regs(regs[i].regmap)<=minimum_free_regs[i])
```
Narann: I also suggest a comment: `// Avoid replacing an allocated temporary register`

Except if you think the code is explicit enough (I'm not an expert of dynarec).

Gillou68310:
Updated with comment.
Should I create a new pull request for this? I don't want to pollute the history with 3 commits for just one line of code.

Narann:
Is rebase an option for you? :)

Gillou68310:
Sure but rebasing won't help me to squash the 3 commits into one?

Narann:
I [guess](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/) rebase is supposed to do this (I'm still learning rebase).

Gillou68310:
Yep you're right rebase has a squash option, I never used it until now. Testing it right now ;-)

Gillou68310:
Done :-p

PR #134, opened 2015-10-12
Use one struct for R4300 registers on all architectures

Nebuleon:
The New Dynarec/ARM redefines global variables in its own memory block in order to bring a lot of them together in memory for easy indexing. This commit does the same for all other R4300 cores on all architectures by moving a lot of often-used global variables from many files into one struct in `r4300.c`:

> `reg[32]`, `hi`, `lo`, `g_cp0_regs[32]`, `FCR0`, `FCR31`, `llbit`, `reg_cop1_simple[32]`, `reg_cop1_double[32]`, `reg_cop1_fgr_64[32]`, `delay_slot`, `next_interupt`, `address`, `rdword`, `cpu_byte`, `cpu_hword`, `cpu_word`, `cpu_dword` => `g_state`

The compiler is then allowed to reference every member of that structure with offsets from a single address, which is faster on ARM and MIPS (and may or may not be faster on x86 due to processor microarchitecture changes) when emitting position-independent code.

In addition, I have aligned the new structure to 4 KiB in order to induce only 1 TLB miss for accessing it.

The new structure is hidden and recreated in the New Dynarec/ARM as usual, but the number of recreated variables is greatly reduced.

I expect the performance to increase in these cores:

* Pure Interpreter: 2%..10% faster, due to fewer reloads of global variable addresses, increased cache locality and reduced TLB misses; this depends on a game's use of multiplier unit instructions, `MFC0`, `MTC0`, `[D]MFC1`, `[D]MTC1`, floating-point width conversion instructions and memory access instructions, which accessed more than one global variable (and now access the members of the structure).
* Cached Interpreter: <2% faster, due to increased cache locality and reduced TLB misses. It loads operand pointers from a `struct precomp_instr`, so after compilation, there are no reloads of global variable addresses.
* Hacktarux JIT/x86, x86-64: <2% faster, due to increased cache locality and reduced TLB misses. Better code could be generated to take advantage of the new structure, but I didn't change anything there.

A possibly undesirable aspect of this change is that the Coprocessor 0-related variables are no longer in `cp0.c`, the FPU is no longer in `cp1.c`, and the global variables used as parameters to the memory accessor functions are no longer in `memory.c`.

Commit 3c9a3fcea55a4f43f60c761e49d62320288ef3fe will definitely not build on ARM, and bbe2a26b623d582a4d83bea55a8b50f8c8a17d7c should build, but I don't know if it works properly because I don't have an ARM device to test with.

### DIFF
--- a/src/memory/memory.c
+++ b/src/memory/memory.c
@@ -48,22 +48,6 @@
 
 extern int fast_memory;
 
-#if NEW_DYNAREC != NEW_DYNAREC_ARM
-// address : address of the read/write operation being done
-uint32_t address = 0;
-#endif
-
-// values that are being written are stored in these variables
-#if NEW_DYNAREC != NEW_DYNAREC_ARM
-uint32_t cpu_word;
-uint8_t cpu_byte;
-uint16_t cpu_hword;
-uint64_t cpu_dword;
-#endif
-
-// addresse where the read value will be stored
-uint64_t* rdword;
-
 // hash tables of read functions
 void (*readmem[0x10000])(void);
 void (*readmemb[0x10000])(void);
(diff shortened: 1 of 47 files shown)


Narann:
Interesting, I would wait for others to comment before merge such PR. Thanks for the hard work!

Nebuleon:
Of course; I even had a build failure, so it's possible that there are more errors lurking. Travis didn't get notified about commit 7bd0078caaaa0508e3ecca4fa828ef5e88fe5186 either, apparently.

PR #135, opened 2015-10-14
new_dynarec: Fixed bomberman 64

Gillou68310:
The delay slot cannot overwrite R0, it's not necessary to make a copy of the branch address register (R0)

### DIFF
--- a/src/r4300/new_dynarec/new_dynarec.c
+++ b/src/r4300/new_dynarec/new_dynarec.c
@@ -4970,7 +4970,7 @@ static void do_ccstub(int n)
     if(itype[i]==RJUMP)
     {
       int r=get_reg(branch_regs[i].regmap,rs1[i]);
-      if(rs1[i]==rt1[i+1]||rs1[i]==rt2[i+1]) {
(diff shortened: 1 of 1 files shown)


src/r4300/new_dynarec/new_dynarec.c
```diff
@@ -4970,7 +4970,7 @@ static void do_ccstub(int n)
     if(itype[i]==RJUMP)
     {
       int r=get_reg(branch_regs[i].regmap,rs1[i]);
-      if(rs1[i]==rt1[i+1]||rs1[i]==rt2[i+1]) {
+      if((rs1[i]==rt1[i+1]||rs1[i]==rt2[i+1])&&(rs1[i]!=0)) {
```
Narann: Wouldn't check `rs1[i]` first would be faster? (I just ask)
```C
(rs1[i]!=0)&&(rs1[i]==rt1[i+1]||rs1[i]==rt2[i+1])
```

Gillou68310: Lol this won't make a big difference. Also the case where rs1==0 is unlikely to happen very often.

Gillou68310: Btw does someone tested my fix for castlevania?
:-p

PR #137, opened 2015-10-23
Dynamic linker

Gillou68310:
dyna_linker and dyna_linker_ds are really hard to understand looking at the asm code, so I decided to port both functions to C for better understanding.
Obviously the C functions will run slower than the asm ones, but the difference is unnoticeable even on android devices.
This is beeing tested on paulscode.com for a month without regressions so far.

### DIFF
--- a/src/r4300/new_dynarec/arm/assem_arm.c
+++ b/src/r4300/new_dynarec/arm/assem_arm.c
@@ -21,6 +21,9 @@
 #include "../../cp0_private.h"
 #include "main/main.h"
 
+void *dynamic_linker(void * src, u_int vaddr);
+void *dynamic_linker_ds(void * src, u_int vaddr);
+
 extern int cycle_count;
 extern int last_count;
 extern int pcaddr;
@@ -107,8 +110,8 @@ const u_int invalidate_addr_reg[16] = {
 static u_int jump_table_symbols[] = {
   (int)invalidate_addr,
   (int)jump_vaddr,
-  (int)dyna_linker,
-  (int)dyna_linker_ds,
+  (int)dynamic_linker,
+  (int)dynamic_linker_ds,
   (int)verify_code,
   (int)verify_code_vm,
   (int)verify_code_ds,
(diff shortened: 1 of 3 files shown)


Narann:
Thanks! :)

The dynarec code is quite hard to review. I mean, I can merge it but I have no idea if all of this is good or not.

Now, I will try to "call" other dynarec members before merge, even a simple "look good to me" is fine.

So, @Nebuleon @bsmiles32, any thoughts?

Nebuleon:
If it's anything like what I think *(i.e. calls to `dynamic_linker` are placed in generated code to be rewritten, later, as direct jumps to native code implementing N64 code at a certain address that hasn't been emitted yet)*, making the `dynamic_linker` C is not going to do much to performance, because it was only <2% of the runtime anyway. But it's at least more understandable than the corresponding assembly.

I can't speak for the correctness of the `ht_bin` and linked list accesses, because I don't work on the New Dynarec. I presume you've tested those.

For commit c060a00a874ffb5a598222c21f513c00934ee914:
* The jump rewrites and corresponding (ARM-only) cache flushing are correct;
* The TLB Refill exception creation code looks like it does what the assembly code did, so it's OK. It doesn't jump to `0x80000180` if an exception is already being taken -- but I guess the handler is in physical memory, so you don't expect nested TLB Refill exceptions (see `src/r4300/exception.c` for more info);
* The assertions to make sure that the calling code is still set up to call `dynamic_linker[_ds]` before rewriting the call into a jump are a nice touch.

On x86, I'm not sure if `PUSH PUSH CALL` is going to be met with `POP POP` inside `dynamic_linker[_ds]`, because there are a multitude of calling conventions. For example, the compiler might decide that the arguments, `void* src, int vaddr`, were passed in `EAX EDX`, not the stack. You could add stdcall attributes to the C functions to ensure that the compiler takes the arguments from the stack and ends the function with clean-up code:

```C
#if defined(__GNUC__)  /* Compiling with gcc */
#  define STDCALL __attribute__((stdcall))
#elif defined(_MSC_VER)  /* Compiling with Microsoft C */
#  define STDCALL __stdcall
#else
#  error "Unknown x86 compiler - please define STDCALL in assem_x86.c to something that grabs function arguments from the stack"
#endif
```

For commit b85e203107026a866be4c8364dbd2fe92524ad9b:

Now you always jump through trampolines, or via full addresses, to get to `dynamic_linker[_ds]`?

For commit b85e203107026a866be4c8364dbd2fe92524ad9b:
* Oh boy, that copy-pasting of code 4 times. I bet this code was already elsewhere and working, though. Maybe you could have a function, present once per architecture, that would do the revalidation/restoration of dirty code, and use that everywhere, submitting that as a separate cleanup PR.

Gillou68310:
@Nebuleon thanks for reviewing my code ;-)

> If it's anything like what I think (i.e. calls to dynamic_linker are placed in generated code to be rewritten, later, as direct jumps to native code implementing N64 code at a certain address that hasn't been emitted yet), making the dynamic_linker C is not going to do much to performance, because it was only <2% of the runtime anyway. But it's at least more understandable than the corresponding assembly.

Yep this is exactly what you think!

> I can't speak for the correctness of the ht_bin and linked list accesses, because I don't work on the New Dynarec. I presume you've tested those.

Yes I did ;-)

> On x86, I'm not sure if PUSH PUSH CALL is going to be met with POP POP inside dynamic_linker[_ds], because there are a multitude of calling conventions. For example, the compiler might decide that the arguments, void* src, int vaddr, were passed in EAX EDX, not the stack. You could add stdcall attributes to the C functions to ensure that the compiler takes the arguments from the stack and ends the function with clean-up code

The new dynarec already do the assumption that stdcall is used, see div64/divu64 + fpu instructions ( eg. add_s ). But I agree with you we should probably make sure stdcall is actually being used.

> For commit b85e203:
Now you always jump through trampolines, or via full addresses, to get to dynamic_linker[_ds]?

Trampolines jump are only used if the jump from recompiled code to dynamic_linker is beyond the 32M limit (Usually this is happening when the cache gets full), otherwise jump via full addresses are used.

> For commit f7a6796:
Oh boy, that copy-pasting of code 4 times. I bet this code was already elsewhere and working, though. Maybe you could have a function, present once per architecture, that would do the revalidation/restoration of dirty code, and use that everywhere, submitting that as a separate cleanup PR.

You're right I picked up that code from ```get_addr```. I agree that the copy-paste is a bit ugly, I'll see what I can do to cleanup this code maybe later.

Narann:
`maybe later.` meaning before or after I merged? :)

Gillou68310:
After ;-)

PR #138, opened 2015-11-02
new_dynarec: Reverted printing method when warning about a block comp…

Gillou68310:
…iled at bogus memory address

This was causing slowdown in CBFD. The assem_debug macro is defined as a null function by default.


Narann:
Thanks!

gizmo98:
Thanks!

Narann:
You're welcome!

Gillou68310:
;-)

PR #140, opened 2015-11-10
Call gfx moveScreen when SDL window moved (Windows only)

Gillou68310:
This is only necessary for angrylion's plugin

### DIFF
--- a/src/main/eventloop.c
+++ b/src/main/eventloop.c
@@ -21,6 +21,7 @@
  * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * */
 
 #include <SDL.h>
+#include <SDL_syswm.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


Gillou68310:
With SDL2 we can use SDL_WINDOWEVENT, but with SDL1.2 we're stuck with this implementation

conchurnavid:
Can this please be reverted. It doesn't even build on windows (mingw64):

     ../../src/main/eventloop.c: In function 'event_sdl_filter':
     ../../src/main/eventloop.c:345:38: error: invalid operands to binary == (have 'union <anonymous>' and 'int')
                  if(event->syswm.msg->msg == WM_MOVE)

Gillou68310:
I don't really see why this is happening, anyway we should better fix this issue instead of just reverting.

conchurnavid:
You can't see why this happens? WM_MOVE is an integer https://msdn.microsoft.com/en-us/library/windows/desktop/ms632631(v=vs.85).aspx and event->syswm.msg->msg makes no sense in C+SDL2 because it is an anonymous union https://wiki.libsdl.org/SDL_SysWMmsg https://hg.libsdl.org/SDL/file/b2c000b256ea/include/SDL_syswm.h#l138

Maybe you meant event.syswm.msg->msg.win.msg

Gillou68310:
> With SDL2 we can use SDL_WINDOWEVENT, but with SDL1.2 we're stuck with this implementation

I assumed you were using SDL1.2 like me. Ok I have to add a check for SDL1.2.

Issue #141, opened 2015-11-12
Fails to build on windows (mingw64)


conchurnavid:
Pull request #140 caused this when building with mingw64 for windows:

     ../../src/main/eventloop.c: In function 'event_sdl_filter':
     ../../src/main/eventloop.c:345:38: error: invalid operands to binary == (have 'union <anonymous>' and 'int')
                  if(event->syswm.msg->msg == WM_MOVE)

conchurnavid:
WM_MOVE is an integer https://msdn.microsoft.com/en-us/library/windows/desktop/ms632631(v=vs.85).aspx and event->syswm.msg->msg makes no sense in C+SDL2 because it is an anonymous union https://wiki.libsdl.org/SDL_SysWMmsg https://hg.libsdl.org/SDL/file/b2c000b256ea/include/SDL_syswm.h#l138

Maybe event.syswm.msg->msg.win.msg was meant

PR #143, opened 2015-11-12
Fixed gfx moveScreen when SDL_VERSION > 1.2

Gillou68310:
@conchurnavid this should fix https://github.com/mupen64plus/mupen64plus-core/issues/141,
could you test please?


conchurnavid:
I can confirm that it builds again with this change

PR #150, opened 2016-02-04
Improve speed limiter accuracy

fzurita:
If the speed limiter sleeps for too long, we will sleep for less time the next time around.

This is useful if the sleep period is so short that the application cannot sleep for that short a time which unnecessarily slows down the frame rate.

### DIFF
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -736,13 +736,31 @@ static void apply_speed_limiter(void)
 {
     unsigned int CurrentFPSTime;
     static unsigned int LastFPSTime = 0;
+    static unsigned int StartFPSTime = 0;
     static float VITotalDelta;
     static float VIDeltas[64];
     static unsigned int VIDeltasIndex;
 
-    double VILimitMilliseconds = 1000.0 / ROM_PARAMS.vilimit;
+    //Reset if the sleep needed is an unreasonable value
+    static const double minSleepNeeded = -50;
+    static const double maxSleepNeeded = 50;
+    static unsigned long totalVIs = 0;
+    static int resetOnce = 0;
+    static int lastSpeedFactor = 100;
+
+    //if this is the first time or we are resuming from pause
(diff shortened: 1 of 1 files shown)


fzurita:
@richard42 I'm not sure if you are receiving notifications for pull requests. Could you please take a look at this? It's a fairly simple change, and I think it's correct. Thanks.

richard42:
Yeah, it's in my inbox.  Sorry, I'm just quite a bit behind in this work at the moment.  I will get to it; I think there was another speed limiter change as well.

fzurita:
I did some more testing and this is not entirely accurate either. I think I see a better way of making this almost entirely correct. I'll update my pull request if it works.

fzurita:
Ok, fix has been comitted

richard42:
Ok I'll merge it, the code looks solid.  But unfortunately I think this area still needs more work.  There is a fundamental conflict between the speed limiters in the audio plugin and the core vi callback.  We need to implement a method for getting feedback from the audio plugin, and choose delay times here which help keep the audio buffer at its target fullness level.  When sound is not being played, or the dummy audio plugin is connected, we can use any speed control mechanism (such as the current one) in the core and it should be good.  It would also be nice to take into account the frame sync delay time (if the opengl swapbuffer is waiting to sync at 60hz intervals) from the video plugin, but that could get very complicated...

fzurita:
This is the audio plugin I'm working on right now:

https://github.com/mupen64plus-ae/mupen64plus-ae/pull/546

My goal is to keep the audio buffers at a certain fullness while the core is running on a separate thread. The way I'm achieving this currently is by using the SoundTouch library and speeding up or slowing down audio if buffers get too big or too small. The SoundTouch library allows audio to speed up it slow down without any pitch changes, so it's not very noticeable unless you have a well trained ear to the specific game.

My initial approach which didn't work too well was by slowing down or speeding up the core to maintain specific buffer size. This led to noticeable changes in the frame rate while playing which was very distracting.

PR #175, opened 2016-08-09
word-wise dma copy

ricrpi:
If DMA transfers are word aligned then a more efficient copy can be done. I have not seen a dma transfer that is not word aligned however I have not extensively checked all roms so have left the safer byte-wise implementation in.

### DIFF
--- a/src/pi/pi_controller.c
+++ b/src/pi/pi_controller.c
@@ -126,10 +126,17 @@ static void dma_pi_write(struct pi_controller* pi)
     dram = (uint8_t*)pi->ri->rdram.dram;
     rom = pi->cart_rom.rom;
 
-    for(i = 0; i < longueur; ++i)
-    {
-        dram[(dram_address+i)^S8] = rom[(rom_address+i)^S8];
-    }
+	if (((dram_address | rom_address | longueur) & 3) != 0)
+	{	
+    	for(i = 0; i < longueur; ++i)
+    	{
+        	dram[(dram_address+i)^S8] = rom[(rom_address+i)^S8];
+    	}
+	}
+	else
+	{
+		MEMCPY4(&dram[dram_address], &rom[rom_address], longueur);
+	}
 
     invalidate_r4300_cached_code(0x80000000 + dram_address, longueur);
     invalidate_r4300_cached_code(0xa0000000 + dram_address, longueur);
--- a/src/rsp/rsp_core.c
+++ b/src/rsp/rsp_core.c
@@ -47,14 +47,25 @@ static void dma_sp_write(struct rsp_core* sp)
     unsigned char *spmem = (unsigned char*)sp->mem + (sp->regs[SP_MEM_ADDR_REG] & 0x1000);
     unsigned char *dram = (unsigned char*)sp->ri->rdram.dram;
 
-    for(j=0; j<count; j++) {
-        for(i=0; i<length; i++) {
-            spmem[memaddr^S8] = dram[dramaddr^S8];
-            memaddr++;
-            dramaddr++;
-        }
-        dramaddr+=skip;
-    }
+	if (((memaddr | dramaddr | length | skip) & 3) != 0)
+	{
+    	for(j=0; j<count; j++) {
+        	for(i=0; i<length; i++) {
+        	    spmem[memaddr^S8] = dram[dramaddr^S8];
+        	    memaddr++;
+        	    dramaddr++;
+        	}
+        	dramaddr+=skip;
+    	}
+	}
+	else
+	{
+		for(j=0; j<count; j++) {
+			MEMCPY4(&spmem[memaddr], &dram[dramaddr], length);
+			memaddr += length;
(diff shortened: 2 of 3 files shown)


src/pi/pi_controller.c
```diff
@@ -126,10 +126,17 @@ static void dma_pi_write(struct pi_controller* pi)
     dram = (uint8_t*)pi->ri->rdram.dram;
     rom = pi->cart_rom.rom;
 
-    for(i = 0; i < longueur; ++i)
-    {
-        dram[(dram_address+i)^S8] = rom[(rom_address+i)^S8];
-    }
+	if (((dram_address | rom_address | longueur) & 3) != 0)
+	{	
+    	for(i = 0; i < longueur; ++i)
```
Narann: Can I suggest to add a tab in from of the for loop to keep code readable?

src/pi/pi_controller.c
```diff
@@ -126,10 +126,17 @@ static void dma_pi_write(struct pi_controller* pi)
     dram = (uint8_t*)pi->ri->rdram.dram;
     rom = pi->cart_rom.rom;
 
-    for(i = 0; i < longueur; ++i)
-    {
-        dram[(dram_address+i)^S8] = rom[(rom_address+i)^S8];
-    }
+	if (((dram_address | rom_address | longueur) & 3) != 0)
```
Narann: I suggest a tiny comment after the if statement `check word-alignment` (or maybe it's obvious).

src/rsp/rsp_core.c
```diff
@@ -47,14 +47,25 @@ static void dma_sp_write(struct rsp_core* sp)
...
 
-    for(j=0; j<count; j++) {
-        for(i=0; i<length; i++) {
-            spmem[memaddr^S8] = dram[dramaddr^S8];
-            memaddr++;
-            dramaddr++;
-        }
-        dramaddr+=skip;
-    }
+	if (((memaddr | dramaddr | length | skip) & 3) != 0)
```
Narann: `check word-alignment` comment?

src/rsp/rsp_core.c
```diff
@@ -73,14 +84,25 @@ static void dma_sp_read(struct rsp_core* sp)
...
 
-    for(j=0; j<count; j++) {
-        for(i=0; i<length; i++) {
-            dram[dramaddr^S8] = spmem[memaddr^S8];
-            memaddr++;
-            dramaddr++;
-        }
-        dramaddr+=skip;
-    }
+	if (((dramaddr | memaddr | skip | length) & 3) != 0)
```
Narann: `check word-alignment` comment?

Narann:
> I have not extensively checked all roms so have left the safer byte-wise implementation in.

From the [manual](https://level42.ca/projects/ultra64/Documentation/man/pro-man/pro02/index2.4.html):

> All DMA transactions in the Nintendo 64 must use 64 bit aligned for data in RDRAM. DMA transactions for data in ROM must use 16 bit aligned addresses. )

So, to keep code clean, I suggest to remove old code and assume DMA are always 16 bits aligned.

What do you think?

wareya:
Depends on what the console actually does. Manuals are known to be "best practice for stability/speed" instead of "the hardware requires this".

I would keep the former behavior until someone pops in with a hardware test.

Narann:
Thinking about it, you're surely right, if it's technically possible we can't assume it will not happen...

richard42:
I'm a little hesitant to merge this because it increases the code complexity a fair amount, for a performance gain on the raspberry pi / arm architectures.  Do you have any quantitative data on the performance improvement from using this aligned memcpy in real world scenarios?  If it's just moving a few hundred or thousand bytes per VI then it's probably not worth it.  But if these loops are run for millions of iterations per VI then it would be.

ricrpi:
The performance increase is 1 to 2 % faster (timing to frame 501), depending on ROM on a Pi B. 
It isn't much of an improvement but given Zelda OOT intro is only running at ~70% full speed it is significant. 

It would be interesting to know if any ROMS require byte-wise copy as if none do then the word-wise only implementation would be simpler than the current byte-wise copy.

loganmc10:
There is merit to this. RSP DMA transfers must be 8-byte aligned. The N64 ignores the lower 3 bits of the source and destination address. It also ignores the lower 3 bits of the length (or rather, it assumes the lower 3 bits are all 1's, then it adds 1).

As long as we properly mask the source and destination addresses, the transfers should always be 8-byte aligned.

@ricrpi I don't know if you're still around. I'm far less confident in the alignment of the PI DMA's, but I am very confident in the 8-byte alignment of the RSP DMA's. If you update this PR to just to 8-byte aligned RSP DMA's (with no fallback) I'd be willing to do a fair bit of testing on it

ricrpi:
I am trying to keep up with all the commits/issues but have very limited time to do any coding at the moment :-(

I will try to take a look at this though as it shouldn't take too long to patch in.

PR #192, opened 2016-09-29
new_dynarec: Allocate memory for original MIPS source code of a recom…

Gillou68310:
…piled block before do_dirty_stub_ds is being called.

### DIFF
--- a/src/r4300/new_dynarec/new_dynarec.c
+++ b/src/r4300/new_dynarec/new_dynarec.c
@@ -10687,6 +10687,13 @@ int new_recompile_block(int addr)
   uint64_t is32_pre=0;
   u_int dirty_pre=0;
   #endif
+
+  copy=NULL;
+  copy=(char*)malloc((slen*4)+4);
+  assert(copy);
+  copy_size+=((slen*4)+4);
+  //DebugMessage(M64MSG_VERBOSE, "Currently used memory for copy: %d",copy_size);
+
   u_int beginning=(u_int)out;
   if((u_int)addr&1) {
     ds=1;
@@ -10912,12 +10919,6 @@ int new_recompile_block(int addr)
     }
   }
   // External Branch Targets (jump_in)
-  copy=NULL;
-  copy=(char*)malloc((slen*4)+4);
-  assert(copy);
-  copy_size+=((slen*4)+4);
-  //DebugMessage(M64MSG_VERBOSE, "Currently used memory for copy: %d",copy_size);
-
   for(i=0;i<slen;i++)
   {
     if(bt[i]||i==0)


fzurita:
I'm still experiencing some crashes on shutdown after this change. It seems very random though and the stack traces are useless.

Gillou68310:
With CBFD only? Or with other games too?

Gillou68310:
I made a few tests but I didn't experienced any crashes for now.

fzurita:
It seems to be any game. And it doesn't crash most of the time.

Gillou68310:
Any advice in order to reproduce it? What's your usual testing scenario?

fzurita:
I haven't found a reliable way to reproduce it. I have the feeling it had to do with which instruction the dynarec is in when it shuts down.

Gillou68310:
Just to be sure it's not a cache issue, could you add ```__clear_cache((void *)base_addr,(void *)base_addr+(1<<TARGET_SIZE_2));```
at the beginning of the ```new_dynarec_cleanup``` function?
Also try to put a ```DebugMessage``` instead of the ```assert(copy_size==0)```; to check if all dynamic allocations have been freed correctly.

fzurita:
Ok, will do.

Gillou68310:
Thanks ;-)

Gillou68310:
I'm still unable to reproduce a crash. I tried almost every game of my library.

fzurita:
Ok, maybe it's something we are doing wrong during shutdown with mupen64plus AE. Is that what you are using for testing?

Gillou68310:
Yep I'm using latest master + my patch (copy_fix branch)
Android frameworks is only used in gliden64 right? Maybe you can try another video plugin to see if you can reproduce the crash.

fzurita:
The android frameworks stuff it's crashing on is not from GLideN64, it's like OS level stuff. It could be related to Android 6.0 which is what I have on the devices I'm testing on. Which Android version are you testing on?

Gillou68310:
4.4.2

fzurita:
Ok, that could be the reason.

Also, by the way, I just updated to the latest Android NDK and it seems like GCC is deprecated and now they are forcing the use of the clang compiler.

I got almost everything compiling fine again (I had to make a few minor changes). The only thing I'm having problems with the x86 FPIC version of the mupen64plus-core. This file specifically:

https://github.com/mupen64plus/mupen64plus-core/blob/master/src/r4300/x86/rjump.c

I'm getting undefined references to "__i686.get_pc_thunk.bx" I tried to define it myself using guides I found on the internet, but when it do that, the x86 version of the core crashes when exiting a game on line ```save_edi=0;``` line 160.

Any ideas? Assembly is really not my strong point.

Gillou68310:
I'll take a look

fzurita:
I was able to force GCC 4.9 again by setting this in Application.mk
```
NDK_TOOLCHAIN_VERSION := 4.9
```

Eventually, we are going to have to deal with the mupen64plus core PIC version. It would be nice to make the new dynarec eventually be compatible with Android 5.0+.

fzurita:
I'm starting to think the remaining crash in short down must be an Android issue. I have another device here with Android 6.0 and I can't get it to crash on shut down. But I can't say for sure since it happens so rarely.

PR #196, opened 2016-10-12
Fix Pokemon Puzzle League

loganmc10:
Taken from https://github.com/libretro/mupen64plus-libretro/commit/9f071ad11e9d0d16ad10950cefd64a1727fa3d26

### DIFF
--- a/src/main/rom.h
+++ b/src/main/rom.h
@@ -124,6 +124,7 @@ typedef struct
    unsigned char savetype;
    unsigned char players; /* Local players 0-4, 2/3/4 way Netplay indicated by 5/6/7. */
    unsigned char rumble; /* 0 - No, 1 - Yes boolean for rumble support. */
+   unsigned char alternate_vi_timing;
    unsigned char countperop;
    uint32_t set_flags;
 } romdatabase_entry;
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -10892,6 +10892,7 @@ CRC=4A1CD153 D830AEF8
 Players=2
 SaveType=Flash RAM
 Rumble=No
+ViTiming=Alternate
 
 [A3BA044DFC00BB766B4B2BFB9D4B5BE9]
 GoodName=Pokemon Puzzle League (F) [!]
@@ -10899,6 +10900,7 @@ CRC=3EB2E6F3 062F9EFE
 Players=2
 SaveType=Flash RAM
 Rumble=No
+ViTiming=Alternate
 
 [000364BAC80E41D9060A31A5923874B7]
 GoodName=Pokemon Puzzle League (G) [!]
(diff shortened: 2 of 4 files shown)


src/main/rom.h
```diff
@@ -124,6 +124,7 @@ typedef struct
    unsigned char savetype;
    unsigned char players; /* Local players 0-4, 2/3/4 way Netplay indicated by 5/6/7. */
    unsigned char rumble; /* 0 - No, 1 - Yes boolean for rumble support. */
+   unsigned char alternate_vi_timing;
```
loganmc10: Ya of course

Narann: No worry, can you just write `fix Pokemon Puzzle League` and also test it actually fix the game? :D

richard42:
This fix is really suspicious.  How exactly is it supposed to fix anything, when it calculates some timing value, and rather than divide it by 1500 to get the result, it takes the number modulo 270?

loganmc10:
@richard42 you can read the discussion here : https://github.com/libretro/mupen64plus-libretro/issues/128

I didn't write it, but it works, Pokemon Puzzle League didn't start before and now it does

src/main/rom.h
loganmc10: It does fix the game I've tested it. I'd rather someone with push access to the repo added a comment rather than me submitting a pull request just for that since this is already merged

wareya:
It's a timing hack. m64p doesn't have very accurate timing in the first place, so having game-specific timing hacks when a game otherwise doesn't even boot doesn't seem like a bad idea.

orbea:
Well, given that the game doesn't work without the hack, it is probably a good temporary fix. It should be fixed properly eventually, but this is better than nothing.

Narann:
I got a warning on gcc:
```c
../../src/vi/vi_controller.c:45:17: warning: type of ‘alternate_vi_timing’ does not match original declaration
 extern unsigned alternate_vi_timing;
                 ^
../../src/main/rom.c:59:15: note: previously declared here
 unsigned char alternate_vi_timing = 0;
               ^
```
Can anybody fix this?

richard42:
ok, I cleaned this up a bit.  I still don't like it though.

loganmc10:
In case anyone is curious, it turns out this "alternate timing" is only really useful if the result is 0, for instance:

```
if (vi->alternate_timing)
    vi->regs[VI_CURRENT_REG] = (vi->delay - (vi->next_vi - cp0_regs[CP0_COUNT_REG])) % (NTSC_VERTICAL_RESOLUTION + 1);
if (vi->regs[VI_CURRENT_REG] != 0)
    vi->regs[VI_CURRENT_REG] = (vi->delay - (vi->next_vi - cp0_regs[CP0_COUNT_REG])) / vi->count_per_scanline;
```

Produces the same result, the game still has issues though, the FMV's are choppy and glitched.

PR #198, opened 2016-10-20
Allow setting a loadstate job before emulator is running.

Gillou68310:
If a loadstate job is set before emulator is running the loadstate will happen on the first safe state interrupt.


Narann:
Related to https://github.com/mupen64plus/mupen64plus-ui-console/pull/27

PR #199, opened 2016-10-26
Add using_tlb flag to savestates

Gillou68310:


fzurita:
Should we still allow loading the old style save state? Specially since the new version only applies to the new dynarec, there won't be a difference with other dynarec versions.

Instead of cancelling there loading of the save state, we could only read the extra field if the save state file version >= the new save state file version.

Gillou68310:
Old savestate are still compatible, I'm not cancelling there loading since the upper 16bits of the version number are the same. Only the lower 16bits changed.

> we could only read the extra field if the save state file version >= the new save state file version.

For now I'm only reading the extra field if the savestate file version == new savestate file version, but using the >= test might be a better idea if we bump the lower 16bits again.

PR #202, opened 2016-11-05
Make SDL usage optional in some files

loganmc10:
This is needed to make the savestate/cheat code work with GLupeN64 (RetroArch).

RetroArch doesn't use/need SDL, so this protects the SDL calls inside "USE_SDL"

It also exposes savestates_save_m64p/savestates_load_m64p, since RetroArch has its own slot management system.

I'm not sure about Visual Studio, somewhere in there it will need to define USE_SDL for this to work.

### DIFF
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -257,7 +257,7 @@ ifeq ($(origin SDL_CFLAGS) $(origin SDL_LDLIBS), undefined undefined)
       $(warning Using SDL 1.2 libraries)
     endif
   endif
-  SDL_CFLAGS  += $(shell $(SDL_CONFIG) --cflags)
+  SDL_CFLAGS  += $(shell $(SDL_CONFIG) --cflags) -DUSE_SDL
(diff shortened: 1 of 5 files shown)


loganmc10:
If it's preferred I can also make those ```#ifndef NO_SDL```s instead and I can just define NO_SDL in my project

richard42:
Can you please remove the extra macro definition from the makefile and do:
`#ifdef SDL_MAJOR_VERSION`
instead?

loganmc10:
Sure makes sense, I wasn't sure if that SDL config script set any preprocessor macros like that

wareya:
SDL_MAJOR_VERSION could be defined even if the downstream user doesn't want to use SDL. They would have to do macro magic or isolate different parts of the project in ways that they might not want to. An extra macro makes more sense to me because it doesn't make assumptions about the environment the downstream developer is compiling with.

richard42:
okay fair enough.  I'll merge it as-is

Issue #139, opened 2015-11-04
Configurable VI Refresh Rate


Gillou68310:
A few people on paulscode.com are asking about a configurable VI refresh rate option (like PJ64).
The main motivation is when using a lag remover cheat code (which typically disable ingame framelimiter), but such a cheat code should be reversed compensated by lowering VI Refresh Rate value in order to get back initial framerate.

Using this combo sometimes help fixing framerate issues in some games.

@richard42 @Narann do you mind if I add such an option?

Narann:
I mainly see such thing as more code to add and maintain to try to workaround thing on some games (and to be honest, I'm not even sure it will improve frame-rate).

You can't have an efficient _lag removing cheat_ effect with HLE without workaround. I'm just afraid the final code will be a bunch of nested if/if everywhere in the code while there is already many.

I don't know about PJ64. Have you any screenshot to share?

richard42:
Hi Gillou68310,

Can you explain a bit more about how this feature would work and what it would
do?  I assume that the main purpose is to just modify the VI timing interval
(real opengl frames drawn per second) and not do anything like changing
emulated CPU instructions per VI ala CounterPerOp, correct?  This is already
done for the Fast Forward feature, so simply adding an option to change the
base VI rate (which is currently set by the game: I think 50 for PAL / 60 for
NTSC) shouldn't be a big deal.  Would this be a user option in the config file
or a per-game option in the INI?

Richard

Gillou68310:
Thanks for your comments,

I guess codes speaks more than words, here's what I have in mind:
https://github.com/Gillou68310/mupen64plus-core/commit/d60eefba9262cb057c89cb1c5f464f494b90ec3c

I don't really need this option myself so I'm probably not the right person to defend it.
I invited people from paulscode.com to join this discussion and explain why they need this option.

richard42:
I see from this commit that it's a pretty simple option, and it's in the global config file.  It looks like this will change the number of r4300 instructions which get executed before the next VI fires.  Does anyone have experience with this setting, and can offer some examples of games which perform better with higher values than the default 1500?

Isotarge:
[...]

PR #204, opened 2016-11-08
Use custom ViRefresh for some games

loganmc10:
This is meant to resolve https://github.com/mupen64plus/mupen64plus-core/issues/139

These values were taken from Project64's RDB file:
https://github.com/project64/project64/blob/master/Config/Project64.rdb

I assume they came to those values via some testing, but this should probably receive some testing as well before it is merged.

For games that had a ViRefresh value I also made sure the CountPerOp (Counter Factor for PJ64) matched. That includes Beetle Adventure Racing, Indiana Jones and the Infernal Machine, and Nightmare Creatures


loganmc10:
By the way, the GitHub diff doesn't do a great job of accurately showing the changes to the ini file, you might have to click the little "Expand" button where it shows the line numbers to get a more accurate picture.

PR #207, opened 2016-11-16
Various refactorings

bsmiles32:
Various refactorings (no functional changes). Main features of this PR are :
- reduced usage of global variables throughout the various modules
- introduction of a device struct which should eventually contain all n64 device state.
- initialization of n64 components is done in 2 steps (init which configures static part of the component, and poweron which set component state like it would be at power on).
- extraction of various backends for external resources (audio, controller, clock, rumble, storage)
- unified management of save files (eep, fla, mpk, sra).


bsmiles32:
Not in mergeable state... I will need some time to fix it. Sorry

PR #208, opened 2016-11-17
Various refactorings

bsmiles32:
Various refactorings (no functional changes). Main features of this PR are :

* reduced usage of global variables throughout the various modules
* introduction of a device struct which should eventually contain all n64 device state.
* initialization of n64 components is done in 2 steps (init which configures static part of the component, and poweron which set component state like it would be at power on).
* extraction of various backends for external resources (audio, controller, clock, rumble, storage)
* unified management of save files (eep, fla, mpk, sra).


Narann:
Looks nicer, good job. I wonder about how much new indirection (like `g_dev.ri.regs`) would cost.

bsmiles32:
I think there is no new indirection, because all structures are embedded inside the dev one, so offsets are easy to compute. On the other hand if I had stored pointers to structs that would have led to extra indirection.

PR #210, opened 2016-11-25
Avoid processing CIC response pif ram as commands

raphnet:
When a CIC challenge took place in update_pif_write(), the pif ram
contains a bunch 0xFF followed by 0x00 at offsets 46 and 47. Offsets
48 onwards contains the challenge answer.

Then when update_pif_read() is called, the 0xFF bytes are skipped
up to the two 0x00 bytes that increase the channel to 2. Then the
challenge answer is (incorrectly) processed as if it were commands
for the third controller.

This cause issues with the raphnetraw plugin since it modifies pif ram
to store the result or command error flags. This corrupts the response
and leads to challenge failure.


loganmc10:
You are using tabs instead of spaces like the rest of the code for indentation it looks like

raphnet:
Oh, sorry. I'll correct this.

PR #212, opened 2016-11-28
Allow CountPerScan and ViTiming in ROM database to be overridden

fzurita:
Allow ViRefresh and ViTiming in ROM database to be overridden in mupen64plus.cfg

This makes it easier to search for a correct value for a ROM because we don't have to find the correct ROM in the database.

This was done in the same style as CountPerOp.

### DIFF
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -244,6 +244,8 @@ int main_set_core_defaults(void)
     ConfigSetDefaultString(g_CoreConfig, "SharedDataPath", "", "Path to a directory to search when looking for shared data files");
     ConfigSetDefaultBool(g_CoreConfig, "DelaySI", 1, "Delay interrupt after DMA SI read/write");
     ConfigSetDefaultInt(g_CoreConfig, "CountPerOp", 0, "Force number of cycles per emulated instruction");
+    ConfigSetDefaultInt(g_CoreConfig, "ViTiming", -1, "Use alternate VI timing (-1=Game default, 0=Don't use alternate timing, 1=Use alternate timing)");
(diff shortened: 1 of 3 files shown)


fzurita:
I just saw that this is probably going to conflict with pull request #208. Once #208 is merged, I can fix this pull request.

richard42:
Okay, I just merged #208.

fzurita:
I updated the pull request so that there is no longer a conflict.

src/r4300/r4300.c
```diff
@@ -59,6 +59,8 @@
 
 unsigned int r4300emu = 0;
 unsigned int count_per_op = COUNT_PER_OP_DEFAULT;
+int g_alternate_vi_timing = ALTERNATE_VI_TIMING_DEFAULT;
+int g_count_per_scanline = DEFAULT_COUNT_PER_SCANLINE;
```
bsmiles32: These don't need to be globals anymore. See previous comment.

src/r4300/r4300.h
```diff
@@ -41,9 +41,15 @@ extern unsigned int r4300emu;
 extern uint32_t next_interupt;
 extern uint32_t last_addr;
 #define COUNT_PER_OP_DEFAULT 2
+#define ALTERNATE_VI_TIMING_DEFAULT 0
+#define DEFAULT_COUNT_PER_SCANLINE 1500
```
bsmiles32: You can move these inside rom.c

fzurita:
Ok, I'll update and resubmit some time tonight.

fzurita:
Ok, I have updated the pull request to address the comments.

bsmiles32:
Thanks for the updated pr. Looks good to me.

fzurita:
@richard42 Anything preventing this from being merged?

PR #213, opened 2016-12-01
No globals

bsmiles32:
This PR depends upon what have been initiated in PR#208.

Main concern of this PR is to move all global variables related to the emulated system inside device (or one of its sub) structures. So that everything about the system state can be accessed from the g_dev global variable. Future work might go a little further and avoid the usage of the g_dev global with carefull passing of appropriate struct around. But for now, g_dev is heavily relied upon.

The main problem that was previously blocking such work is that assembly files can't use directly C struct fields. Assemblers only know about numerical offsets. To complicate the situation, C structure offsets are compiler/implementation dependent. So the only "sane" way to solve this is to make the compiler generate a file which will define these numerical offsets, and include it in assembly files. In practice, we make the compiler emit an assembly file from a crafted dummy C file and parse it with gawk.

Note however that **this PR is still incomplete** and I will need help on the following points:
- VisualStudio project file needs to be updated to trigger the generation of the asm_defines.h file.
This might also require the addition of (g)awk.exe (?) as a build dependency.
(bonus point if the asm_defines.h file is re-generated when structures it depends upon are changed).
- new_dynarec/ARM: @Gillou68310 any thought on how to proceed with this ?
- general testing to ensure no regression.

Also, I might not have much free time to work on that for the coming months, so expect slow progress.


Gillou68310:
> So the only "sane" way to solve this is to make the compiler generate a file which will define these numerical offsets, and include it in assembly files.

@bsmiles32 you could apply the same method for the ARM architecture. Currently the ARM backend has it's own structure defined in linkage_arm.S. The address of the structure ( ```&dynarec_local```) is loaded into the FP register in ```new_dyna_start``` so all load/store can use FP as the base address + the offset to the variable (```ldr    r0, [fp, #stop-dynarec_local]```). With your solution FP should be loaded with the address of ```g_dev``` (```ldr    r0, [fp, #offsetof_struct_device_r4300 + offsetof_struct_r4300_core_stop]```.

However there are a few limitations to take into account:

1) The beginning of your structure should start with a 64 bytes (16 registers * 4 bytes) free space used as a stack for pushing/popping registers.

2) All new_dynarec specific variables should be added to your structure.
-```int cycle_count;```
-```int last_count;```
-```int pending_exception;```
-```int pcaddr;```
-```char *invc_ptr;```
-```uint64_t readmem_dword;```
-```u_int rounding_modes[4];```
-```int branch_target;```
-```precomp_instr fake_pc;```
-```int ram_offset;```
-```u_int mini_ht[32][2];```
-```u_char restore_candidate[512];```
-```u_int memory_map[1048576];```

3) The offset value has a limited range, so all variables accessed by the new_dynarec should be as close as possible to the address of ```g_dev```.
-ldr/str 0-4095
-ldrb/strb 0-4095
-ldrh/strh 0-255 most contraining!!! (cpu_hword)

Be careful because the compiler will throw on error if an offset is out of range, but recompiled code won't. You can write a routine that load/store a value to all variables accessed by the new_dynarec to make sure all offsets are in a valid range.

Also on ARM64, 16bit variables should be 2 bytes aligned, 32bit variables should be 4 bytes aligned and 64bit variables should be 8 bytes aligned, on the other hand the offset range is usually bigger. Here's the structure definition for ARM64 https://github.com/Gillou68310/mupen64plus-core/blob/0fbadb4baf598850170a9da90e0b62be19d665f6/src/r4300/new_dynarec/arm64/linkage_arm64.S

Gillou68310:
Forgot to say ```memory_map``` is a special case and only the start address is accessed, no need to try accessing memory_map[1048575], it's not in valid offset range anyway ;-)
That's why it's at the end of the structure BTW.

bsmiles32:
Thanks for detailled answer.
As you outlined, the biggest issue is keeping variables used in linkage_arm close to fp. I can see 2 options here :
- do as previously and let linkage_arm do it's manual allocation, and just code a bunch of accessors functions to hide this from other modules. This might clutter the code with conditional compilation directives.
- or rework the r4300_core, cp0 cp1 and memory structs to ensure locality constrains are met. This would avoid the conditional compilation that we have now but we will loose the clear separation of the cp0, cp1 and memory modules. @Nebuleon already proposed a PR doing something like that 1 year ago.

To me both solutions are acceptables, so I'd like more opinions.

Gillou68310:
Well it's just a design decision. I trust your judgment ;-)

bsmiles32:
Rebased against master.

Finally, I added accessor wrappers to take care of the difference in variables allocation between the core and the new_dynarec ARM. I renamed some variables along the way. I didn't touch other variables in new_dynarec. This might be done in a later PR as this one is already big enough.

@Gillou68310: can you test the ARM stuff ? I hope I didn't break anything, but I cannot compile and test myself.

Is there someone with VisualStudio that can update the VisualStudio file to generate the asm_defines.h file ?

Thanks.

bsmiles32:
Updated with some minor fixes.

Gillou68310:
> @Gillou68310: can you test the ARM stuff ? I hope I didn't break anything, but I cannot compile and test myself.

@fzurita do you have time to do that? My android build setup is not really up to date.

I can take care of the visual studio update.

fzurita:
Yeah, sure, I'll test this some time today.

Gillou68310:
Thanks ;-)

fzurita:
Well, it's not building any more.

```
[armeabi-v7a] Compile arm    : mupen64plus-core <= memory.c
./mupen64plus-core/src/memory/memory.c: In function 'memory_address':
./mupen64plus-core/src/memory/memory.c:57:13: error: 'g_dev_mem_address' undeclared (first use in this function)
     return &g_dev_mem_address;
             ^
./mupen64plus-core/src/memory/memory.c:57:13: note: each undeclared identifier is reported only once for each function it appears in
./mupen64plus-core/src/memory/memory.c: In function 'memory_wbyte':
./mupen64plus-core/src/memory/memory.c:67:13: error: 'g_dev_mem_wbyte' undeclared (first use in this function)
     return &g_dev_mem_wbyte;
             ^
./mupen64plus-core/src/memory/memory.c: In function 'memory_whword':
./mupen64plus-core/src/memory/memory.c:77:13: error: 'g_dev_mem_whword' undeclared (first use in this function)
     return &g_dev_mem_whword;
             ^
./mupen64plus-core/src/memory/memory.c: In function 'memory_wword':
./mupen64plus-core/src/memory/memory.c:87:13: error: 'g_dev_mem_wword' undeclared (first use in this function)
     return &g_dev_mem_wword;
             ^
./mupen64plus-core/src/memory/memory.c: In function 'memory_wdword':
./mupen64plus-core/src/memory/memory.c:97:13: error: 'g_dev_mem_wdword' undeclared (first use in this function)
     return &g_dev_mem_wdword;
             ^
make: *** [C:/Users/Francisco/Documents/workspaces/mupen64plus-ae/app/build/intermediates/ndkBuild/debug/obj/local/armeabi-v7a/objs/mupen64plus-core/./mupen64plus-core/src/memory/memory.o] Error 1
make: *** Waiting for unfinished jobs....


* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```

Gillou68310:
The extern declaration is probably missing for these variables.

Could you check again by adding the following lines:
```
extern uint32_t g_dev_mem_address;
extern uint8_t g_dev_mem_wbyte;
extern uint16_t g_dev_mem_whword;
extern uint32_t g_dev_mem_wword;
extern uint64_t g_dev_mem_wdword;
```

fzurita:
Yep, I didn't see them defined anywhere. So I guess it's that macro that is defining them.

fzurita:
Ok, that was the problem in memory.c, unfortunately, the same type of problem repeats in multiple places.

fzurita:
Ok, I managed to fix all the similar errors, and I'm down to a few I'm not sure about.

In linkage_arm.S: ```tlb_LUT_r``` is undefined, what am I supposed to put there? I tried ```g_dev.r4300.cp0.tlb.LUT_r``` but it doesn't work.

In savestate_convert.c: There are still references to ```tlb_LUT_r```. Should they be ```g_dev.r4300.cp0.tlb.LUT_r```.

bsmiles32:
Huge thanks for the help both of you 😀
For savestate_convert you guessed correctly.
For linkage_arm.S it's harder because it is an assembly file. You have to use the same approach as linkage_x86.S and generate a gas-compatible asm_defines.h from the main/asm_defines.c by filtering its compilation with awk. If I have time i can try to do that part.

bsmiles32:
Ok, I updated the Makefile to generate gas-compatible asm_defines.h and included that in linkage_arm.S.
@fzurita : Feel free to push your fixes here :)

fzurita:
Ok, will do. Also, I just realized that savestate_convert.c is a separate tool not part of mupen64plus-core executable.

fzurita:
The pull request has been made to your branch.

fzurita:
Ok, one problem. main/asm_defines_gas.h doesn't exist in your branch.

bsmiles32:
main\asm_defines_gas.h should be automatically generated. See Makefile.
If it isn't, try typing from unix directory : make ../../src/main/asm_defines_gas.h
If it is, maybe it is just an include directory problem, so instead of #include "main/asm_defines_gas.h" in linkage_arm.S, try "../../src/main/asm_defines_gas.h"

bsmiles32:
@fzurita : My bad, the auto generation didn't work because I typed linkage_arm.asm instead of linkage_arm.S ...

fzurita:
@bsmiles32 I'm building for Android, which has its own custom makefile, I'll see if I can incorporate the generation of that file.

fzurita:
I'm having problems understanding what the Makefile is supposed to be doing to generate asm_defines_gas.h. What's the specific command to generate the file?

bsmiles32:
The command is:

$(SRCDIR)/main/asm_defines_gas.h: $(SRCDIR)/main/asm_defines.c
	$(GENS.c) -fno-lto $< -o - | awk '($$1 == "->") { print "#define " $$2 "_" $$3 " " substr($$4,2) }' > $@
With 
GENS.c = $(Q_CC)$(CC) $(CFLAGS) $(CPPFLAGS) $(TARGET_ARCH) -S

What it does is that it generates assembler from asm_defines.c using the same compilation flags as the rest of the program (exception for lto). This output is filtered and reformatted to only retain offsets defines.

You can generate the asm_defines_gas.h individualy by typing in the unix directory:
make ../../src/main/asm_defines_gas.h
You can inspect the generated file in src/main/asm_defines_gas.h

bsmiles32:
Fixed some regression which prevented pokemon stadium from booting.

bsmiles32:
@fzurita, @Gillou68310: any progress ?

fzurita:
I have not gotten around to generating that file since I have to setup a Linux environment. It's the only thing that doesn't work for me. Unfortunately, it's been a busy couple of days for me there last few days. I'll try to get to it today.

Gillou68310:
Sorry I've been busy with other things too. I'll try to work on this before the end of the week.

bsmiles32:
@Gillou68310: I've made a PR to add gawk in win32 deps repo as it will likely be needed to generate the asm_defines_*.h headers.

bsmiles32:
Updated msvc file to help generate the asm_defines_nasm.h file.
I can't test that myself (no win/msvc), but it should be a step in the right direction. Please test and report if it works.

fzurita:
Ok, I'm starting to get somewhere generating this file in windows...

First of all these are not defined for the ```NEW_DYNAREC_ARM```
```
DEFINE(struct r4300_core, regs);
DEFINE(struct r4300_core, hi);
DEFINE(struct r4300_core, lo);
DEFINE(struct r4300_core, stop);
DEFINE(struct memory, wbyte);
DEFINE(struct memory, whword);
DEFINE(struct memory, wword);
DEFINE(struct memory, wdword);
DEFINE(struct memory, address);
```

Second, commenting those out temporarily still produces errors:
```
main/asm_defines.c: Assembler messages:
main/asm_defines.c:20: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:29: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:30: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:32: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:34: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:37: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:38: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:39: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:41: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:42: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:43: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:45: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:46: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:48: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:56: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:57: Error: junk at end of line, first unrecognized character is `-'
main/asm_defines.c:58: Error: junk at end of line, first unrecognized character is `-'
```

Not sure what those mean.

bsmiles32:
I don't think you can generate this file in Windows for the arm new dynarec. Just x86 and x64. And I guess I will also have to change the DEFINE macro to be compatible with msvc... Oups.

fzurita:
I'm using the Android cross compiler. Either way, I get the same errors when building the Windows version. I'm probably missing something.

Here is the command I'm using: 
```
BASE_DIR=`pwd`
echo "Generating asm defines"
cd "${BASE_DIR}"/app/src/main/jni/mupen64plus-core/src
x86_64-linux-android-g++ -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes \
 -IC:/Users/Francisco/Documents/workspaces/mupen64plus-ae/app/src/main/jni/./mupen64plus-core/src \
 -IC:/Users/Francisco/Documents/workspaces/mupen64plus-ae/app/src/main/jni/png/include/  \
 -IC:/Users/Francisco/Documents/workspaces/mupen64plus-ae/app/src/main/jni/SDL2/include/ \
 -IC:/Users/Francisco/android-sdks/ndk-bundle/build//../sources/cxx-stl/gnu-libstdc++/4.9/include \
 -IC:/Users/Francisco/android-sdks/ndk-bundle/platforms/android-15/arch-arm/usr/include \
 -IC:/Users/Francisco/android-sdks/ndk-bundle/build//../sources/cxx-stl/gnu-libstdc++/4.9/libs/armeabi-v7a/include \
 -IC:/Users/Francisco/android-sdks/ndk-bundle/build//../sources/cxx-stl/gnu-libstdc++/4.9/include/backward \
 -IC:/Users/Francisco/Documents/workspaces/mupen64plus-ae/app/src/main/jni/SDL2/include \
 -IC:/Users/Francisco/Documents/workspaces/mupen64plus-ae/app/src/main/jni \
 -DANDROID -O3 -ffast-math -fno-strict-aliasing -fomit-frame-pointer -frename-registers -fvisibility=hidden -DANDROID  \
 -DIOAPI_NO_64 -DNOCRYPT -DNOUNCRYPT -DUSE_GLES=1 -DDYNAREC -DNEW_DYNAREC=1 -Wa,--noexecstack -Wformat -Werror=format-security  \
 -fno-lto main/asm_defines.c > main/asm_defines_gas.h

```

bsmiles32:
Add -S -o - to force generation of asm file on stdout and pipe it to awk to extract and generate defines.
g++ ... -S -o - | awk '($$1 == "->") { print "%define " $$2 "_" $$3 " " substr($$4,2) }' > main/asm_defines_nasm.h

bsmiles32:
Also, I've just read that msvc doesn't support inline assembly for x64 and arm, so I will have to find an alternative way of generating the asm_defines .... :(

fzurita:
That's a shame about the lack of support by visual studio.

bsmiles32:
There is still the option of turning asm_defines.c into a separate executable that prints the desired asm_defines_nasm.h . If compiled with the very same options as m64p core, it should work and avoid the usage of inline assembly trick. It might also avoid the dependency on awk. In fact, I might try that option un the coming days :)

bsmiles32:
I've investigated the possibility of turning asm_defines into a separate executable, but that would be problematic for cross-compilation scenario. My other idea is to make asm_defines export global variables such as char offsetof_struct_xxx[] whose size is the value of the offset. That way I can retrieve these values easily with objdump/dumpbin and awk. And that would still work un cross-compile scenario.

bsmiles32:
Updated with previously mentioned changes.
Please help refine the msvc project file so we can move on, I have some cool stuff to push after this PR :)

fzurita:
This looks interesting. I see you are using objdump now in the Makefile. I'll have to see if it exists in the Android environment.

Gillou68310:
@fzurita sure, objdump is part of the ndk toolchain!

Gillou68310:
```error C2466: cannot allocate an array of constant size 0```

@bsmiles32 I get the above error in VS2013 with the following lines:

 ```
DEFINE(device, r4300);
DEFINE(r4300_core, regs);
DEFINE(cp0, regs);
DEFINE(tlb, entries);
DEFINE(cached_interp, invalid_code);
 ```

The offset from the beginning of the structure is 0 for the above defines and VS2013 doesn't allow allocating an array of size 0.

**Comment by bsmiles32** [comment]
My Bad ! GCC didn't complained 🙁. I've checked and it is indeed not valid c. A hacky workaround would be to add 1 to allow sizes in asm_defines and subtract 1 in objdump/awk. Or maybe you have other ideas to defines symbols with given size ?

**Comment by bsmiles32** [comment]
Updated with the +1 -1 trick to avoid zero-sized arrays. I also cleaned a little the implementation (awk patterns, offsetof from stddef, don't hardcode programs in Makefile).

**Comment by Gillou68310** [comment]
I have another issue, the generated object file is not passed to dumpbin when using the following command ```cl /c /FoCON %(FullPath) | dumpbin.exe /SYMBOLS```

What does the CON value means next to the /Fo option?

**Comment by Gillou68310** [comment]
However using the following command works:

```cl /c %(FullPath) & dumpbin.exe /SYMBOLS asm_defines.obj```

**Comment by Gillou68310** [comment]
> What does the CON value means next to the /Fo option?

Nevermind, printing to console.

**Comment by Gillou68310** [comment]
https://github.com/Gillou68310/mupen64plus-core/commit/9890a39f3143f02c23be594bf609b5586a01e38f

Build fine but crashes at startup using 32bit new dynarec, I need time to investigate.

**Comment by Gillou68310** [comment]
Is it working on linux?

**Comment by bsmiles32** [comment]
64 bit works but I cannot test 32 bit due to missing libraries. Can you get a backtrace of the crash with debug symbols ? Does it crashes with all emumodes ( interpreter, cached , dyna) ?

**Comment by bsmiles32** [comment]
Also, I think I gave you push access to this branch, do you can directly updated it 😎

**Comment by Gillou68310** [comment]
It only crashes with new dynarec, the crash happens in the recompiled code so the backtrace is useless.

Don't worry I will cherry pick my commits to your branch when I'm finished ;-)

**Comment by bsmiles32** [comment]
Can you also check that the asm_defines_nasm.h has sensible values ?

**Comment by bsmiles32** [comment]
And that asm_defines.c is compiled with the same cflags as the rest of the project (except for GlobalOptimization which must be disabled).

**Comment by Gillou68310** [comment]
Here's the generated header file if you want to take a look:
[asm_defines_nasm.txt](https://github.com/mupen64plus/mupen64plus-core/files/677120/asm_defines_nasm.txt)

Here's the cflags used in the project for the New_Dynarec_Debug config:
```/GS /analyze- /W3 /Zc:wchar_t /I"..\..\src" /I"..\..\..\mupen64plus-win32-deps\SDL-1.2.15\include" /I"..\..\..\mupen64plus-win32-deps\zlib-1.2.8\include" /I"..\..\..\mupen64plus-win32-deps\libpng-1.6.18\include" /I"..\..\..\mupen64plus-win32-deps\freetype-2.6\include" /ZI /Gm /Od /Fd"Debug\vc120.pdb" /fp:precise /D "WIN32" /D "_DEBUG" /D "_WINDOWS" /D "_USRDLL" /D "_CRT_SECURE_NO_DEPRECATE" /D "DYNAREC" /D "M64P_OSD" /D "M64P_PARALLEL" /D "NEW_DYNAREC=1" /D "_WINDLL" /D "_MBCS" /errorReport:prompt /WX- /Zc:forScope /RTC1 /Gd /Oy- /MDd /Fa"Debug\" /EHsc /nologo /Fo"Debug\" /Fp"Debug\mupen64plus.pch"```

Since we are using a custom compile command none of these flags are inherited while compiling asm_defines.c and I don't know if there's a way to include them other than doing it manually.

**Comment by bsmiles32** [comment]
Your asm_defines_nasm is for 32bit new_dynarec (x86), right ?
Because when I generate it from my desktop (GCC, make BITS=32 NEW_DYNAREC=1 all) I have some different values (for instance for r4300_core_stop which is 0x128 instead of 0x12c for you). Can you try passing manually all the same CFLAGS to see if it helps ?

**Comment by Gillou68310** [comment]
> Your asm_defines_nasm is for 32bit new_dynarec (x86), right ?

Yep.

> Can you try passing manually all the same CFLAGS to see if it helps ?

I just tried and no difference in the generated header file.

**Comment by Gillou68310** [comment]
Ok I found the cause:

The following line should be replaced:
```emit_readword((int)r4300_cp1_regs_double()[(source[i]>>16)&0x1f],temp);```

by:
```emit_readword((int)&r4300_cp1_regs_double()[(source[i]>>16)&0x1f],temp);```

in c1ls_assemble.

You'll have to change it for all reference to r4300_cp1_regs_double/simple within assem_x86/arm.c and new_dynarec.c.

**Comment by bsmiles32** [comment]
Well done !
Does it mean that with these changes everything works (msvc support, new dynarec x86 and arm) ?
/Me is very happy

**Comment by Gillou68310** [comment]
I didn't apply the fix right now so I don't know if everything else is working.

**Comment by Gillou68310** [comment]
Will do tomorrow!

**Comment by fzurita** [comment]
So I'm trying to generate asm_defines_gas.h in the Android environment. But I'm running into issues. Here is my script so far. Ignore the local paths, I plan to replace those with environment variables:

[generate-core-asm-defines.zip](https://github.com/mupen64plus/mupen64plus-core/files/678346/generate-core-asm-defines.zip)

Here is the generated objdump file:
[asm_defines_obj_dump.zip](https://github.com/mupen64plus/mupen64plus-core/files/678348/asm_defines_obj_dump.zip)

And here is the generated asm_defines_nasm.h:
[asm_defines_nasm.zip](https://github.com/mupen64plus/mupen64plus-core/files/678350/asm_defines_nasm.zip)

Is there something I'm missing in my script?

**Comment by bsmiles32** [comment]
Try adding -C to objdump (to demangle symbols as you compile in C++ not C),
and change in awk print the $$1 into $$5 as the symbols size seems to appear in fifth column not the first.
Also you might need to change the field separator to space (ie add -F " " to awk).

**Comment by bsmiles32** [comment]
Also for nasm it's %define not #define

**Comment by fzurita** [comment]
Ok, it's looking better. But still not compiling. Here is the updated objdump for arm:

https://github.com/mupen64plus-ae/mupen64plus-ae/blob/bsmiles_noglobals/app/src/main/jni/mupen64plus-core/src/asm_defines_obj_dump_arm.txt

Here is the generated header file for arm:
https://github.com/mupen64plus-ae/mupen64plus-ae/blob/bsmiles_noglobals/app/src/main/jni/mupen64plus-core/src/main/asm_defines_gas.h

**Comment by bsmiles32** [comment]
awk doesn't seem to correctly parse your columns, have you set the field separator -F " " as suggested ?

**Comment by fzurita** [comment]
Yep. My columns seem to be separated by tabs. I'm looking up the man page on awk.

**Comment by bsmiles32** [comment]
Try -F "\t" ?

**Comment by fzurita** [comment]
No luck. The cygwin version of awk seems to be acting weird.

**Comment by fzurita** [comment]
I think I figured it out. awk doesn't like the ```$$```. It works correctly with just 1 $.

**Comment by fzurita** [comment]
ARM compiled fine... Now I have to figure out why x86 is not compiling...

**Comment by bsmiles32** [comment]
Seems strange because inside shell scripts I would have expected doubling the $ mandatory, but if it works for you that's great !
So does it compile now and runs ? :)

**Comment by bsmiles32** [comment]
Have you changed #define into %define as suggested ?

**Comment by fzurita** [comment]
Yes, I did change #define to %define. It doesn't like this: ```%include "../../src/main/asm_defines_nasm.h"``` inside the .asm files. I need to figure out proper relative path. Unfortunately, the way my environment is setup, I can't run the ARM version until I get the x86 version compiling.

**Comment by fzurita** [comment]
Ok, I tried to run it, it crashed.

**Comment by fzurita** [comment]
Here is the core dump:
```
********** Crash dump: **********
Build fingerprint: 'ZTE/P996A01/ailsa_ii:6.0.1/MMB29M/20161014.114339:user/release-keys'
pid: 24192, tid: 24961, name: CoreThread  >>> org.mupen64plusae.v3.alpha <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
#00 pc 000178f0  /system/lib/libc.so (memset+48)
#01 pc 00041c54  libmupen64plus-core.so: Routine poweron_r4300 at ./mupen64plus-core/src/r4300/r4300_core.c:50 (discriminator 1)
#02 pc 0000f35c  libmupen64plus-core.so: Routine poweron_device at ./mupen64plus-core/src/main/device.c:76
#03 pc 00012afc  libmupen64plus-core.so: Routine main_run at ./mupen64plus-core/src/main/main.c:1014
#04 pc 00003ee1  libmupen64plus-ui-console.so (SDL_main+3568): Routine SDL_main at ./mupen64plus-ui-console/src/main.c:782
#05 pc 007f6e2d  /data/app/org.mupen64plusae.v3.alpha-1/oat/arm/base.odex (offset 0x4f9000) (int paulscode.android.mupen64plusae.jni.NativeExports.emuStart(java.lang.String,java.lang.String, java.lang.Object[])+144)
#06 pc 007e6be7  /data/app/org.mupen64plusae.v3.alpha-1/oat/arm/base.odex (offset 0x4f9000) (void paulscode.android.mupen64plusae.jni.CoreInterface$4.run()+2930)
#07 pc 02108839  /system/framework/arm/boot.oat (offset 0x20eb000)
```

It seems to have crashed in this line in r4300_core.c
```
memset(r4300_regs(), 0, 32*sizeof(int64_t));
```

**Comment by fzurita** [comment]
All I can think of is that ```#define offsetof_struct_device_r4300 (0x00000001-1)``` is not giving me the correct memory address.

**Comment by bsmiles32** [comment]
offsetof_struct_device_r4300 should be zero because it is the offset of r4300 member inside a struct device.
So here it is OK. I think there might be a problem when you compile both x86 and ARM within the same directory, some defines might not play well together, such as NEW_DYNAREC being sometimes  equal to and different from NEW_DYNAREC_ARM.

**Comment by fzurita** [comment]
OK, can you double check this file? https://github.com/mupen64plus-ae/mupen64plus-ae/blob/bsmiles_noglobals/app/src/main/jni/mupen64plus-core/src/main/asm_defines_gas.h

I can compile and run this in Android with the regular mupen64plus-core. I'm not sure why only this branch would be an issue.

**Comment by bsmiles32** [comment]
The gas file looks good to me for the ARM.

**Comment by fzurita** [comment]
OK, maybe @Gillou68310 will have an idea. I'll check again tomorrow.

**Comment by bsmiles32** [comment]
Ok ! Have some good sleep, we made great progress already :)

**Comment by Gillou68310** [comment]
I confirm that the new_dynarec x86 is working on windows since your last commit. My android setup is at work so I won't be able to work on the ARM backend until monday, sorry for the delay :-(

**Comment by bsmiles32** [comment]
No worries ! Can you push your updated msvc project files here ?

If I understand correctly after we get confirmation that the ARM new dynarec works, this pr will be mergeable ? 😀

@richard42 Any suggestions on this ?

**Comment by fzurita** [comment]
I found the problem I was having with the ARM version. This problem is in numerous places, here is one example though, some of this may have been my doing. In r4300_core.c ```extern int64_t* g_dev_r4300_regs;``` should be ```extern int64_t g_dev_r4300_regs;```

And
```
int64_t* r4300_regs(void)
{
#if NEW_DYNAREC != NEW_DYNAREC_ARM
/* ARM dynarec uses a different memory layout */
    return g_dev.r4300.regs;
#else
    return &g_dev_r4300_regs;
#endif
}

fzurita:
Ok, I'm running now without core dumping in the ARM version. I'll update your branch.

fzurita:
@Gillou68310 Can you check that the generate-core-asm-defines.sh script works correctly in a linux environment?

bsmiles32:
I took care of the ASFLAGS for upstream with this commit.
You can do the same for your fork now.

fzurita:
It works correctly now without any modifications.

Gillou68310:
> @Gillou68310 Can you check that the generate-core-asm-defines.sh script works correctly in a linux environment?

I'm working on windows too :-p.

@bsmiles32 I cleaned up the new dynarec code a bit, I didn't succeed pushing on your branch but you can cherry pick my work from my local branch https://github.com/Gillou68310/mupen64plus-core/commits/no_globals

The asm_defines_gas header for ARM is not necessary anymore but I left it in place just in case we need it in the future.

bsmiles32:
@Gillou68310: I've sent you a collaboration invitation so you can push here. In your branch there are 2 commits, one for msvc file and one for new dynarec cleanings. Is the cleaning part of this PR ? Because if not, I would prefer it to be part of a separate pr as this one is already big enough.

To summarize the state of this PR:

* Makefile and msvc files have been updated
* Both 32/64 bits versions are working
* New Dynarec x86/ARM is also working

Is that right ? Then it should be pretty soon ready for merge (eg After adding Gillou's commit(s))

Gillou68310:
Ok I pushed the updated msvc files.
I will open another PR for the new dynarec code cleanup part.

loganmc10:
I don't mean to sound rude, and I'm not sure who all has push access to this repo, but perhaps that list could be expanded? Many PR's seem to get submitted to mupen64plus that just sort of sit there for weeks or months without much action or discussion...

bsmiles32:
There are several devs with merge rights, myself included. Since I'm proposing the PR I'd prefer someone else to merge it, because last time I merged m'y own PR I got bitten by regressions 🙁 I'm not sure why @richard42 hasn't responded yet, but let's hope it's nothing too serious.

fzurita:
It's around Christmas/new year's time. Maybe he went on vacation.

richard42:
Sorry I've been out of it lately.  I was out of town for several weeks at the end of December, and I've been busy with everything else.  My inbox is overflowing with Mupen64plus stuff to review; I'll get to it before long. :)

PR #215, opened 2016-12-06
Fix assertion failure when starting Kuiki Uhabi Suigo

Gillou68310:
The parameter of isspace should be in the range 0-255.

### DIFF
--- a/src/main/util.c
+++ b/src/main/util.c
@@ -262,10 +262,10 @@ char *trim(char *str)
 {
     char *start = str, *end = str + strlen(str);
 
-    while (start < end && isspace(*start))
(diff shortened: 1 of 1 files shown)


richard42:
wouldn't it be better just to change the pointer definitions to unsigned char * than to typecast the dereferenced chars?

Gillou68310:
Do you mean casting the pointer to ```(unsigned char *)``` before dereferencing it?
```*((unsigned char*)start)```

If you mean changing ```char *start = str;``` to ```unsigned char *start = str;``` it means implicit cast of ```str``` from ```char*``` to ```unsigned char*``` I would rather use explicit cast.

PR #227, opened 2017-01-22
Device refactorings

bsmiles32:
This PR depends on the work done in PR 213 (no globals).
Main concern of this PR is to move all the emulation logic inside the device directory.
Also backends are partly reworked to ease their usage in the upcoming gameboy/transferpak support.


bsmiles32:
Rebased against master.

Issue #228, opened 2017-01-30
Savestates in memory


fr500:
I was wondering if you could consider something like this:
https://github.com/fr500/GLupeN64/commit/0bd03fcacf7aab024fb1097c96dd1962b36bcf33

Basically I want to have a way to not compress or flush save data to disk at all for our libretro port.
@loganmc10 is trying to keep his fork upstream friendly so if you would provide this functionality in upstream we could all have our frontend features (rewind, netplay) and keep it friendly.

Let me know!

fr500:
I duplicated everything so it's easier to pull, I'm not asking this to be merged or anything.
maybe I could just change the function signature to allow both codepaths to work in one function?

PR #229, opened 2017-01-30
Allow libretro to save states to memory

loganmc10:
Fixes https://github.com/mupen64plus/mupen64plus-core/issues/228

### DIFF
--- a/src/main/savestates.c
+++ b/src/main/savestates.c
@@ -87,6 +87,9 @@ struct savestate_work {
     char *data;
     size_t size;
     struct work_struct work;
+#ifdef __LIBRETRO__
+    void *mempointer;
+#endif
 };
 
 /* Returns the malloc'd full path of the currently selected savestate. */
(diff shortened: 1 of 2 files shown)


bsmiles32:
Thanks for your contribution !
The deep  issue this pr addresses is that upstream saves/loads to/from file (and compress/uncompress) savestates. While convenient for upstream, this is not desirable for libretro, which would benefit from savestates to memory.
Your proposed solution, while acceptable in the short term, will complicate long term maintainance (lots of ifdef).
What I personnally would prefer for the long term is to split the serialization logic from the actual destination (memory for libretro, compressed file for m64p). That way would avoid the ifdef and improve upstream code 🙂

rlabrecque:
I agree with bobby, I'd rather this be accepted but having a cleanup pass on it would be even better.

loganmc10:
No problem, you can hold off merging this until I clean it up/make it less intrusive

Issue #230, opened 2017-02-01
fails to build since 5aebce38204de4fe3671f78e3cee66838f9a1d87


charlemagnelasse:
The build fails since 5aebce38204de4fe3671f78e3cee66838f9a1d87

     ../../src/main/rom.c: In function 'romdatabase_open':
     ../../src/main/rom.c:449:38: error: 'DEFAULT' undeclared (first use in this function)
                  search->entry.savetype = DEFAULT;

See https://travis-ci.org/mupen64plus/mupen64plus-core/builds/197207056

charlemagnelasse:
@bsmiles32 please have a look

bsmiles32:
This is due to some merging oops in commit  beb8ad3. See my comment there

richard42:
this should be fixed now.

PR #231, opened 2017-02-01
Fix for #230

loganmc10:


richard42:
thanks, but I fixed it a different way.  Having a macro called DEFAULT just seems like a bad idea.

PR #237, opened 2017-02-09
Add ConfigExternalGetParameter

loganmc10:
This adds a new API function: ConfigExternalGetParameter.

Basically you pass it the path to an ini file, the section, and the parameter name, and it returns the value. This will allow plugins to leverage the already existing ini parser built into mupen64plus-core.

GLideN64 will use this to read it's custom ini file with game specific settings.


richard42:
I like the simplicity of what you've done here, and I like the fact that you updated the documentation, but there are a few problems with this.
1. It opens and parses the entire file to return a single parameter, so reading all the parameters in a file will be O(N^2).  If there aren't very many parameters then it won't be a problem but if a plugin had a very large config file (like the INI file with rice video) it will be a performance issue.
2. You are returning a const char * into a char buffer which is allocated on the stack (char buffer[256]).  That is a bug.  You could strdup() it but the documentation would need to make clear that the caller would be responsible for freeing the returned string.
3. With any change in the core API, we need to update the version number (src/main/version.h) for the modified API, and the API versioning documentation (in Mupen64Plus-v2.0-API-Versioning.mediawiki)

loganmc10:
Thanks for the feedback

1. I agree it is very inefficient, but making it more efficient would require future extensions to the API (ConfigOpenExternal/CloseExternal or something like that), and some kind of tracking system in the core to keep track of the open files. Since INI parsing should happen before the start of a game, I don't think performance is a huge concern, there shouldn't be a reason for a plugin to parse a config file during any time where performance would be of consequence 

2. Thanks, I don't code in C much. Would it be better to return an m64p_error and have the plugin pass a  char* into the function? Then it should be obvious that they are in charge of managing that memory, but I could also add a note to the doc.

3. Thanks I can do that as well.

Papermanzero:
Any update on this one?
Many users would like to see the features to load the config file for GlideN64.
If Bullet 2 and 3 are resolved I would propose to merge it.
Bullet 1 can be improved with future revisions and PRs. 
At least for GlideN64 you need anyway performant hardware.

loganmc10:
Ok this PR has been updated to address points 2 & 3. The function now returns an m64p_error. And a new variable, ParamPtr, is passed into the function, which is a pre-allocated char array created by the plugin that should be at least 256 bytes. The value is passed into that pointer via strncpy.

src/api/config.c
```diff
@@ -612,6 +612,49 @@ m64p_error ConfigShutdown(void)
...
+                else
+                    foundSection = 0;
+                break;
+            case INI_PROPERTY:
+                if (foundSection)
+                {
+                    if (osal_insensitive_strcmp(ParamName, l.name) == 0)
+                    {
+                        fclose(fPtr);
+                        strncpy(ParamPtr, l.value, 256);
```
Gillou68310: This is dangerous in case of the dev didn't read the doc and pass a pointer to a tab whose size is less than 256 bytes.
One solution is to add an additional input parameter for the size of the input tab and pass it to strncpy.

But IMO the best solution is to do what @richard42 proposed and return a dynamically allocated string, the caller would be responsible for freeing the returned string.

loganmc10: Isn't the risk the same? What if the dev didn't read the doc and doesn't know they are responsible for freeing the returned string?

Gillou68310: Memory leak is always best than memory corruption ;-)

loganmc10: haha I didn't realize there was a hierarchy :P 

To me it makes more sense to pass the function a variable I allocated. It seems strange to me as a plugin dev to free a pointer that I never malloc'ed.

Generally I think it's fair to assume that a dev should read the docs if they are going to implement an API. I think passing a size variable would be fine, but then you are passing a lot of variable into this function, and you are just protecting against a situation where someone didn't read the docs.

Gillou68310: From my experience devs never read the docs :-)

loganmc10: I think anytime you are passing a pointer into a function, expecting the function to populate the pointer, you have to have an idea of what the function is going to try to put in that pointer.

For instance, if I expected the function to populate it with an Int, I wouldn't pass it a Char. In this case, the expected value is a 256-byte array. I don't think it's unreasonable to expect the dev to figure out (by reading the docs) what kind of pointer should get passed into the function.

Gillou68310: Another solution would be to allocate a static tab in config.c and return a pointer to it. The problem is that you need to allocate a tab which is big enough to fill your string. Is 256 bytes the maximum possible size of the string?

loganmc10: Well I suppose that is up for debate, we have to pick some max size either way. I would assume the longest value that would get returned from a config file would be a "file path", like a setting that specifies the path to a font file or texture file or something. So I guess you could set the max length to PATH_MAX, isn't that defined on Linux and Windows?

Gillou68310: PATH_MAX isn't always defined and since we need to check if it's defined or not, I think we better use an hardcoded value. I guess 4096 should be enough, this is the usual value of PATH_MAX on linux.

loganmc10: So is a static variable of length 4096 in config.c an acceptable solution? I'm just struggling to get this merged here lol

Gillou68310: For me it is. We still have to wait for @richard42 opinion :-)

richard42: Developers who don't read the documentation are not very good. ;)  I agree in principle that a memory leak is better than memory corruption.  Apple's Cocoa API has a convention in the method names whereby it's clear whether or not the caller is expected to destroy an object that gets returned, but we have no such convention.  Using a static buffer to return to the caller is an okay solution but it's not thread-safe.  If the front-end application was using threads and calling this function at the same time as a plugin, it could cause trouble.  I think the best overall solution is to add an extra parameter, ParamMaxLength or something like that, to specify the size of the buffer being passed in by the caller.

loganmc10:
Ok this has been updated to add a "ParamMaxLength" variable, which is used to allocate the buffer passed to ```ini_parse_line()```. It is also used as the max value passed to ```strncpy```

src/api/config.c
```diff
@@ -612,6 +612,49 @@ m64p_error ConfigShutdown(void)
 /* Selector functions, exported outside of the Core */
 /* ------------------------------------------------ */
 
+EXPORT m64p_error CALL ConfigExternalGetParameter(const char *FileName, const char *SectionName, const char *ParamName, char* ParamPtr, int ParamMaxLength)
+{
+    FILE *fPtr;
+    char buffer[ParamMaxLength + 64]; //ParamMaxLength is the size of the value, we add 64 to make sure we have enough room to store the rest of the line.
```
Gillou68310: This is an implicit dynamic allocation, you should use malloc/free instead. Also I'm not sure to understand why adding 64 is necessary?

loganmc10: I must not understand C at all, doesn't the variable get destroyed automatically when the function ends?

"buffer" is supposed to hold the contents of the entire line. ParamMaxLength is specifying the max legth of the value. So if the line is:

device = carrot

They only need to specify the max length of "carrot", but our buffer needs to hold "device = carrot"

Gillou68310: > I must not understand C at all, doesn't the variable get destroyed automatically when the function ends?

Local variables are allocated on the stack (or in a register) at the beginning of the function and de-allocated at the end of the function, this is done at compile time. http://www.avabodh.com/cin/memorymanagement.html

The problem here is that the compiler doesn't know the value of ParamMaxLength at compile time, that's why you have to use dynamic allocation malloc+free. I'm actually surprised the the compiler didn't complained about this.

In this situation I would rather use a fixed length for buffer, being the max size of a line, maybe 4096 like we discussed before?

loganmc10: Ok I'll change it to malloc/free

> In this situation I would rather use a fixed length for buffer, being the max size of a line, maybe 4096 like we discussed before?

What if the plugin specifies a value for ParamMaxLength that is larger than 4096?

Gillou68310: Let's use malloc ;-)

loganmc10:
Ok it's been updated to use malloc/free

src/api/config.c
```diff
@@ -612,6 +612,53 @@ m64p_error ConfigShutdown(void)
...
+        return M64ERR_INPUT_INVALID;
+    if (FileName == NULL || (fPtr = fopen(FileName, "rb")) == NULL)
+    {
+        DebugMessage(M64MSG_ERROR, "Unable to open config file '%s'.", FileName);
+        return M64ERR_INPUT_INVALID;
+    }
+
+    char *buffer = malloc(ParamMaxLength + 64); //ParamMaxLength is the size of the value, we add 64 to make sure we have enough room to store the rest of the line.
+
+    while (fgets(buffer, sizeof(buffer), fPtr) != NULL)
```
Gillou68310: You need to replace ```sizeof(buffer)``` by ```(ParamMaxLength + 64)``` and I still don't understand why you're adding 64. 

Anyway I was checking how ```ini_parse_line``` was used in the rest of the code and I like the way it's used in ```ConfigInit```. Instead of reading the file line by line, the entire file is read into a dynamically allocated buffer (```sizeof(file length)```). The buffer is than parsed from beginning to end using the ```ini_parse_line``` function.

Sorry for asking you to amend this commit again and again :-(

loganmc10: No problem, I don't want to cause any memory leaks or anything. I've added a commit to have it work like it does in ConfigInit, it does look like a much better way to do it.

I've tested it with reading GLideN64.custom.ini and it still works so that's a good sign :)

Gillou68310: I like it, nice work.

Gillou68310:
I wonder if the api should be responsible for opening/closing the file, maybe we should pass the file pointer instead of the filename. What do you guys think?

loganmc10:
That would certainly improve performance, since right now it is reading the whole file for every individual parameter, but then you need to keep track of file pointers, etc... Because multiple plugins could theoretically be opening multiple config files, etc..

loganmc10:
I don't mind doing that but I'd like some guidance before I do, so I don't end up rewriting it.

I would add ```ConfigExternalOpen(file path, &myHandle)``` and ```ConfigExternalClose(myHandle)```. ConfigExternalGetParameter would stay the same (from the perspective of the plugin), except the plugin would pass myHandle instead of the filepath, does that seem reasonable?

loganmc10:
Alright I swallowed my pride and created a ConfigExternalOpen() and ConfigExternalClose()

ConfigExternalOpen() reads the whole file and copies the contents into a buffer it allocates, which is pointed to by a m64p_handle, it returns the size of the buffer.

ConfigExternalGetParameter() takes that pointer and parses through the buffer to read the parameters.

ConfigExternalClose() free's the buffer created by ConfigExternalOpen()

This should perform a lot better since it only reads the file once. It's still not quite as high performance as the native config file, since it still parses through the whole buffer every time a parameter is read, but it's reading memory instead of disk, so it will be much faster.

rlabrecque:
It's easy enough to turn that into like a hash table lookup later on. That sounds like an improvement.

Papermanzero:
any update? Can the PR be merged?

Gillou68310:
I'm thinking about an alternative implementation for this, I'll try to propose something by the end of the week.

loganmc10:
Any update on this?

Papermanzero:
Just a proposal: Why not merging this PR and optimising the feature step by step in later points in time. At least the the feature is available and can be used. So experiences can be collected.

loganmc10:
Just checking in again. If there is anything that needs changing, I don't mind changing it (if someone can think of a better way to handle this, just explain it to me, I don't mind doing the coding).

Gillou68310:
Sorry I didn't find time to work on this these days. Anyway here's my remarks:

1. From what I understand this API will be used to parse a database not a config file, maybe the "ConfigExternal" name is not appropriate.

2. I don't like the idea of passing the file size to "GetParameter". One solution is to create a linked list containing all sections from the database and possibly a hash table to speed up section lookup.

3. The linked list could be created in the "Open" API, returning the address into the handle.

4. "GetParameter" will look for the section name into the linked list and extract the parameter from the section. We can also add a type parameter (see ConfigGetParameter) for more flexibility?

loganmc10:
1. Yeah I guess it's more aimed at parsing INI files, I don't really care what it's called to be honest, so I'm open to any name suggestion.

2. I agree the size thing was a bit clunky. I've modified the PR so that it allocates a struct now, with one item being a pointer to the file contents, and the other item being the size of the file, so it no longer needs the API to pass that info along.

> We can also add a type parameter (see ConfigGetParameter) for more flexibility?

ConfigGetParameter works by passing the data long to ConfigGetParamInt/ConfigGetParamFloat/ConfigGetParamBool/ConfigGetParamString. So if we went this route, we would likely have to add 4 additional functions. It seemed like this would be simpler, the plugin knows what kind of data it is expecting, and it can convert the data from a string to whatever it wants.

Gillou68310:
1. Nevermind ConfigExternal is just fine

2. I'm okay with this solution for now but I really think we should create a list when opening the file to avoid parsing the whole file every time GetParameter is called. Anyway this can be done later as it will have no impact on the current API definition.

> ConfigGetParameter works by passing the data long to ConfigGetParamInt/ConfigGetParamFloat/ConfigGetParamBool/ConfigGetParamString. So if we went this route, we would likely have to add 4 additional functions. It seemed like this would be simpler, the plugin knows what kind of data it is expecting, and it can convert the data from a string to whatever it wants.

I agree but the 4 functions could be static to config.c so it will not add complexity to the API.

loganmc10:
I guess I disagree here. I don't really see the point of making the plugin pass the type in as an argument (the function already has 5 arguments). It's pretty simple for the plugin to just take the string and do atoi or atof if they want to convert it.

Gillou68310:
I just want to avoid changing the API later if someone thinks that the convertion should be done by the API ;-)

loganmc10:
Well I agree that the API shouldn't be changed later, the function should stay the same after it gets merged. I guess I'll leave that one up to @richard42 , whatever he wants.

Issue #195, opened 2016-10-11
arm new dynarec doesn't build with clang


loganmc10:
The Android NDK is going to get rid of GCC soon, they are just going to use clang. I can't seem to compile mupen64plus-core using clang/Android NDK (r13)

```
   ldr r3, [fp, #g_cp0_regs+48-dynarec_local]
                  ^
mupen64plus-core/src/r4300/new_dynarec/arm/linkage_arm.S:265:19: error: constant expression expected
    ldr r4, [fp, #g_cp0_regs+16-dynarec_local]
                  ^
mupen64plus-core/src/r4300/new_dynarec/arm/linkage_arm.S:267:19: error: constant expression expected
    str r0, [fp, #g_cp0_regs+56-dynarec_local]
                  ^
mupen64plus-core/src/r4300/new_dynarec/arm/linkage_arm.S:269:19: error: constant expression expected
    str r1, [fp, #g_cp0_regs+32-dynarec_local]
                  ^
mupen64plus-core/src/r4300/new_dynarec/arm/linkage_arm.S:271:19: error: constant expression expected
    str r3, [fp, #g_cp0_regs+48-dynarec_local]
                  ^
mupen64plus-core/src/r4300/new_dynarec/arm/linkage_arm.S:273:19: error: constant expression expected
    str r2, [fp, #g_cp0_regs+52-dynarec_local]
                  ^
mupen64plus-core/src/r4300/new_dynarec/arm/linkage_arm.S:275:19: error: constant expression expected
    str r1, [fp, #g_cp0_regs+40-dynarec_local]
                  ^
mupen64plus-core/src/r4300/new_dynarec/arm/linkage_arm.S:277:19: error: constant expression expected
    str r4, [fp, #g_cp0_regs+16-dynarec_local]
                  ^
mupen64plus-core/src/r4300/new_dynarec/arm/linkage_arm.S:460:20: error: constant expression expected
    str r10, [fp, #cycle_count-dynarec_local]
                   ^
mupen64plus-core/src/r4300/new_dynarec/arm/linkage_arm.S:462:20: error: constant expression expected
    ldr r10, [fp, #cycle_count-dynarec_local]
                   ^
mupen64plus-core/src/r4300/new_dynarec/arm/linkage_arm.S:466:19: error: constant expression expected
    str r8, [fp, #branch_target-dynarec_local]
                  ^
[...]

PR #240, opened 2017-02-19
Avoid g_dev in r4300 code

bsmiles32:
Small PR which avoids usage of the g_dev global variable inside TLB code.

This a small step toward a "g_dev free" emulation code. By that, I mean, emulation code (code residing inside the device directory or any of its subdirectory) should hopefully one day not rely directly on the global variable g_dev. But instead access this structure through pointers. At least, that's what I'm trying to do in the long term.

### DIFF
--- a/RELEASE
+++ b/RELEASE
@@ -363,7 +363,7 @@ what's new in 0.5:
       feature. Mariokart's monitor in first race running fullspeed and
       puzzle effect in banjo's intro are two examples that i can think
       about but there are many more.
-    + detection of VI interupt rate works on weird country codes
+    + detection of VI interrupt rate works on weird country codes
     + better detection of self modifying code in dma
     + warnings fixed on new gcc versions
     + rsp's dmem and imem are now contiguous in PC's memory (some windows
@@ -510,7 +510,7 @@ what's new in 0.1:
   - Core
     + sram bug zelda oot fixed
     + flashram is working
-    + a new interupt system that'll enabled
+    + a new interrupt system that'll enabled
       more accurate timing in the future but can cause
       some compatibility issues right now
     + bug in DMULT/DMULTU opcode fixed
--- a/doc/new_dynarec.mediawiki
+++ b/doc/new_dynarec.mediawiki
@@ -186,7 +186,7 @@ As in the original mupen64plus, the emulated clock runs at 37.5 MHz, and each in
 
 ==Interrupt handler==
 
-When the cycle count register reaches its limit, cc_interrupt is called, which in turn calls gen_interupt [sic].  If interrupts are not enabled, cc_interrupt returns.  If interrupts are enabled, and an interrupt is to be taken, the pending_exception flag will be set.  In this case, cc_interrupt does not return, and instead pops the stack and causes an unconditional jump to the address in pcaddr (usually 0x80000180).
+When the cycle count register reaches its limit, cc_interrupt is called, which in turn calls gen_interrupt.  If interrupts are not enabled, cc_interrupt returns.  If interrupts are enabled, and an interrupt is to be taken, the pending_exception flag will be set.  In this case, cc_interrupt does not return, and instead pops the stack and causes an unconditional jump to the address in pcaddr (usually 0x80000180).
 
 There is one additional case where the interrupt handler may be called.  If interrupts were disabled, and are enabled by writing to coprocessor 0 register 12, any pending interrupts are handled immediately.
 
--- a/projects/VisualStudio2013/mupen64plus-core.vcxproj
+++ b/projects/VisualStudio2013/mupen64plus-core.vcxproj
@@ -124,7 +124,7 @@
     </ClCompile>
     <ClCompile Include="..\..\src\device\r4300\exception.c" />
     <ClCompile Include="..\..\src\device\r4300\instr_counters.c" />
-    <ClCompile Include="..\..\src\device\r4300\interupt.c" />
+    <ClCompile Include="..\..\src\device\r4300\interrupt.c" />
     <ClCompile Include="..\..\src\device\r4300\mi_controller.c" />
     <ClCompile Include="..\..\src\device\r4300\new_dynarec\arm\arm_cpu_features.c">
       <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">true</ExcludedFromBuild>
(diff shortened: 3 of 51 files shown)


bsmiles32:
Added some more r4300 refactorings. With these additional changes r4300_core and cached_interp files relies much less on the g_dev global variable.

bsmiles32:
Added some more refactorings to avoid direct usage of g_dev global variable.

bsmiles32:
Okay, I've done most of the easy passing of structures inside the r4300 module. I'll very likely wait a bit more before making further changes, so if possible I'd like to get that merged soonish.

fzurita:
@bsmiles32 something is wrong in the last set of commits. The core is crashing during shutdown. Here is the backtrace:
```
Build fingerprint: 'ZTE/P996A01_N/ailsa_ii:7.1.1/NMF26F/20170301.161705:user/release-keys'
pid: 12220, tid: 12404, name: CoreThread  >>> org.mupen64plusae.v3.alpha <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0xc3fc15f8
#00 pc 00038d9c ./libmupen64plus-core.so: Routine free_blocks at ../mupen64plus-core/src\device/r4300/cached_interp.c:607
#01 pc 0004589c ./libmupen64plus-core.so: Routine run_r4300 at ../mupen64plus-core/src\device/r4300/r4300_core.c:189
#02 pc 0000f5c4 ./libmupen64plus-core.so: Routine main_run at ../mupen64plus-core/src/main/main.c:1116
#03 pc 00004903 ./libmupen64plus-ui-console.so (SDL_main+3738): Routine SDL_main at ../mupen64plus-ui-console/src/main.c:798
```

The odd thing is the is crashing in the cached interpreter code, but I'm running in dynarec mode. I tried adding some logging to try to determine how it was getting there, and the back trace completely changed....

fzurita:
Why are we initializing and cleaning up cached interpreter blocks in the dynarec anyways? Although, I don't see how that could cause the crashing.

bsmiles32:
Didn't experienced that. Can you bisect it ?
I'm unfortunately away for a couple of weeks...

Gillou68310:
This is not happening on windows either. So this seems to be an ARM specific issue.

I see 3 possible causes:
-The ```r4300``` pointer has been corrupted. Could you print the value of ```r4300``` before and after ```new_dyna_start```?
-The context has been corrupted. Could you print the content of ```dynarec_local``` (64 bytes) before and after ```new_dyna_start```?
-A compiler issue? Could you print the result of an ```objdump -d``` on ```libmupen64plus.so.2.0.0```?

loganmc10:
I was testing things on my Raspberry Pi (ARM) the other day and I never ran into this, perhaps it's something clang is doing?

fzurita:
It didn't happen before these set of commits, so it's probably something wrong in the code somewhere.

loganmc10:
I'll try compiling with clang on my Pi today and see what I get

Gillou68310:
One thing you can test is to use the gdev struct instead of using the r4300 pointer passed in the function parameter, if this fixes the issue it's very likely to be a compiler issue. This can only be confirmed by looking at the disassembly.

loganmc10:
I couldn't get this to compile with clang, I ran into this issue: https://github.com/mupen64plus/mupen64plus-core/issues/195. Android must pass some parameters to clang when it's used with ndk-build that gets around that problem or something.

Anyway I compiled with gcc and printed out the value of the r4300 pointer before and after new_dyna_start():

Core: Init new dynarec
Core: ARM CPU Features:, Half, Thumb, FastMult, VFP, ESDP, NEON, VFPv3, TLS, VFPv4, IDIVa, IDIVt
**address before 0x715745b8**
Core Status: Stopping emulation.
**address after 0x724921dc**
Core: R4300 emulator finished.
Core Status: Rom closed.

Here is the diff from r4300_core.c so you can make sure I did this right:

```
 #ifdef NEW_DYNAREC
         new_dynarec_init();
+printf("address before %p\n",r4300);
         new_dyna_start();
+printf("address after %p\n",r4300);
         new_dynarec_cleanup();
 #else
         dyna_start(dynarec_setup_code);
```

Gillou68310:
It doesn't looks good, could you do an ```objdump -d``` on ```libmupen64plus.so.2.0.0```?

fzurita:
By the way, the Android NDK compiler is smart enough to use the proper arguments based on the file extension. So compilation is failing probably because of some missing arguments needed to compile assembly.

Gillou68310:
You may have to specify that linkage_arm.S in an assembly file ```-x assembler```

Gillou68310:
Try changing the following in the unix makefile:

```
$(OBJDIR)/%.o: $(SRCDIR)/%.S $(SRCDIR)/asm_defines/asm_defines_gas.h
	$(COMPILE.c) -o $@ $<
```

```
$(OBJDIR)/%.o: $(SRCDIR)/%.S $(SRCDIR)/asm_defines/asm_defines_gas.h
	$(COMPILE.c) -o -x assembler $@ $<
```

Gillou68310:
You could also try to change linkage_arm.S to linkage_arm.s updating the makefile consequently.
You can check if the assembly file is parsed correctly directly by building from command line.
Something like ```clang -c linkage_arm.s``` or ```clang -x assembler -c linkage_arm.s```

loganmc10:
> It doesn't looks good, could you do an objdump -d on libmupen64plus.so.2.0.0?

[objdump.txt](https://github.com/mupen64plus/mupen64plus-core/files/859063/objdump.txt)

Gillou68310:
Woaaa the disassembly looks awfull it seems like function have been inlined. Could you try to build with -fno-inline?

loganmc10:
> Could you try to build with -fno-inline?

[objdump3.txt](https://github.com/mupen64plus/mupen64plus-core/files/859162/objdump3.txt)

Gillou68310:
Thanks, I guess it's time to resurrect my rpi2 to look at this more in details.

Gillou68310:
Ok I checked on my rpi2 and the saved context in ```dynarec_local``` is being overwritten during the new dynarec execution, I think I know where this is happening. The problem was invisible before because the corrupted register (r4) was not used within the ```run_r4300``` function, but looking at the objdump r4 is now used to store the r4300 pointer, thus the crash ;-)

fzurita:
Nice, it's good when a refactor reveals some bugs.

Gillou68310:
I opened the PR https://github.com/mupen64plus/mupen64plus-core/pull/255

PR #242, opened 2017-02-22
Fix excessive warnings when compiling using clang.

fzurita:
Just a couple of warning fixes. This generates a lot of output when building due to the amount those macros are used.

### DIFF
--- a/src/device/r4300/new_dynarec/new_dynarec.c
+++ b/src/device/r4300/new_dynarec/new_dynarec.c
@@ -64,12 +64,20 @@ static void nullf() {}
 #if ASSEM_DEBUG
     #define assem_debug(...) DebugMessage(M64MSG_VERBOSE, __VA_ARGS__)
 #else
+#ifdef __clang__
+    #define assem_debug(...)
+#else
     #define assem_debug nullf
+#endif
 #endif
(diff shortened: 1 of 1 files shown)


Gillou68310:
This is generating excessive warnings with MSVC:
```
warning C4550: expression evaluates to a function which is missing an argument list
```

Using the following fixes the warning:
```
#define assem_debug(...) nullf(__VA_ARGS__)
#define inv_debug(...) nullf(__VA_ARGS__)
```

Could you check with clang?

fzurita:
I just tried that.... Unfortunately, that generates excessive warnings in clang again.... Maybe a compiler specific macro is in order?

fzurita:
I updated the pull request so the compiler type is check at compile time. If using clang, it will use a different macro.

Gillou68310:
What about: ```#define assem_debug(...)```
This will avoid jumping to the nullf function every time assem_debug and inv_debug are called, saving some instructions at the same time and hopefully fix the warnings with clang.

fzurita:
There are no warnings if the nullf is removed, I'll update the pull request.

Gillou68310:
Cool :-)

Issue #236, opened 2017-02-06
regression: slowdowns since 0814daaad842868d8283f686b80245186c425280


krnlyng:
I am experiencing noticeable slowdowns in almost all games.
git bisect says it is 0814daaad842868d8283f686b80245186c425280 which sounds fishy to me (i will try and bisect again sometime soon).
But anyways if i use the apply_speed_limit implementation as follows:
```
static void apply_speed_limiter(void)
{
    unsigned int CurrentFPSTime;
    static unsigned int LastFPSTime = 0;
    static float VITotalDelta;
    static float VIDeltas[64];
    static unsigned int VIDeltasIndex;

    double VILimitMilliseconds = 1000.0 / g_dev.vi.expected_refresh_rate;
    double AdjustedLimit = VILimitMilliseconds * 100.0 / l_SpeedFactor;  // adjust for selected emulator speed
    float ThisFrameDelta, IntegratedDelta, TimeToWait;

    timed_section_start(TIMED_SECTION_IDLE);

#ifdef DBG
    if(g_DebuggerActive) DebuggerCallback(DEBUG_UI_VI, 0);
#endif

    // if this is the first frame, initialize our data structures
    if(LastFPSTime == 0)
    {
        LastFPSTime = SDL_GetTicks();
        VITotalDelta = 0;
        memset(VIDeltas, 0, sizeof(VIDeltas));
        VIDeltasIndex = 0;
        return;
    }

    // calculate # of milliseconds that have passed since the last video interrupt
    CurrentFPSTime = SDL_GetTicks();
    ThisFrameDelta = CurrentFPSTime - LastFPSTime - AdjustedLimit;

    // are we too fast?
    if (ThisFrameDelta < 0)
    {
        // calculate the total time error over the last 64 frames
        IntegratedDelta = VITotalDelta  + ThisFrameDelta;
        // if we are still too fast, and then speed limiter is on, then we should wait
        if (IntegratedDelta < 0 && l_MainSpeedLimit)
        {
            TimeToWait = (IntegratedDelta > ThisFrameDelta) ? -IntegratedDelta : -ThisFrameDelta;
            DebugMessage(M64MSG_VERBOSE, "    apply_speed_limiter(): Waiting %ims", (int) TimeToWait);
            SDL_Delay((int) TimeToWait);
            // recalculate # of milliseconds that have passed since the last video interrupt,
            // taking into account the time we just waited
            CurrentFPSTime = SDL_GetTicks();
[...]

PR #250, opened 2017-03-05
Don't reset the speed limiter when falling behind

fzurita:
Don't reset the speed limiter when falling behind. Instead adjust the time. This prevents unwanted resets when the game is falling behind on slower devices. These resets cause unneeded sleeps on slower devices which make the game run even slower.

See:

https://github.com/mupen64plus/mupen64plus-core/issues/236

This code change was taking from https://github.com/mupen64plus-ae/mupen64plus-ae/blob/d5c5f321548fc4d7a1d72d19364671e3fcc207a5/app/src/main/jni/mupen64plus-audio-sles/main.cpp

A similar fix was made there by me.

### DIFF
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -760,6 +760,7 @@ void new_frame(void)
     }
 }
 
+#define SAMPLE_COUNT 3
 static void apply_speed_limiter(void)
 {
     static unsigned long totalVIs = 0;
@@ -768,6 +769,8 @@ static void apply_speed_limiter(void)
     static unsigned int StartFPSTime = 0;
     static const double defaultSpeedFactor = 100.0;
     unsigned int CurrentFPSTime = SDL_GetTicks();
+    static double sleepTimes[SAMPLE_COUNT];
+    static unsigned int sleepTimesIndex = 0;
 
     // calculate frame duration based upon ROM setting (50/60hz) and mupen64plus speed adjustment
     const double VILimitMilliseconds = 1000.0 / g_dev.vi.expected_refresh_rate;
@@ -805,12 +808,31 @@ static void apply_speed_limiter(void)
     {
        resetOnce = 0;
     }
-    else if(sleepTime > 0.0 && l_MainSpeedLimit)
-    {
-       DebugMessage(M64MSG_VERBOSE, "    apply_speed_limiter(): Waiting %ims", (int) sleepTime);
 
-       SDL_Delay((int) sleepTime);
+    sleepTimes[sleepTimesIndex%SAMPLE_COUNT] = sleepTime;
+    sleepTimesIndex++;
+
+    int elementsForAverage = sleepTimesIndex > SAMPLE_COUNT ? SAMPLE_COUNT : sleepTimesIndex;
+
+    // compute the average sleepTime
+    double sum = 0;
+    unsigned int index;
+    for(index = 0; index < elementsForAverage; index++)
+    {
+        sum += sleepTimes[index];
     }
+
+    double averageSleep = sum/elementsForAverage;
+
+    int sleepMs = (int)averageSleep;
+
+    if(sleepMs > 0 && sleepMs < maxSleepNeeded*SpeedFactorMultiple && l_MainSpeedLimit)
+    {
+       DebugMessage(M64MSG_VERBOSE, "    apply_speed_limiter(): Waiting %ims", sleepMs);
+
+       SDL_Delay(sleepMs);
+    }
+
 
     timed_section_end(TIMED_SECTION_IDLE);
 }


loganmc10:
Any comment on this?

@krnlyng perhaps you can give this a try and see if it fixes your performance issues?

loganmc10:
By the way @fzurita does a similar fix need to be applied to audio-sdl? Or does it look ok?

wareya:
The sdl audio backend just uses audio backend synchronization, not a timer. Which is a problem because it stalls the emulator as a whole and causes microstuttering.

fzurita:
Yes to what wareya said. It's a different mechanism.

loganmc10:
@fzurita for mupen64plus-ae have you considered just disabling the core speed limiter if the audio plugin is in use, would the audio plugin be able to keep proper speed just by itself?

wareya:
You can do that but if you do you should make the audio plugin smarter about stalling for the backend to want more audio samples so it doesn't make bad stutter. As is, the framerate limiter can cause audio samples to drop if the framerate limiter's timer runs slightly slower than the audio backend, but it makes gameplay smoother despite that.

loganmc10:
Looking through the mupen64plus-ae code, it seems that if "audioSynchronize" is enabled, then it uses the audio plugins frame limiter, and if that is disabled, then it uses the cores frame limiter, am I understanding that right @fzurita?

fzurita:
Yes, this is correct.

richard42:
yeah this is a fundamental problem.  The grand speed control mechanism needs to be redesigned to make this work properly.  I have a bunch of notes on what I want to do with this, but it will take a fair amount of work.

fzurita:
Can this be merged? It should fix slow down when the host machine can't keep up with emulation. I think it would be good until a better fix comes around.

loganmc10:
I'll do some testing on this tomorrow to make sure there aren't any performance regressions. It would be nice if @krnlyng could verify whether this solves their problem or not.

loganmc10:
I just tested this on my Linux laptop using Rogue Squadron (I ran it using the Pure Interpreter at a high resolution, so it runs about half speed for me).

I didn't notice any difference, good or bad. The FPS was the same in both scenarios for me. I'm using an audio plugin that doesn't have a speed limiter, so the core limiter is the only one in use.

I also tested by making the emulator run for 1000 new_vi() calls. I counted how many seconds total SDL_Delay ran for. With the current master, it waited a total of 2302 ms. With this change, it waited a total of 2386. If I let it run for 2000 VI's, the current master waited 3576 ms, this change was 3717 ms.

They both seem to have an issue when the emulator is speeding up. Like if it was running at 10 FPS, and it went to 20-30 FPS, that's when it would start waiting, in that transition from slower to faster (but not full) speed.

fzurita:
Try setting the game speed to 300% and then compare FPS.

loganmc10:
In that case the current master waited a total of 268 ms, and this changed waited 199 ms.

Once thing I'll also note, they both actually call SDL_Delay(0) sometimes. The current master actually calls that quite frequently. I assume this isn't intended behavior, but sleepTime is a double cast to an int, so any value from 0.00000001 to 0.9999999 is makes it makes the ```sleepTime > 0.0``` test and then SDL_Delay(0) gets called.

fzurita:
That's ok, it will just sleep less the next time around.

loganmc10:
Yeah but why are we calling SDL_Delay(0)? Shouldn't it just get changed to ```if sleepTime > 1.0```, then SDL_Delay(1) would be the minimum sleep. The SDL docs don't even specify what SDL_Delay(0) does.

fzurita:
That's going to be up to the OS. I believe if you look at the implementation, it's either calling ```usleep``` or ```nanosleep``` under the hood. If you look up the behavior for those when being called with zero time, you will get your answer.

wareya:
It's undefined so the code shouldn't do it.

fzurita:
nanosleep with zero time is allowed by Linux last time I looked (is within the valid range), so it shouldn't be undefined. I'm not sure how long the sleep will be though. I suspect that is OS dependent.

wareya:
If SDL doesn't define how it's supposed to behave in its spec, that means it's literally undefined. Undefined doesn't mean "undefined behavior". It's **very bad** to use behaviors that a library's spec doesn't define, no matter what a particular version is doing under the hood right now on a particular platform.

loganmc10:
usleep and nanosleep are Unix functions. On Windows it probably uses Sleep() and this is what Sleep() does with 0:

> A value of zero causes the thread to relinquish the remainder of its time slice to any other thread that is ready to run. If there are no other threads ready to run, the function returns immediately, and the thread continues execution.

It's not really what we want. I think 1 should be the minimum sleep.

As for the change in this PR, it seems fine to me, it does seem to reduce the amount of time slept. When I tested Rogue Squadron, the current speed limiter seemed to insert sleeps intermittently even when things were slow. With this change, it only seemed to insert sleeps when thing were running at 60FPS, or when the speed was increasing (like going from 10FPS to 30FPS)

loganmc10:
@wareya it's not really undefined behavior. The docs say ```It waits at least the specified time, but possibly longer due to OS scheduling.``` So you can take that to mean "It waits at least 0 seconds, maybe more". But we know that under the hood it calls usleep or nanosleep or Sleep or some other function, and it seems that "0" is a special case in some of those functions.

fzurita:
Yeah, that's a good point. SDL forums are saying that Sdl_delay of zero is equivalent to yield. Still, not what we want, but harm should be minimal. Still, I will update the pull request so they we only sleep if sleep is above zero.

wareya:
The definition is "Use this function to wait a specified number of milliseconds before returning." The "This function waits a specified number of milliseconds before returning. It waits at least the specified time, but possibly longer due to OS scheduling." is in the remarks section. The remark is just saying that the definition doesn't wait an exact number of milliseconds, but it shouldn't be a lot shorter. The definition doesn't really define how time passes, just what the purpose of the function is.

fzurita:
Pull request updated to not sleep at all if sleep time is 0.

krnlyng:
hi, sorry for the delay. i will test it as soon as possible :)

loganmc10:
> gles2n64 drops significantly more frames in this implementation

Are you talking about the current master? Or the changes that @fzurita has proposed? Or both?

> speed limiter i mentioned in #236 the game still runs faster than with this implementation

I guess the question for now is, does this change from @fzurita improve things at all vs how it is currently? It seems unlikely that someone is going to re-write the speed limiter from scratch in the near future, so if this improves things then it should probably be merged.

krnlyng:
"this implementation" refers to the latest patch from @fzurita, this patch/implementation makes it worse than current master, so in short:
legacy speed limiter is faster than master and master is faster than this patch.
i think the problem is that the game is not doing any sudden bursts of speed anymore so if it reaches a situation where it is slow it will stay slow (and it doesn't seem to recover). does this make sense?
In any case i tested my theory and removed the entire: if(sleepTime < minSleepNeeded) { StartFPSTime -= minSleepNeeded; }
and the game runs (actually too) fast ;) probably because it is almost never sleeping anymore. So in some sense the patch improves it by doing the reset only if -5000 is reached but it makes it worse on the other hand by removing the sudden bursts and keeping the game slow the entire time.

i hope we can fix this issue before it gets merged into master, in any case i will test more.

krnlyng:
I have a proposal: https://bpaste.net/show/823cf9090a7e

This speed limiter averages out the sleepTime. This has the effect that we don't: "wait nothing at all and in the next frame too much" which was happening in my case. It also takes care of smoothing out the transitions between slow and fast. It is not optimized though but if somebody of you can test it too it i would prepare it.

Why has the speed limiter has been changed? I am sure there were good reasons. I will compare my limiter to the one in #236 too.
My version works well on my phone and it works well on my desktop computer (bad case vs good case).

I am not saying that my version is ideal but it makes the games very much playable even on my slow device :).

wareya:
AFAIK the old speed limiter "leaked" over time, the new one doesn't.

fzurita:
@krnlyng maybe your clock is not reporting elapsed time accurately?

wareya:
Clocks are quantized. For example most of the ways of measuring time on windows are multiples of 5ms or 16ms.  Unless the program manages to get a high resolution clock, which isn't always possible, it's going to have a hard time avoiding stutter because of that.

krnlyng:
@fzurita i am pretty sure it does, i've printed the values in every iteration during my testing and the seem to make sense (Obviously when doing performance tests i removed that code again) but i can do in-depth tests tomorrow. note that we can smooth out the transitions even more byincreasing the sample count and we can remove the need for the memmove by using a smallbring buffer.

loganmc10:
To be honest, the implementation by @krnlyng (which is similar to the old speed limiter) makes a lot more sense to me. It makes more sense to take the moving average of the frames and using that to calculate the current speed, rather than taking the cumulative total starting from a (rather arbitrary) reset point.

wareya:
If you do that, you "leak" fractions of seconds over time. Calculating the desired *Point In Time* when the next frame should start is the right way to write a framerate limiter. The thing is that you have to derive that point in time in a way that doesn't "leak" over time, and if your frametimes aren't measured in an integral unit (for example, if they're a repeating decimal in the computer's number system: 16.6666...ms), or you average them together somehow in your calculations, you're going to lose or gain fractions of milliseconds and eventually skew slightly away from your target framerate.

Using an "arbitrary" reset point (not arbitrary at all, really, just the last time the framerate limiter went from not being needed to being needed) and seeking forwards in time relative to it, you keep numerical precision errors from building up over time, since the error happens for the entire span of frames since then, instead of for each frame individually.

krnlyng:
the limiter i proposed only computes the moving average over the last 60 frames currently. the actual sleep time computation i have taken from fzuritas patch/master. the average over the last 60 frmes only avoids jumps between negative and positive sleep times in case the hardware is too slow. but we should always converge to the intended framerate.

fzurita:
@krnlyn Your idea has merit. The way you have it now, I don't think will drift over time. By averaging the sleep times, it will prevent large sleeps.

Although, if the game falls behind, I can see it taking longer to catch back up if the conditions allow it. I think that if you make the number of samples for the average into a smaller number, it should mitigate that. Can you try something like 5 samples and see if your issue is still fixed?

loganmc10:
@krnlyng I tried your patch. It didn't work at all, there was no speed limit. Are you sure that you have vsync disabled on your computer? It's possible that that patch just unlimited the speed limit for you, and your computer kept it in line with vsync, which gave you better performance

krnlyng:
@loganmc10 hmm thats interesting. can you make sure the speed limiter is active by running mupen64plus with --verbose? because the message:
"Waiting %ims" is printed for me in this case, is it for you? (i even changed the message slightly just now to make sure i am running the right binary).

i was using glide64mk2 and both the setting VerticalSync in Video-General and the vsync Setting in glide64mk2 are set to false.

are you sure that the speed limiter is enabled (l_MainSpeedLimit variable) and that l_SpeedFactor is 100?

loganmc10:
@krnlyng yes, the speed limiter was working, I tested the current master to be sure.

To be honest this all seems like over-thinking it to me. For an NTSC game, a VI should take 16.66666 ms. All you need to do is calculate how long it took you to process a frame, subtract that from 16.66666ms, and that is your sleep time. This is the implementation I came up with, it seems to work perfectly:

```
#define FRAME_NUM 5

static void apply_speed_limiter(void)
{
    const uint64_t SEC_TO_NSEC = 1000000000;
    static uint64_t oldNS = 0;
    static int frame_index, i = 0;
    static uint64_t averages[FRAME_NUM];
    struct timespec monotime;

    clock_gettime(CLOCK_MONOTONIC, &monotime);
    uint64_t currentNS = monotime.tv_sec * SEC_TO_NSEC + monotime.tv_nsec;

    averages[frame_index++] = currentNS - oldNS;
    if (frame_index == FRAME_NUM)
        frame_index = 0;
    uint64_t frame_average = 0;
    for (i = 0; i < FRAME_NUM; i++)
        frame_average += averages[i];
    frame_average /= FRAME_NUM;

    const uint64_t VINanoSec = (uint64_t)((1000.0 / g_dev.vi.expected_refresh_rate) * 1000000.0);
    if (VINanoSec > frame_average) {
        uint64_t diff = VINanoSec - frame_average;
        monotime.tv_sec = diff / SEC_TO_NSEC;
        monotime.tv_nsec = diff - (monotime.tv_sec * SEC_TO_NSEC);
        nanosleep(&monotime, NULL);
    }
    clock_gettime(CLOCK_MONOTONIC, &monotime);
    oldNS = monotime.tv_sec * SEC_TO_NSEC + monotime.tv_nsec;
}
```
This implementations averages out the frame time over 5 frames to smooth out any jittering, thoughts? (Right now it doesn't take into account l_MainSpeedLimit or l_SpeedFactor, it also uses nanosleep)

krnlyng:
@loganmc10 isnt that what @wareya was trying to prevent? because that way we add the error due to clock resolution in every frame don't we?

loganmc10:
The current implementation uses SDL_GetTicks() every frame, wouldn't that lead to the same error? I'm not sure what the difference would be, at least in this implementation it is using nanoseconds for the clock time and sleep time, instead of milliseconds in the current implementation.

I'm not sure how it could "leak over time", since the average is only calculated over the last 5 frames, so the only "time" that matters is the last 5 frames, it wouldn't get worse as time goes on.

wareya:
The *absolute time* of the current frame will drift from its intended value if you don't either use integer value math absolutely everywhere or calculate the desired time for the next frame based on an absolute reference point that doesn't change.

fzurita:
Ok, I updated my pull request. It uses @krnlyng concept to average the sleep times to reduce jitter. I'm using 5 samples currently. @krnlyng could you test the updated pull request to see how it works for you? If needed, can you increase the number of samples if it doesn't work with 5?

fzurita:
Whoops, build error in gcc... it compiled fine in clang.

loganmc10:
@fzurita I think you have a winner here. I tested Rogue Squadron for 2000 VI/s on very high/slow settings. With the current master it sleeps 7072 ms on my PC. With this PR it sleeps 5566 ms, so that makes me happy.

As another note (maybe something for another PR, I don't want to make this too complicated). There are a lot of modern machines that are capable of sleeping and keeping time in the nanosecond range. I think on machines that support nanosleep it would be better to use that rather than SDL_Delay/SDL_GetTicks

krnlyng:
@wareya yes i totally agree but this does have one problem: if the hardware is too slow for the game, e.g. there is almost never time to sleep and additionally there are frames which need too much time then you most likely can never catch up to the correct time. maybe we need the limiter to be aware of both cases, as in the case where we can actually catch up, or don't need to, because the hardware is fast enough and the case where the hardware is too slow and we cannot catch up?
i have done a quick test of the new limiter and it seems to work better, but i will do direct comparisions once i am at home

src/main/main.c
```diff
@@ -802,16 +805,41 @@ static void apply_speed_limiter(void)
...
+    for(index = 0; index < elementsForAverage; index++)
+    {
+        sum += sleepTimes[index];
+    }
+
+    double averageSleep = sum/elementsForAverage;
+
+    int sleepMs = (int)averageSleep;
+
+    if(sleepMs > 0 && sleepTime < maxSleepNeeded*SpeedFactorMultiple && l_MainSpeedLimit)
```
loganmc10: This should probably be
```
if (sleepMs > 0 && sleepMs < maxSleepNeeded*SpeedFactorMultiple && l_MainSpeedLimit)
```

fzurita: Sure

wareya:
@krnlyng The easy way to deal with that is to give the simulation run a little behind the framerate limiter without resetting the "initial" frametime. That lets the simulation catch up after small hitches that only last for a single frame.

src/main/main.c
fzurita: Ok, done.

fzurita:
@wareya That was the intention of this pull request.

loganmc10:
> You can also turn off the framerate limiter automatically if the general performance of the past <span of time> is worse than the target framerate, framerate limiter excluded, and turn it back on in the opposite case.

That will cause a burst of speed if you go from running at 20FPS to being able to run at 80FPS (but you should be running at 60FPS)

wareya:
>That was the intention of this pull request.

It already did that, just with a 50ms leeway instead of a 5 second leeway. I was responding to the way logan interpreted the new code. Smoothing out the sleep timer is a different thing that makes up for the low resolution of SDL's clock.

>That will cause a burst of speed if you go from running at 20FPS to being able to run at 80FPS (but you should be running at 60FPS)

It only causes if a burst of speed if the logic that turns it back on is way too conservative.

krnlyng:
@loganmc10 i am not aware of the full details but if you don't have a RTOS you will have a hard time sleeping accurately, there are a lot of posts around the web "complaining" that the sleep functions in many operating systems do have latency and random results, due to other processes running on the system with higher priority and so on

loganmc10:
> Edit: Shifting StartFPSTime forwards in time by minSleepNeeded when the desired frametime is more than minSleepNeeded in the past doesn't keep it from being reset, it just pretends the game sort of caught up to the game starting next frame.

It does prevent it from being reset, before, if ```sleepTime < minSleepNeeded``` that would cause a reset, now it doesn't

wareya:
That's a different part of the code.

loganmc10:
Using my "Rogue Squadron" test the current master resets 282 times in 2000 VI's. With this PR, it never resets, so you're wrong about that.

wareya:
I didn't say the PR didn't fix resets. the PR changes the reset threshold from 50 milliseconds overdue to 5000 milliseconds overdue. That would certainly make the framerate limiter reset less.

I said a specific *other* line doesn't fix resets. Rather, that's not what that specific line's behavior is.

loganmc10:
commenting out
```
if(sleepTime < minSleepNeeded)
{
    StartFPSTime -= minSleepNeeded;
}
```
Makes it reset 4 times. Furthermore, leaving those lines in, and changing minSleepNeededForReset from -5000 to -100 makes it reset 79 times (instead of 282 in the current master), so those lines obviously make a difference in how many times it resets.

loganmc10:
> It's the same thing as resetting it, except it's pretending it's not doing so, and it undershoots instead of hard resetting to the current time exactly.

Obviously not, since I measured how many ms it sleeps for in both scenarios, and over 2000 VI's it's a difference of roughly 2000ms, so it's not just a wash.

wareya:
>those lines obviously make a difference in how many times it resets.

*Line*. Not *lines*. I was commenting on *a single specific line*. Yes, minSleepNeededForReset makes it reset way less. I never said anything contrary to that. That doesn't mean minSleepNeeded doesn't currently (in this PR) do what I said it does, which is reset the framerate limiter without recording a reset.

loganmc10:
> Communicate. Read the words I'm writing, not what you think I'm writing.

You talk a lot, but you've never contributed any code to this project. You talk like you're an expert, but what would really be helpful is if you measured what you're talking about, or if you think you can do a better job of it, then write the code and we can test it out and see how it works. You're not being helpful.

loganmc10:
Again, your words aren't helpful, post some code or measurements

loganmc10:
I'm not aware of a compiler that would accept that as code.

Anyway this has obviously gotten away from me, I'm going to "recuse myself" from this issue/PR, but my testing does show that it reduces the amount of sleep taken when the game isn't running at full speed, I think it should be merged.

fzurita:
@wareya just to clarify some of the behavior without getting too involved in this argument. A reset causes a sleep to happen. Shifting the start time doesn't cause any additional sleeps, so it's different in practice. I guess you could say, reset type A and reset type B.

krnlyng:
@fzurita
yes the speed limiter works better but there is still a noticeable amount of frames being dropped. commenting out:
`if(sleepTime < minSleepNeeded)
{
    StartFPSTime -= minSleepNeeded;
}`
avoids this problem, but thats not a solution :)

fzurita:
Hmm, interesting. The piece of code you pointed out was the main piece of code intended to fix the problem. If you comment that code out I would expect more sleeps happening. What do you mean by frame drop? No frames should be dropped. Do you mean unintended sleeps?

krnlyng:
@fzurita the version of gles2n64 i am using has auto frame skip built in.
anyways i did further testing and have made the following experiences:
in my case the speed limiter almost always wants to sleep less than 10 ms. this is bad because if we expect a granularity of ~10ms from the operating system (due to higher priority tasks, task switching and the os timer...) then the sleep would always wait ~10ms even though we only want it to sleep for example for 6ms. this might also be what causes the jitter, since i've tried the following:

https://bpaste.net/show/cdac45e433d6

if we have more than the granularity of the OS timer to sleep we sleep until just before we've reached the granularity, then we do a busy loop for the time thats left to sleep. (I know it's **not** ideal.) This way we don't give up some precious milliseconds that easily.

note that i've commented out the averaging again and did not notice drawbacks from it.

fzurita:
Busy loop is not very ideal. It's a killer on battery life for mobile devices. If we intend to sleep for 6ms but end up sleeping for 10ms, we will end up sleeping for 4ms less the following time around. Unfortunately, like you said, that is introducing jitter. A lot of this is going to depend on how the kernel for a device is configured, one would hope that a kernel would be configured to have granularity better than 1ms. Unfortunately, there are many linux devices out there all with different settings, which makes the problem hard.

Maybe as a middle ground solution, a setting could be added to the core to make it use a busy loop instead of sleep. This would leave it up to the front end or user to choose whether they want better performance or battery savings.

If you compare this updated pull request to the implementation prior to https://github.com/mupen64plus/mupen64plus-core/commit/0814daaad842868d8283f686b80245186c425280 how does it compare in your point of view?

loganmc10:
commenting out ```if(sleepTime < minSleepNeeded) { StartFPSTime -= minSleepNeeded; }``` is an acceptable solution, it does actually seem to work well for me.

You can try this:
https://pastebin.com/gGqNDe0k

That is basically just adding in the averaging bit. Because of the averaging, after a reset, the new sleepTime is just 1 input of 5, so it doesn't necessarily trigger a sleep.

krnlyng:
forget the busy loop, it was probably an unfair test which i did with the charger disconnected in one case which causes more aggressive power saving measures on my device. the current patch with the averaging does work very similar to the one with the busy loop and very similar to the old limiter from 0814daa so i would be fine with your current patch.

fzurita:
@loganmc10 so commenting out ```if(sleepTime < minSleepNeeded) { StartFPSTime -= minSleepNeeded; }``` Gives you the same result as the pull request? If so I might as well take that part out.

loganmc10:
Let me compare the 2 just to be sure and I'll let you know

loganmc10:
Yes, I tested this PR as it is now, and also this PR with ```if(sleepTime < minSleepNeeded) { StartFPSTime -= minSleepNeeded; }``` removed. I also removed ```minSleepNeededForReset``` and ```sleepMs < maxSleepNeeded*SpeedFactorMultiple```. So basically the averaging is the only thing new.

Same Rogue Squadron test, one slept for 2311ms, one slept for 2358ms, close enough for me (I'm testing this on a different, slower computer than before, which is why it is sleeping less than my earlier tests). I will also say that it does a really good job of not sleeping when things are slow, or even when it's speeding up, it only sleeps when it is running 60 VI/s, very nice.

The diff is here if it's helpful:
```
diff --git a/src/main/main.c b/src/main/main.c
index 527b78f..81e827c 100644
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -760,6 +760,7 @@ void new_frame(void)
     }
 }
 
+#define SAMPLE_COUNT 5
 static void apply_speed_limiter(void)
 {
     static unsigned long totalVIs = 0;
@@ -768,6 +769,8 @@ static void apply_speed_limiter(void)
     static unsigned int StartFPSTime = 0;
     static const double defaultSpeedFactor = 100.0;
     unsigned int CurrentFPSTime = SDL_GetTicks();
+    static double sleepTimes[SAMPLE_COUNT];
+    static unsigned int sleepTimesIndex = 0;
 
     // calculate frame duration based upon ROM setting (50/60hz) and mupen64plus speed adjustment
     const double VILimitMilliseconds = 1000.0 / g_dev.vi.expected_refresh_rate;
@@ -805,13 +808,30 @@ static void apply_speed_limiter(void)
     {
        resetOnce = 0;
     }
-    else if(sleepTime > 0.0 && l_MainSpeedLimit)
-    {
-       DebugMessage(M64MSG_VERBOSE, "    apply_speed_limiter(): Waiting %ims", (int) sleepTime);
 
-       SDL_Delay((int) sleepTime);
+    sleepTimes[sleepTimesIndex%SAMPLE_COUNT] = sleepTime;
+    sleepTimesIndex++;
+
+    int elementsForAverage = sleepTimesIndex > SAMPLE_COUNT ? SAMPLE_COUNT : sleepTimesIndex;
+
+    // compute the average sleepTime
+    double sum = 0;
+    unsigned int index;
+    for(index = 0; index < elementsForAverage; index++)
+    {
+        sum += sleepTimes[index];
     }
 
+    double averageSleep = sum/elementsForAverage;
+
+    int sleepMs = (int)averageSleep;
+
+    if(sleepMs > 0 && l_MainSpeedLimit)
+    {
+       DebugMessage(M64MSG_VERBOSE, "    apply_speed_limiter(): Waiting %ims", sleepMs);
+
+       SDL_Delay(sleepMs);
+    }
     timed_section_end(TIMED_SECTION_IDLE);
 }
```

fzurita:
Ok, I'll update my pull request so that it only does the averaging.

fzurita:
@loganmc10 and @krnlyng Please give the pull request one more try to make sure everything is in order.

loganmc10:
hm no there is still something wrong. I was testing how it operated when things were slow, I didn't test much how things were when it ran normally. It's running too fast, the audio is falling behind. It gets worse the larger SAMPLE_COUNT gets, and better the smaller I make that number.

fzurita:
So if you set sample_count to 1, does it go back to normal?

loganmc10:
I believe the problem is that ```sleepTime``` can be negative, so you are actually adding negative values into the average. If the number is less than 0, you should add 0 to the average. I made this change, and it fixed the timing with the audio, but the bad news is that it also re-introduced the micro-stuttering into Rogue Squadron when things are slow.

fzurita:
Looks like I need to re-add some of the logic I had before.

fzurita:
I'll do it tomorrow, it's getting late for me tonight. Feel free to experiment with adding it back.

loganmc10:
Yeah I was just gonna say, re-adding ```if(sleepTime < minSleepNeeded) { StartFPSTime -= minSleepNeeded; } ``` and -5000 for the reset time reduced (but didn't eliminate) the amount of time it spent sleeping when Rogue Squadron was slow, I'll take proper measurements tomorrow

krnlyng:
i think removing negative values from the sleep is also not the correct solution :/ because that way you increase the overall time it spends sleeping. if you add the negative sleeps to the average you balance it out

krnlyng:
I've done some measurements:
Setup: Ocarina of Time with glide64mk2 running for 60s, 6 runs each
fzuritas current version: 1996ms time spent sleeping on average
my busy wait: 2151ms time spent sleeping on average with 2130ms spent in the busy loop itself on average (almost everything so very bad for battery life)
limiter from before 0814daa: 1892ms time spent sleeping on average

and i think this has to be interpreted in the following way:
If we assume that the speed limiter works properly, e.g. is not waiting too much that means more sleep is actually better since the emulator can give up more time when it is running better.
So in my opinion we can keep fzuritas current version and optionally keep an optional option for the busy loop method. But we can also forget about the busy loop since the effect on gameplay is not very noticeable.

fzurita:
@loganmc10 if we don't allow some negative sleep times, we will have drift over time. Negative sleep times are normal even when running full speed. You say that we are running too fast because audio lag is increasing over time because audio is not playing fast enough to keep up? How much of a difference are you seeing?

loganmc10:
It's fairly significant and causes the audio to crackle. I'll do some more testing today

fzurita:
Hmm, they sound like the opposite. If you hear cracking, that means that audio buffers are not being kept full. This can only happen if the core is not running fast enough to keep up with the audio playback rate. So we are sleeping for too long.

fzurita:
Ok, gotcha, different source of crackling than what I thought.

fzurita:
I'm looking at the source code of your version of audio SDL, why is there no resampling?

loganmc10:
It just plays it at the native sample rate of the game, trying to keep it simple, I'm not actually sure what the purpose or resampling is if the sound card can use the original sample rate

fzurita:
Ok, the old SDL didn't seem to make that assumption. Also in my experience with Android, even with Mario 64, I couldn't get the plugin to play at the audio frequency of the game. Maybe it's different with your PC.

loganmc10:
Yeah I've used it on Windows and Linux with no issue, but it's probably not the most portable thing. If someone ever complains I'll probably add the resampling back in, I might be the only person using it right now lol

fzurita:
Back to the topic, can you determine what percentage of samples you are dropping on your version of audio SDL? It would tell us how much faster than normal we are running.

fzurita:
Another experiment, probably a better one than the previous I suggested. Could you change this ```SecondaryBufferSize * SDL_SAMPLE_BYTES * 2``` to something like 10 instead of 2? And see if it recovers before hitting 10. I'm wondering if it had increased jitter when playing at normal speed compared to the old limiter.

loganmc10:
Sure I'll test both, first test:

Using the current master, 2000 calls to AiLenChanged, it never skips. All 2000 sample "chunks" played.

With this PR, it skipped 51 "chunks" and played 1949.

Changing ```* 2``` to ```* 4``` solves it completely, it doesn't miss any (I would have expected it to miss around 25 if it was permanently behind)

fzurita:
That confirms that this is increasing jitter when running at normal speeds, not good.

loganmc10:
If I try the regular audio SDL plugin, I don't hear any issues though. It sounds good, and it doesn't print out any messages about an overflow (https://github.com/mupen64plus/mupen64plus-audio-sdl/blob/master/src/main.c#L447) which I would expect if it's getting behind.

Maybe my audio plugin needs a greater tolerance. Maybe testing it on an Android device with time stretching disabled would be a good test?

fzurita:
Yeah, I did do that. It sounds fine. But the additional jitter is very visible.

loganmc10:
Maybe 5 is just too big of a sample size? If I change it to 3, then I get all 2000 audio sample chunks at ```* 2``` tolerance, and it doesn't sleep during my Rogue Squadron tests when it's slow

fzurita:
I'll give 2 a shot

fzurita:
I see no issues with the value set to three. So that's probably good enough.

fzurita:
Ok, pull request updated. Hopefully for the last time ;). @bsmiles32 Do you think this could be merged?

loganmc10:
Just to confirm, using my "Rogue Squadron test" of 2000 VI's. The current master slept for 3768 ms, with the PR as it is right now, it slept for 2618 ms. So a big improvement. The audio also sounds good and it's keeping in sync. I think it's ready to be merged.

krnlyng:
@fzurita not sure if this is a problem but i've printed the FPS count on my device, and it seems that if sample count is small, e.g. 3 the game will sometimes show more than 60 frames per second. if it is increased to 5 again that doesn't happen as often, because it is averaged out.
I've only noticed it because i was dynamically adjusting the speed of the playback to the games speed :D and audio was sometimes too fast. (I did this to avoid stuttering audio/crackling but it doesn't sound too good if frame rates really drop :D)

fzurita:
It's expected for the current algorithm to sometimes go slightly too fast to catch up if it fell behind earlier. It does this because it's trying to keep the game time aligned to the real time.

loganmc10:
My tests showed that it kept the audio pretty tightly aligned with the video, it wasn't dropping any of the audio samples with the way the PR is now

krnlyng:
@fzurita yes i was just thinking whether it really makes sense to go fast for one second, then slow for one second, then fast again... but probably thats fine at least visually it looks alright at ~60fps (with the occasional framerate jump up and down).

loganmc10:
Just a reminder @bsmiles32 @richard42 that this has been tested fairly thoroughly and should be ok to be merged

bsmiles32:
Thank you guys for working on this. Merging this now.

PR #253, opened 2017-03-10
Restore old behavior when allocating memory for RDRAM

fzurita:
As seen in https://github.com/mupen64plus/mupen64plus-core/blob/b67105b031d51607305376175281078a45bff0c2/src/main/main.h#L48

Correct me if I'm wrong, but it seems like we always used to allocated the full RDRAM size even though the expansion pak was disabled. After the refactor we are allocating just enough.

This is causing a crash when loading and saving save states because we try to copy the full expansion pak size to and from the save state. This pull request fixes the crash.

This is the commit that broke it:
https://github.com/mupen64plus/mupen64plus-core/commit/99d489dbcc4396ff29a1461aff7a01db1ee2940a

### DIFF
--- a/src/api/frontend.c
+++ b/src/api/frontend.c
@@ -97,7 +97,7 @@ EXPORT m64p_error CALL CoreStartup(int APIVersion, const char *ConfigPath, const
     /* allocate memory for rdram */
     disable_extra_mem = ConfigGetParamInt(g_CoreConfig, "DisableExtraMem");
     g_rdram_size = (disable_extra_mem == 0) ? 0x800000 : 0x400000;
-    g_rdram = malloc(g_rdram_size);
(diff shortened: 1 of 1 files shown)


fzurita:
How would you determine how big the expansion pak is in the save state when loading it? I guess we would need a new parameter stored in the file.

bsmiles32:
That's exactly the problem, the RDRAM size is not saved in the savestate, so it's not possible to know.
Adding that to the savestate would break backward compatibility, so I'd prefer to delay that change at a later time when we have enough reason to break backward compatibility. Overallocating 8M of RDRAM is a cheap fix and far easier than to change the savestate format.

fzurita:
I just thought about this, a simple fix for the save state format would be to not change the save state format at all. Instead, if the expansion pak is disabled, just load the first 4MB of RDRAM.

PR #255, opened 2017-03-24
new_dynarec: Fix crash during shutdown

Gillou68310:


loganmc10:
I don't really understand assembly at all, but is that ```stack_space``` actually used anywhere?

Gillou68310:
Yep, this space is used by the recompiled code for pushing/popping caller saved registers before/after C function call.

```
// Save registers before function call
static void save_regs(u_int reglist)
{
  reglist&=0x100f; // only save the caller-save registers, r0-r3, r12
  if(!reglist) return;
  assem_debug("stmia fp,{");
  if(reglist&1) assem_debug("r0, ");
  if(reglist&2) assem_debug("r1, ");
  if(reglist&4) assem_debug("r2, ");
  if(reglist&8) assem_debug("r3, ");
  if(reglist&0x1000) assem_debug("r12");
  assem_debug("}");
  output_w32(0xe88b0000|reglist);
}
// Restore registers after function call
static void restore_regs(u_int reglist)
{
  reglist&=0x100f; // only restore the caller-save registers, r0-r3, r12
  if(!reglist) return;
  assem_debug("ldmia fp,{");
  if(reglist&1) assem_debug("r0, ");
  if(reglist&2) assem_debug("r1, ");
  if(reglist&4) assem_debug("r2, ");
  if(reglist&8) assem_debug("r3, ");
  if(reglist&0x1000) assem_debug("r12");
  assem_debug("}");
  output_w32(0xe89b0000|reglist);
}
```

For example if the recompiler needs to save r0 before a function call the following code will be generated:
```stmia fp, {r0}```

This will store ```r0``` at the address in ```fp``` or ```fp``` points to the address of ```dynarec_local``` (which is the same address as ```stack_space```).

I added the ```stack_space``` label to help understanding, I guess it didn't help that much :-)

**Comment by loganmc10** [comment]
This appears to have fixed the problem:

```
Core: Starting R4300 emulator: Dynamic Recompiler
Core: Init new dynarec
Core: ARM CPU Features:, Half, Thumb, FastMult, VFP, ESDP, NEON, VFPv3, TLS, VFPv4, IDIVa, IDIVt
address before 0x715ba5b8
^CCore Status: Stopping emulation.
address after 0x715ba5b8
Core: R4300 emulator finished.
Core Status: Rom closed.
```
I would merge this ASAP since it represents a fairly serious bug in ARM, especially on Android it seems

bsmiles32:
Merged as per @loganmc10 request.

PR #257, opened 2017-04-07
Get World Driver Championship booting

loganmc10:
This is not something that should be merged, but I figured since there was code involved a PR would be an easier place to look at it than an Issue. Plus it's awesome.

To boot WDC, you need to use rsp-cxd4 and enable "SupportCPUSemaphoreLock". That, plus this change, should allow WDC to boot. I assume it will also fix Stunt Racer 64 and similar games.

The graphics are messed up (polygons everywhere). I know nothing about how the N64 works or interrupts or anything. This is just the result of me messing around for the last 4-5 days based on things I had read from HatCat/cxd4, angrylion, etc..

I am hoping that this will be a good hint for people much smarter than me @richard42 @bsmiles32 @Gillou68310 @fzurita @cxd4 and that someone will be able to take it from here and get these games working properly.

### DIFF
--- a/src/device/rsp/rsp_core.c
+++ b/src/device/rsp/rsp_core.c
@@ -40,7 +40,7 @@ static void dma_sp_write(struct rsp_core* sp)
     unsigned int length = ((l & 0xfff) | 7) + 1;
     unsigned int count = ((l >> 12) & 0xff) + 1;
     unsigned int skip = ((l >> 20) & 0xfff);
- 
+
     unsigned int memaddr = sp->regs[SP_MEM_ADDR_REG] & 0xfff;
     unsigned int dramaddr = sp->regs[SP_DRAM_ADDR_REG] & 0xffffff;
 
@@ -143,8 +143,8 @@ static void update_sp_status(struct rsp_core* sp, uint32_t w)
     if (w & 0x800000) sp->regs[SP_STATUS_REG] &= ~SP_STATUS_SIG7;
     if (w & 0x1000000) sp->regs[SP_STATUS_REG] |= SP_STATUS_SIG7;
 
-    //if (get_event(&sp->r4300->cp0.q, SP_INT)) return;
-    if (!(w & 0x1) && !(w & 0x4))
+    if (sp->rsp_task_locked && (get_event(&sp->r4300->cp0.q, SP_INT))) return;
+    if (!(w & 0x1) && !(w & 0x4) && !sp->rsp_task_locked)
         return;
 
     if (!(sp->regs[SP_STATUS_REG] & (SP_STATUS_HALT | SP_STATUS_BROKE)))
@@ -167,6 +167,7 @@ void poweron_rsp(struct rsp_core* sp)
     memset(sp->regs, 0, SP_REGS_COUNT*sizeof(uint32_t));
     memset(sp->regs2, 0, SP_REGS2_COUNT*sizeof(uint32_t));
 
+    sp->rsp_task_locked = 0;
     sp->regs[SP_STATUS_REG] = 1;
 }
 
@@ -284,6 +285,12 @@ void do_SP_Task(struct rsp_core* sp)
         new_frame();
 
         cp0_update_count();
+        sp->rsp_task_locked = 0;
+        if ((sp->regs[SP_STATUS_REG] & (SP_STATUS_HALT | SP_STATUS_BROKE)) == 0)
+        {
+            sp->rsp_task_locked = 1;
+            sp->r4300->mi.regs[MI_INTR_REG] |= MI_INTR_SP;
+        }
         if (sp->r4300->mi.regs[MI_INTR_REG] & MI_INTR_SP) {
             add_interrupt_event(&sp->r4300->cp0, SP_INT, 1000);
         }
@@ -305,6 +312,12 @@ void do_SP_Task(struct rsp_core* sp)
         sp->regs2[SP_PC_REG] |= save_pc;
 
         cp0_update_count();
+        sp->rsp_task_locked = 0;
+        if ((sp->regs[SP_STATUS_REG] & (SP_STATUS_HALT | SP_STATUS_BROKE)) == 0)
+        {
+            sp->rsp_task_locked = 1;
+            sp->r4300->mi.regs[MI_INTR_REG] |= MI_INTR_SP;
+        }
         if (sp->r4300->mi.regs[MI_INTR_REG] & MI_INTR_SP) {
             add_interrupt_event(&sp->r4300->cp0, SP_INT, 4000/*500*/);
         }
@@ -318,6 +331,12 @@ void do_SP_Task(struct rsp_core* sp)
         sp->regs2[SP_PC_REG] |= save_pc;
 
         cp0_update_count();
+        sp->rsp_task_locked = 0;
+        if ((sp->regs[SP_STATUS_REG] & (SP_STATUS_HALT | SP_STATUS_BROKE)) == 0)
+        {
+            sp->rsp_task_locked = 1;
+            sp->r4300->mi.regs[MI_INTR_REG] |= MI_INTR_SP;
+        }
         if (sp->r4300->mi.regs[MI_INTR_REG] & MI_INTR_SP) {
             add_interrupt_event(&sp->r4300->cp0, SP_INT, 0/*100*/);
         }
@@ -328,9 +347,12 @@ void do_SP_Task(struct rsp_core* sp)
 
 void rsp_interrupt_event(struct rsp_core* sp)
 {
-    /* XXX: assume task has fully completed */
-    sp->regs[SP_STATUS_REG] |=
-        SP_STATUS_TASKDONE | SP_STATUS_BROKE | SP_STATUS_HALT;
+    if(!sp->rsp_task_locked)
+    {
+        /* XXX: assume task has fully completed */
+        sp->regs[SP_STATUS_REG] |=
+            SP_STATUS_TASKDONE | SP_STATUS_BROKE | SP_STATUS_HALT;
+    }
 
     if ((sp->regs[SP_STATUS_REG] & SP_STATUS_INTR_BREAK) != 0)
     {
--- a/src/device/rsp/rsp_core.h
+++ b/src/device/rsp/rsp_core.h
@@ -79,6 +79,7 @@ struct rsp_core
     uint32_t mem[SP_MEM_SIZE/4];
     uint32_t regs[SP_REGS_COUNT];
     uint32_t regs2[SP_REGS2_COUNT];
+    uint32_t rsp_task_locked;
 
     struct r4300_core* r4300;
     struct rdp_core* dp;


fzurita:
Cool, I'm sure this will be a very big hint.

loganmc10:
For the uninformed (this is based on my understand of what I read and could be wrong):

WDC and similar games get stuck in doRspCycles, enabling "SupportCPUSemaphoreLock" allows it to exit the loop with SP_STATUS_HALT set to 0, instead of the usual 1. This is a hint that it has exited this problematic loop.

I'm not sure what triggers it to set SP_STATUS_HALT to 1 eventually. Right now, it just keeps setting SP_STATUS_HALT to 0 after the loop, that is why I have "repeat_count", I just let it do that 100 times, then I force it to quit basically.

I assume that there is a more proper way to deal with that, since 100 was just a random number I chose.

loganmc10:
Also for reference, I based this work mostly off of these comments:

https://code.google.com/archive/p/mupen64plus/issues/614

loganmc10:
I believe exiting the loop properly (without relying on a counter) involves checking ```sp->r4300->mi.regs[MI_INTR_REG] & MI_INTR_SP```, like if it's 1 you can stop repeating, 0 if you need to keep repeating, something like that.

But I haven't figured out yet how to properly deal with that variable.

loganmc10:
I've updated the PR with a version that doesn't require an arbitrary counter. The results still look just as bad, but now it doesn't just randomly count to 100, it exits naturally.

src/device/rsp/rsp_core.c
```diff
@@ -321,16 +326,26 @@ void do_SP_Task(struct rsp_core* sp)
...
         }
-        sp->r4300->mi.regs[MI_INTR_REG] &= ~MI_INTR_SP;
         sp->regs[SP_STATUS_REG] &= ~SP_STATUS_TASKDONE;
     }
+
+    if ((sp->regs[SP_STATUS_REG] & SP_STATUS_HALT) == 0 && sp->r4300->mi.regs[MI_INTR_REG] == 0) {
+        repeat = 1;
+        add_interrupt_event(&sp->r4300->cp0, SP_INT, 0x200);
+    } else {
+        add_interrupt_event(&sp->r4300->cp0, SP_INT, 0x200);
```
loganmc10: I don't think this is right, it looks like Project64 only add an interrupt when it has to repeat.

Gillou68310:
I took a quick look at this and it seems that the RSP is locked in an endless loop waiting for a signal from the CPU (SP_STATUS_SIGNAL0 / SP_STATUS_YIELD). Here's the offending code:

```
04001138: mfc0   r2, SP_STATUS
0400113C: andi   r2, r2, $0080
04001140: bne    r2, 0, $04001124
04001144: nop
04001148: j      $04001138
0400114C: nop
```

@cxd4 uses the following hack to break the endless loop and give back control to the CPU.

```
#ifdef WAIT_FOR_CPU_HOST
    if (rd == 0x4)
    {
        ++MFC0_count[rt];
        GET_RCP_REG(SP_STATUS_REG) |= (MFC0_count[rt] >= MF_SP_STATUS_TIMEOUT);
        CPU_running = ~GET_RCP_REG(SP_STATUS_REG) & 1;
    }
#endif
```
The "SupportCPUSemaphoreLock" option should not be necessary for this game.

From what I understand the core should detect that the RSP task ended with the SP_STATUS_HALT/BROKE flags equal 0 and schedule a new RSP interrupt allowing the CPU to continue execution until the new interrupt is triggered.

Here's my implementation, thanks @loganmc10 for the hint ;-)
https://github.com/Gillou68310/mupen64plus-core/commit/b75be1febcc25daea4b286edd92457a3b95e7776

I tested WDC and Gauntlet Legend, both game seem to work fine.

loganmc10:
Awesome, I'll test it out tomorrow, I thought that SupportCPUSemaphoreLock was required for the RSP to exit with halt/broke set to 0, but I guess it isn't required

Gillou68310:
Gauntlet legend has issue with gliden64 but works fine with angrylion.

fzurita:
What about Indiana Jones?

Gillou68310:
The flickering textures are still present if that was the question ;-)

Gillou68310:
Stunt Racer 64 works too but has a lot of stuttering

fzurita:
Last time I tried Indiana Jones didn't boot. Good that it is I guess with that.

What are the chances of getting this incorporated? It doesn't look like to bad of a hack, at least no worse than Mupen64plus' current inaccuracies.

fzurita:
@Gillou68310  I did a few tests with your change.
* Indiana Jones still didn't boot for me.
* WDC does boot, but the core is crashing somewhere.
* GLideN64 LLE mode does seem to boot Gauntlet Legends, HLE mode still doesn't work. 
* Stunt racer, boots in LLE mode, but the core quickly crashes.

loganmc10:
Both WDC and Stunt Racer worked for me, no crashing. They are both very choppy though, and the CPU is only at around 25-30%, so unless the GPU is really maxed out (I'll try an Nvidia GPU tomorrow), there seems to be something wrong with it still.

I couldn't get Indiana Jones to boot, perhaps it is a different problem, maybe the same as All Star Baseball 2000/2001? They still don't work either.

> GLideN64 LLE mode does seem to boot Gauntlet Legends, HLE mode still doesn't work.

When you say "HLE mode", do you mean using rsp-hle? Or using cxd4 with HLE video? This "hack" will only work with rsp-cxd4, and I'm not sure if it would work with cxd4 + HLE video

fzurita:
By HLE mode I mean CXD4 with HLE video.

loganmc10:
Maybe it's something in the arm dynarec? I tested it on x86-64

Gillou68310:
Yes the new dynarec is broken with those games on x86 too, I need to investigate.

This hack only apply to cxd4 in LLE mode.

No problem booting Indiana Jones by my side, also it doesn't require this hack.

Gillou68310:
> hey are both very choppy though, and the CPU is only at around 25-30%, so unless the GPU is really maxed out (I'll try an Nvidia GPU tomorrow), there seems to be something wrong with it still.

I just realized that WDC is stuttering too on my home computer. I didn't have such issue on my office laptop.

fzurita:
@Gillou68310 what version of Indiana Jones are you using?

Gillou68310:
US version

fzurita:
Maybe I don't have a good dump of Indiana​ Jones. What is your md5 @Gillou68310 ? Mine is 70DE1EAB508596B6BBEFD168B5D07194

Gillou68310:
Same md5 here, what's your plugin config?

fzurita:
I'm using CXD4-LLE with Angrylion on Android.

Gillou68310:
New dynarec?

fzurita:
Yes

Gillou68310:
Try interpreter

fzurita:
I just did, it's no better.

fzurita:
Any special CXD4 config?

Gillou68310:
Weird

Gillou68310:
Nope

Gillou68310:
Well now that you ask it's possible that I was using rsp-z64 when I tested the game.

Gillou68310:
I don't have access to my computer right now so I cannot tell for sure

fzurita:
Hmm, maybe it's worthwhile to get that ported to Android.

Gillou68310:
rsp-cxd4 is faster

Gillou68310:
I'll check when I get home if indi is working with cxd4 and let you now

fzurita:
ok

Gillou68310:
Hum I remember checking if the semaphore lock option in cxd4 could help with the flickering textures, so I'm pretty sure it was working with it.

fzurita:
Strange, I just tried toggling all the options in CXD4. I couldn't get it to work.

Gillou68310:
Which version of the cxd4 plugin are you using in m64p-ae? Upstream one?

fzurita:
m64p-ae, which should be identical to https://github.com/mupen64plus/mupen64plus-rsp-cxd4

Gillou68310:
Could you try without the neon optimization?

fzurita:
I just did, no luck.

Gillou68310:
Do you have an x86 android device?

fzurita:
Yeah, I do, I'll give it a shot.

Gillou68310:
Don't use new dynarec I'm pretty sure it's broken with indi.

fzurita:
Yeah, I have been using the cached interpreter.

fzurita:
I tried in the Nexus Player (x86) and it's not booting either. Something must be different. Can you post your mupen64plus.cfg?

loganmc10:
I was using the dynarec during my test, I'll test the interpreter and cxd4 tonight

Gillou68310:
I'll share my config file tomorrow, one last test you can do in the meantime is to use gliden64 instead of angrylion.

@loganmc10 old dynarec should work fine only new dynarec is broken

loganmc10:
Here is the fix for Indiana Jones:

https://github.com/mupen64plus/mupen64plus-core/pull/263

fzurita:
@loganmc10 Thanks for that. That worked.

loganmc10:
So I just tried WDC on my laptop with an Nvidia GPU and I still get the same stuttering, GLideN64 reports 60 VI/S though, so it seems to think it's running at full speed, therefore I assume it's not the case of the computer not being able to keep up, there must be something wrong.

Gillou68310:
I just checked stunt racer on my office laptop and no stuttering at all!

Gillou68310:
What is happening? CPU low power mode?

loganmc10:
I don't know I'll have to look into it more. I tested Project64 using the same laptop and I don't experience the stuttering there. I used GLideN64 for both emulators.

fzurita:
Could it be CountPerOp again?

loganmc10:
I tried different values for those settings, no difference.

loganmc10:
So another hint: disabling the audio plugin fixes the stuttering for me. @Gillou68310 are you using audio-sdl? Do you have any special settings for the audio plugin?

loganmc10:
I notice that for an audio RSP event it does:
```
add_interrupt_event(&sp->r4300->cp0, SP_INT, 4000/*500*/);
```
And for GFX it does:
```
add_interrupt_event(&sp->r4300->cp0, SP_INT, 1000);
```
In Project64, if the RSP task has to repeat, it always uses ```0x200``` as the timer value. Perhaps it is waiting too long to repeat the RSP task? Again, I don't really know how the interrupt system works, so it's just a wild guess.

Gillou68310:
Nice catch I forgot that I was using the dummy audio plugin at work! Sorry about that. I'll see what I can do about the stuttering issue tomorrow.

loganmc10:
It seems like there may be something wrong with the speed limiter in Audio-SDL. Just commenting out ```SDL_Delay(WaitTime);``` seems to fix the stuttering for me.

wareya:
The problem is the length of the period of time the audio plugin waits for SDL to need more samples. It's basically a second framerate limiter, but it limits to a low framerate.

loganmc10:
> The problem is the length of the period of time the audio plugin waits for SDL to need more samples. It's basically a second framerate limiter, but it limits to a low framerate.

Well in this case I assume it is miscalculating where it "should be". I have the core speed limiter disabled, so the only speed limiter in use is Audio-SDL's. If I use the dummy audio plugin and the core's speed limiter, then it works properly.

wareya:
What are the SDL audio backend's current settings?

loganmc10:
Everything is the default.

loganmc10:
For what it's worth, enabling "Sync to audio" in Project64 breaks the timing in WDC, and it works in other games. Perhaps Audio-SDL should have an option to disable the speed limiter?

Is the core's speed limiter not enough to keep the audio in sync? It seems like in Project64, audio sync is disabled by default, and it can be enabled if desired, perhaps it should be the same for Audio-SDL?

loganmc10:
Besides all these issues, it seems that the changes to the core are working perfectly. I think it's worth submitting a PR for that, since it can be a lengthy process... Plus @Gillou68310 you have a habit of doing awesome work and then never sending a PR :P (I'm looking at you DK bone displacement and RE2 background microcode)

src/device/rsp/rsp_core.c
```diff
@@ -31,6 +31,8 @@
 #include "main/profile.h"
 #include "plugin/plugin.h"
 
+static int repeat_rsp_task = 0;
```
bsmiles32: No global state here please. Move that variable inside rsp_core structure.

src/device/rsp/rsp_core.c
```diff
@@ -263,6 +265,9 @@ int write_rsp_regs2(void* opaque, uint32_t address, uint32_t value, uint32_t mas
 void do_SP_Task(struct rsp_core* sp)
 {
     uint32_t save_pc = sp->regs2[SP_PC_REG] & ~0xfff;
+    uint32_t save_regs;
+    if (repeat_rsp_task)
+        sp->r4300->mi.regs[MI_INTR_REG] &= ~(MI_INTR_SP | MI_INTR_DP | MI_INTR_AI | MI_INTR_SI);
```
bsmiles32: Why clearing all rcp interrupts us needed ?

loganmc10: Sorry you can ignore this PR, the code we are talking about is here:

https://github.com/Gillou68310/mupen64plus-core/commit/b75be1febcc25daea4b286edd92457a3b95e7776

loganmc10:
Ok I updated this PR with @Gillou68310 implementation, except that I moved rsp_task_locked into the rsp_core struct.

I believe this should be ok to be merged, there are still issues to sort out with the audio speed limiter/synchronization for these games though.

bsmiles32:
Thanks for updating ! Looks good to me. Do you want me to merge this or wait for @richard42 to take a look at it ?

loganmc10:
Well I would wait for Gilles to chime in, but I've tested WDC with these changes (and the Audio-SDL speed limiter disabled) and it works the same as Project64 now.

Gillou68310:
Thanks @loganmc10 for creating the PR ;-) I think this is ready to be merged since we have the same behavior as project64.

WDC and Stunt racer are both generating audio samples with AI_LEN_REG == 16384. This is problematic with the default buffer size in audio-sdl. The stuttering issue could be fixed by multiplying PRIMARY_BUFFER_SIZE, PRIMARY_BUFFER_TARGET and SECONDARY_BUFFER_SIZE by 16 but the audio latency will be awful! Also it seems like the audio is not handled by an RSP task for those games because the core never calls ```doRspCycles``` with ```sp->mem[0xfc0/4] == 2```.

There's no stuttering in Gauntlet legend BTW ;-)

cxd4:
>Awesome, I'll test it out tomorrow, I thought that SupportCPUSemaphoreLock was required for the RSP to exit with halt/broke set to 0, but I guess it isn't required

Nope, just a separate trick that very few games use.  Gauntlets/Boss Games are not among them.

Of all the semaphores there is only one that the CPU uses as a mutex to manually cap the RSP slave time in conjunction with the master.  I only remember _Mario no Photopie_ requiring it.

olivieryuyu:
can we get WDC running in HLE by any chance? Without that, ucode reverse enginnering would not be possible.

PR #258, opened 2017-04-12
Reduce usage of g_dev

bsmiles32:
Still trying to reduce usage of global variable.
In this PR, I unified the way new_dynarec "hot" variables are accessed. This allowed me to remove usage of g_dev in many simple accessors (in r4300_core, cp0, cp1 mostly).
Then I did some more refactorings to reduce usage of g_dev inside r4300 module.


bsmiles32:
Added some more commits which reduce further direct g_dev variable usage.

loganmc10:
The PR that was just merged created a merge conflict here, FYI

bsmiles32:
rebased against master

bsmiles32:
Fixed some errors reported by @dankcushions in PR 269.

Gillou68310:
I just realized that movw and movt are not available on rpi1's armv6. If ARMv5_ONLY is defined you should use literal pool instead.

```
LDR Rd,=const
```

src/device/r4300/new_dynarec/arm/assem_arm.c
```diff
@@ -2697,12 +2720,12 @@ static void emit_extjump2(int addr, int target, int linker)
   //assert((target>=0x80000000&&target<0x80800000)||(target>0xA4000000&&target<0xA4001000));
 //DEBUG >
 #ifdef DEBUG_CYCLE_COUNT
-  emit_readword((int)&last_count,ECX);
+  emit_readword((u_int)&g_dev.r4300.new_dynarec_hot_state + fp_last_count,ECX);
```
Gillou68310: Why not using ```&g_dev.r4300.new_dynarec_hot_state.last_count``` instead?

src/device/r4300/new_dynarec/arm/linkage_arm.S
```diff
@@ -70,84 +69,46 @@
...
-    fake_pc                         = g_dev_r4300_pc                    + 4
-    ram_offset                      = fake_pc                           + 132
-    mini_ht                         = ram_offset                        + 4
-    restore_candidate               = mini_ht                           + 256
-    memory_map                      = restore_candidate                 + 512
-
-END_SECTION
+.global g_dev
+
+g_dev_r4300_new_dynarec_hot_state_dynarec_local = (g_dev + offsetof_struct_device_r4300_core + offsetof_struct_r4300_core_new_dynarec_hot_state + offsetof_struct_new_dynarec_hot_state_dynarec_local)
```
Gillou68310: ```offsetof_struct_device_r4300_core``` should be ```offsetof_struct_device_r4300``` that's probably explaining this error ```Error: invalid operands (*UND* and *UND* sections) for `+'```

Gillou68310: If so this commit https://github.com/mupen64plus/mupen64plus-core/pull/258/commits/2ef57320fb82a6905620fd80a68309b327531d19 should be reverted, except for the PIC part

Gillou68310:
You should also remove ```.cpu arm9tdmi``` so we can use armv7 instructions

src/device/r4300/new_dynarec/arm/assem_arm.c
bsmiles32: I guess that is just me mixing up asm and C struct member style accesses.

bsmiles32:
@Gillou68310 : just curious, any downside of removing the .cpu arm9tdmi statement ? If so, maybe trying to help the compiler align struct members properly (eg same as before) could avoid these changes ?

bsmiles32:
rebased against latest less_g_dev. Hope no regression happened :)

Gillou68310:
Removing the .cpu arm9tdmi should have no impact. Also I don't even understand why we are specifying all these attributes in the first place. We should probably let the compiler do it's job.

> If so, maybe trying to help the compiler align struct members properly (eg same as before) could avoid these changes ?

Alignment is not really the issue here, the problem is that the compiler added some padding in the hot state structure and there's not much we can do about it.

Also thanks for this https://github.com/mupen64plus/mupen64plus-core/pull/258/commits/8c3a6a786b914867ca4fb344145b67a4c30b0154 ;-)

bsmiles32:
@Gillou68310 : Looking more closely at the new offsets issue, I think it would be possible to use the old code with immediates specified as 8 bit value shifted by an even number. If we look at the faulty members
: restore_candidates (@ 0x4b8) and memory_map (@ 0x6b8), they can be made compatible again if they were @ 0x4c0 and @ 0x6c0, eg 16-byte aligned. We can ask for it with __attribute__ aligned((16)) (GCC extension) or use _Alignas (C11). Or try the packed structure approach and manually insert padding as needed. What do you think would be preferable in term of maintenance/performance ?

Gillou68310:
I think it's preferable to let the compiler decide the alignment and adapt the assembly code like we did. It will be easier to maintain and should have no performance impact.

Gillou68310:
extra_memory was previously aligned on 2^12 not sure if this was required for mmap?

src/device/r4300/new_dynarec/arm/linkage_arm.S
Gillou68310: Did you see this?

Gillou68310:
```
void *mmap(void *addr, size_t length, int prot, int flags, int fd, off_t offset);

If addr is NULL, then the kernel chooses the address at which to
create the mapping; this is the most portable method of creating a
new mapping.  If addr is not NULL, then the kernel takes it as a hint
about where to place the mapping; on Linux, the mapping will be
created at a nearby page boundary.  The address of the new mapping is
returned as the result of the call.
```

Looks like it is ;-)

Gillou68310:
```
On ARMv7, both mappings will (in general) work, and any user-allocatable virtual address can be specified provided that it is aligned at a 4KB page boundary.
```

Found at https://community.arm.com/processors/b/blog/posts/page-colouring-on-armv6-and-a-bit-on-armv7

src/device/r4300/new_dynarec/arm/linkage_arm.S
bsmiles32: Oups I didn't ! Will fix later :) That should indeed explain the remaining compiling error.

bsmiles32:
I thought 4k alignment was for better performance because it would put hot state in it's own page. If required, we will need to tell compiler to align this member at 4k with aligned attribute or _Alignas.

bsmiles32:
@loganmc10, @dankcushions Can you test if this PR works on your ARM device ? That would allow to reduce the range of suspicious commits to look for the breakage.

loganmc10:
I just tested on my Raspberry Pi 3. It didn't segfault. I'm testing this over SSH, I don't have my rpi plugged into my TV so I can't comment on what it looks like.

Tomorrow I'll test the commits in the other PR to track down where the segfault is starting.

bsmiles32:
Thanks. So it looks like this PR is ready ?

loganmc10:
Yeah I would say so. Tomorrow I can test this PR on my Android device to make sure the audio/video is still good, and I'll track down the bad commit in the other PR

bsmiles32:
Great ! Thanks for your dedication.

loganmc10:
Here are a couple of warnings I see when compiling for Android:
```
In file included from /home/travis/mupen64plus-ae/app/src/main/jni/./mupen64plus-core/src/device/r4300/new_dynarec/new_dynarec.c:50:
/home/travis/mupen64plus-ae/app/src/main/jni/./mupen64plus-core/src/device/r4300/new_dynarec/x86/assem_x86.h:1:9: warning: 'M64P_DEVICE_R4300_NEW_DYNAREC_X86_ASSEM_X86_H' is used as a header guard here, followed by #define of a different macro [-Wheader-guard]
#ifndef M64P_DEVICE_R4300_NEW_DYNAREC_X86_ASSEM_X86_H
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/travis/mupen64plus-ae/app/src/main/jni/./mupen64plus-core/src/device/r4300/new_dynarec/x86/assem_x86.h:2:9: note: 'M64P_DEVICE_R4300_NEW_DYNARER4300_ASSEM_X86_H' is defined here; did you mean 'M64P_DEVICE_R4300_NEW_DYNAREC_X86_ASSEM_X86_H'?
#define M64P_DEVICE_R4300_NEW_DYNARER4300_ASSEM_X86_H
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        M64P_DEVICE_R4300_NEW_DYNAREC_X86_ASSEM_X86_H
/home/travis/mupen64plus-ae/app/src/main/jni/./mupen64plus-core/src/device/r4300/new_dynarec/new_dynarec.c:2234:70: warning: comparison of unsigned expression >= 0 is always true [-Wtautological-compare]
      if((start!=0)&&(page<2048)&&((start-(u_int)g_dev.ri.rdram.dram)>=0)&&((end-(u_int)g_dev.ri.rdram.dram)<0x800000)) {
                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^ ~
```
```
/home/travis/mupen64plus-ae/app/src/main/jni/./mupen64plus-core/src/device/r4300/new_dynarec/arm/assem_arm.c:433:8:
warning:
initializing 'int *' with an expression of type 'u_int *' (aka 'unsigned int *') converts between pointers to integer types with different sign [-Wpointer-sign]
  int *ptr=((u_int *)addr)+6;
       ^   ~~~~~~~~~~~~~~~~~
/home/travis/mupen64plus-ae/app/src/main/jni/./mupen64plus-core/src/device/r4300/new_dynarec/new_dynarec.c:2234:70:
warning: 
comparison of unsigned expression >= 0 is always true [-Wtautological-compare]
      if((start!=0)&&(page<2048)&&((start-(u_int)g_dev.ri.rdram.dram)>=0)&&((end-(u_int)g_dev.ri.rdram.dram)<0x800000)) {
                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^ ~
/home/travis/mupen64plus-ae/app/src/main/jni/./mupen64plus-core/src/device/r4300/new_dynarec/new_dynarec.c:10899:35:
warning: 
passing 'u_char *' (aka 'unsigned char *') to parameter of type 'char *' converts between pointers to integer types with different sign [-Wpointer-sign]
  __clear_cache((void *)beginning,out);
                                  ^~~
```

loganmc10:
I just tested on my Android ARM device (Shield Tablet) and it all works fine, everything looks good.

bsmiles32:
Great ! I fixed the header guard warning as it was a simple one. However the other warnings (mostly signedess related) predate this PR and might require deeper work.

bsmiles32:
@loganmc10 : Did you noticed any speed regression with this PR ?
@loganmc10 , @Gillou68310 : Any objections before I merge that PR ?

loganmc10:
I didn't really look for that. There was no obvious problems, but I didn't look at the FPS or anything. I have no issues with merging it unless someone else wants to do some more detailed performance testing before merging it.

bsmiles32:
Good. Will wait for @Gillou68310 approval then :)

Gillou68310:
Approved ;-)

dankcushions:
just FYI, i just compiled master on my RPI3 and it seems to run fine :) no idea about performance changes, but the ARM dynarec previously didn't come close to maxing the CPU and still doesn't 👍

PR #259, opened 2017-04-14
Allow user-requested window position

Meriipu:
Reads the two g_CoreConfig options ScreenPosX and ScreenPosY
Sets the final window to this position, or returns to the old behaviour if unset (checks for default values -1,-1)

I feel these changes are much less clean than the changes I made to ui-console to allow storing the options.
I also have no idea whether screen coordinates work the same way on all platforms, or whether there are other cases where these changes might lead to issues.

Finally they might fill a pretty niche need. For me it was spawning multiple instances at once and having them tiled out rather than all on top of each other. 
My attempts to use external applications to move the windows when they were created were not as successful as this change was.


Meriipu:
I should probably learn how to use branches/releases sometime
I was a bit surprised to see the last newline change instantly appear

rlabrecque:
You're going to need to rebase and squash those commits then force push to your branch. Then there will be just once clean commit if/when this gets merged.

Meriipu:
I am not sure what is going on with the bsmiles32-commits

Meriipu:
no this includes a bunch of changes I did not make sorry. 
It might be better to close it and reopen it fixed later, I have no idea how to fix it.

PR #260, opened 2017-04-14
Allow user-requested window position

Meriipu:
Reads the two g_CoreConfig options ScreenPosX and ScreenPosY
Sets the final window to this position, or returns to the old behaviour if unset (checks for default values -1,-1)

I feel these changes are much less clean than the changes I made to ui-console to allow storing the options.
I also have no idea whether screen coordinates work the same way on all platforms, or whether there are other cases where these changes might lead to issues.

Finally they might fill a pretty niche need. For me it was spawning multiple instances at once and having them tiled out rather than all on top of each other.
My attempts to use external applications to move the windows when they were created were not as successful as this change was.

Remade request #259 as I had some issue with another pull request being mixed into it that I was not good enough at git to solve

### DIFF
--- a/src/api/vidext_sdl2_compat.h
+++ b/src/api/vidext_sdl2_compat.h
@@ -21,7 +21,7 @@
 
 #include <SDL_config.h>
 #include <SDL_surface.h>
-
+#include "main/main.h"
 #ifndef USE_GLES
 
 #ifndef SDL_VIDEO_OPENGL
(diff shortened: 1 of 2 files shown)


loganmc10:
The problem with this is that not every frontend uses SDL (mupen64plus-ui-console is just one of the frontends). So with this change you have 2 new config items for the core (ScreenPosX, ScreenPosY), but they only apply to mupen64plus-ui-console

Meriipu:
I see how that can be messy, having unapplicable options.

Are there other frontends than ui-python? If (in most cases) not, would including a similar change to qt work out? Or storing the settings in a different category?

Otherwise I would just keep them local to the fork I made

loganmc10:
Well I wrote/use this:
https://github.com/m64p/mupen64plus-gui, similar to ui-python but it's written in C++.

There have been others in the past (CuteMupen, mupen64plus-qt), some of them aren't really frontends though, they are just launchers that launch mupen64plus-ui-console.

I assume something like this could be accomplished via a shell script. Once mupen64plus has launched you could get the process ID and tell the window manager to re-position the window (I assume, I haven't looked into it).

Meriipu:
I tried that, but the race between which application window was created before another required too long delays between window spawns to avoid one spawning before the other and getting placed in the wrong location. I was launching several (15-30 instances) at the same time.

Placing it in the desired location by having the application request its position from the very beginning solved it for me.

loganmc10:
Have you tried using the environment variable ```SDL_VIDEO_WINDOW_POS=x,y```?

I just tested doing:

```
SDL_VIDEO_WINDOW_POS=0,0 ./mupen64plus ~/N64/Super\ Mario\ 64\ \(USA\).n64
```
And it did indeed put the window at the top left corner.

Meriipu:
That works. Nice. I suppose this change is pretty superfluous then.

loganmc10:
The environment variable trick will work for ui-console. It won't work for other frontends like ui-python, but other frontends can implement their own solutions for repositioning the window if needed, this can probably be closed

Meriipu:
Already implemented for sdl as environment variable SDL_VIDEO_WINDOW_POS

PR #263, opened 2017-04-23
Fix Indiana Jones

loganmc10:
These values were taken from Project64 and they don't seem to work here, the game doesn't boot with these settings


fzurita:
Hmm... good find.

PR #270, opened 2017-05-02
Remove custom CountPerScanline values

loganmc10:
When these were initially committed, I just copied the values from Project64. Given that that change broke Indiana Jones, it's probably best to remove these values. There is no evidence that they make any improvement, so it's safer to remove them.

At least now there is a core option to modify this value, so if someone is so inclined, they could test out different values to see if they make any difference for a particular game.

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -792,7 +792,6 @@ RefMD5=90448C4175EE9D0247674474DCABDFED
 GoodName=All-Star Baseball 2001 (U) [!]
 CRC=5446C6EF E18E47BB
 RefMD5=90448C4175EE9D0247674474DCABDFED
-CountPerScanline=1400
 
 [CD04ABC5979F22EF5D6FDC4DCFAA4D50]
 GoodName=All-Star Baseball 2001 (U) [f1] (PAL)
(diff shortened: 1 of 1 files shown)


fzurita:
Are you sure that it was CountPerScanline and not the CountPerOp setting that was breaking Indiana Jones?

loganmc10:
Sorry by "these settings" I mean the CountPerOp/CountPerScanline changes that I took from Project64

Setting CountPerOp to 1 on Indiana Jones breaks the sound for me (on x64), CountPerScanline doesn't seem to make any difference.

CountPerScanline doesn't really seem to make any difference for any game to be honest. Project64 and mupen64plus calculate the timing differently, so they might not do the same thing. My point is that since these settings weren't tested, they were just copied from Project64, they should be removed. If someone tests a game with a certain CountPerScanline, and it improves things, then it should be added. That was my error when I initially committed these values.

However, CountPerOp=3 for Beetle Adventure Racing (which was copied from Project64) does seem to be the right change. If you watch the intro, there is a point where the car jumps off a ramp and crashes into the water, then it says "Action!". With CountPerOp=2, "Action!" is displayed at the wrong time, with CountPerOp=3, it is displayed correctly, as confirmed via a YouTube clip somebody posted of a real N64 playing the intro. So I've kept that changed.

PR #279, opened 2017-05-17
Fix E and J cheat code values for MK64

loganmc10:


loganmc10:
I still need to test this so please don't merge yet

PR #282, opened 2017-05-18
No g dev in interrupt

bsmiles32:
Remove usage of g_dev in interrupt. The tricky commit is the last one which reworks how the first VI is triggered. This is because, the code which handle initialization, reset, NMI, pifbootrom, ... is quite bad. But I guess that's will be for another PR.


dankcushions:
Tested on RPI3 (ARM) + mario 64. Built and ran fine 👍

Issue #200, opened 2016-10-30
Top Gear Rally bad audio


fzurita:
I've been trying to figure out what's wrong with audio in this game, but I'm not having much luck. The one thing that seems to be special about this game is that audio comes in 16384 byte chunks to the audio plugin when `AiLenChanged` is called, normally I see around 2KB chunks.

I dumped these audio samples to a text file as they come into the audio plugin just to see what the values look like, I don't see anything unreasonable, but when audio is played back from this data, it doesn't sound right.

Attached are audio samples for that game in text format. Each row is a byte array that is read when AiLenChanged is called.

[audio_output.zip](https://github.com/mupen64plus/mupen64plus-core/files/560707/audio_output.zip)

At this point, I think that the audio coming from the core is not right. I have tried switching RSP between CXD4 and RSP-HLE with no effect.

Gillou68310:
Did you tried with cxd4 plugin?

fzurita:
Yeah, no improvement.

Gillou68310:
Sorry I just saw you already mentioned it in the first post.

fzurita:
No problem

Gillou68310:
What about pj64?

fzurita:
Very similar issue with default settings.

Gillou68310:
Ok this will probably be hard to fix then. We have no working case to compare with.

mudlord:
You mean you need working audio samples of the background music and stuff?

Top Gear Rally's music is tracked. You can get it from:
http://ftp.modland.com/pub/modules/Fasttracker%202/Barry%20Leitch/

loganmc10:
Now that World Driver Chamionship and Stunt Racer are working, they seem to have the same issue. Really bad audio, and the audio comes in 16384 byte chunks. The audio doesn't sound great in Project64 either.

loganmc10:
I've been looking into this, and I believe I know the cause. I'm going to write out all the details in case anyone in the future finds this interesting.

When a game wants to play some audio, it writes the address of the audio buffer to AI_DRAM_ADDR_REG, and it writes the length of the audio buffer to AI_LEN_REG. Writing to AI_LEN_REG triggers a DMA that starts audio playing.

The way we emulate this is: When AI_LEN_REG is written (DMA starts), we copy the contents of the audio buffer to the audio plugin and play it.

For a majority of games this works well, but Boss Games are unique.

[...]

PR #289, opened 2017-05-24
Fix Twisted Edge audio

### LINKED ISSUE #285: Twisted Edge Extreme Snowboarding no audio
There is a cheat in Project64 that makes it work:

```
"exhalatio audio hack",81051840 AD99
```

And in the AziAudio plugin it was fixed this way:
https://github.com/Azimer/AziAudio/pull/181/commits/7140f74be33615b4f9c62164e5e7e4d261cc6e48

I tried both methods in mupen64plus without success, still no audio

loganmc10:
Fixes #285 

This is a combination of these 2 commits:
https://github.com/project64/project64/commit/f24c464f3f8783dd9fdbeeaa21b02658c1340ec3

https://github.com/Azimer/AziAudio/commit/7140f74be33615b4f9c62164e5e7e4d261cc6e48

Firstly, AI_LEN_REG should be a multiple of 8, so I modified ```get_remaining_dma_length``` accordingly. This may fix other audio issues if they exist, I'm not sure. This is what allowed the audio samples to start getting sent to the audio plugin.

Secondly, Twisted Edge has some strange behaviour with AI_DRAM_ADDR_REG, it points to the wrong address. The "FixedAudioPos" hack takes the first AI_DRAM_ADDR_REG value, and uses that for the rest of the game, ignoring future values.

I compared to Project64 and it sounds the same to me.

### DIFF
--- a/src/device/ai/ai_controller.c
+++ b/src/device/ai/ai_controller.c
@@ -56,7 +56,8 @@ static uint32_t get_remaining_dma_length(struct ai_controller* ai)
 
     remaining_dma_duration = next_ai_event - cp0_regs[CP0_COUNT_REG];
 
-    return (uint64_t)remaining_dma_duration * ai->fifo[0].length / ai->fifo[0].duration;
+    uint64_t dma_length = (uint64_t)remaining_dma_duration * ai->fifo[0].length / ai->fifo[0].duration;
+    return dma_length&~7;
 }
 
 static unsigned int get_dma_duration(struct ai_controller* ai)
@@ -139,12 +140,13 @@ void init_ai(struct ai_controller* ai,
              struct r4300_core* r4300,
              struct ri_controller* ri,
              struct vi_controller* vi,
-             struct audio_out_backend* aout)
+             struct audio_out_backend* aout, unsigned int fixed_audio_pos)
 {
     ai->r4300 = r4300;
     ai->ri = ri;
     ai->vi = vi;
     ai->aout = aout;
+    ai->fixed_audio_pos = fixed_audio_pos;
 }
 
 void poweron_ai(struct ai_controller* ai)
@@ -152,6 +154,7 @@ void poweron_ai(struct ai_controller* ai)
     memset(ai->regs, 0, AI_REGS_COUNT*sizeof(uint32_t));
     memset(ai->fifo, 0, AI_DMA_FIFO_SIZE*sizeof(struct ai_dma));
     ai->samples_format_changed = 0;
+    ai->audio_pos = 0;
 }
 
 int read_ai_regs(void* opaque, uint32_t address, uint32_t* value)
@@ -178,6 +181,15 @@ int write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
 
     switch (reg)
     {
+    case AI_DRAM_ADDR_REG:
+        masked_write(&ai->regs[AI_DRAM_ADDR_REG], value, mask);
+        if (ai->fixed_audio_pos) {
+            if (ai->audio_pos == 0)
+                ai->audio_pos = ai->regs[AI_DRAM_ADDR_REG];
+            ai->regs[AI_DRAM_ADDR_REG] = ai->audio_pos;
+        }
+        return 0;
+
     case AI_LEN_REG:
         masked_write(&ai->regs[AI_LEN_REG], value, mask);
         fifo_push(ai);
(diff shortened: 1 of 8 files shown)


loganmc10:
This change (specifically making sure that ```get_remaining_dma_length``` returns a multiple of 8) also fixes the choppy engine sound in Top Gear Rally (https://github.com/mupen64plus/mupen64plus-core/issues/200) @fzurita

fzurita:
Cool! I'll have to try it out.

src/device/ai/ai_controller.c
```diff
@@ -139,12 +140,14 @@ void init_ai(struct ai_controller* ai,
...
              struct vi_controller* vi,
-             struct audio_out_backend* aout)
+             struct audio_out_backend* aout, unsigned int fixed_audio_pos)
 {
     ai->r4300 = r4300;
     ai->ri = ri;
     ai->vi = vi;
     ai->aout = aout;
+    ai->fixed_audio_pos = fixed_audio_pos;
+    ai->audio_pos = 0;
```
bsmiles32: move ai->audio_pos = 0 in poweron_ai as it is a "state" variable and not a "configure" variable.

src/device/device.c
```diff
@@ -53,7 +53,7 @@ void init_device(struct device* dev,
     uint16_t eeprom_id, struct storage_backend* eeprom_storage,
     struct clock_backend* clock,
     /* vi */
-    unsigned int vi_clock, unsigned int expected_refresh_rate, unsigned int count_per_scanline, unsigned int alternate_timing)
+    unsigned int vi_clock, unsigned int expected_refresh_rate, unsigned int count_per_scanline, unsigned int alternate_timing, unsigned int fixed_audio_pos)
```
bsmiles32: move fixed_audio_pos parameter upper in the parameter list be in the AI section please.

src/device/device.h
```diff
@@ -78,7 +78,7 @@ void init_device(struct device* dev,
     uint16_t eeprom_id, struct storage_backend* eeprom_storage,
     struct clock_backend* clock,
     /* vi */
-    unsigned int vi_clock, unsigned int expected_refresh_rate, unsigned int count_per_scanline, unsigned int alternate_timing);
+    unsigned int vi_clock, unsigned int expected_refresh_rate, unsigned int count_per_scanline, unsigned int alternate_timing, unsigned int fixed_audio_pos);
```
bsmiles32: same here

src/main/main.c
```diff
@@ -1065,7 +1065,7 @@ m64p_error main_run(void)
                 gb_carts,
                 (ROM_SETTINGS.savetype != EEPROM_16KB) ? 0x8000 : 0xc000, &eep_storage,
                 &clock,
-                vi_clock_from_tv_standard(ROM_PARAMS.systemtype), vi_expected_refresh_rate_from_tv_standard(ROM_PARAMS.systemtype), count_per_scanline, alternate_vi_timing);
+                vi_clock_from_tv_standard(ROM_PARAMS.systemtype), vi_expected_refresh_rate_from_tv_standard(ROM_PARAMS.systemtype), count_per_scanline, alternate_vi_timing, ROM_PARAMS.fixedaudiopos);
```
bsmiles32: same

src/plugin/emulate_speaker_via_audio_plugin.c
```diff
@@ -53,6 +53,11 @@ void push_audio_samples_via_audio_plugin(void* user_data, const void* buffer, si
 
     /* exploit the fact that buffer points in g_dev.ri.rdram.dram to retreive dram_addr_reg value */
     ai->regs[AI_DRAM_ADDR_REG] = (uint8_t*)buffer - (uint8_t*)ai->ri->rdram.dram;
+    if (ai->fixed_audio_pos) {
```
bsmiles32: This logic shouldn't be here as it is only meant as a "compatibility layer" to use zilmar spec audio plugins. Not a core behavior. This should go in write_ai_regs. Just add a case AI_DRAM_ADDR_REG in there and these lines.

bsmiles32:
Did a quick review of this PR. While it fixes a known issue, I don't like these kind of hacks as I feel it is a missed opportunity for improve emulation accuracy. Still I won't object to merge it when it's ready. Good job none the less !!!

loganmc10:
Ok I've updated the PR. I agree that a hack is not ideal. At least it's fairly non-intrusive/small. Hopefully somebody in the future can figure out how it should work properly.

PR #296, opened 2017-05-28
Fix undefined behavior in the interpreter

### LINKED ISSUE #87: Signed integer overflow in core interpreter
Signed integer overflow/underflow aren't well defined by the C standard and should be avoided. Problem can be seen when compiling like explained below and testing the save file attached to the thread "Ocarina of Time - Crash with Glide64mk2 (Bug Tracking) https://groups.google.com/forum/#!topic/mupen64plus/7tgGftNKNEA"

```
$ wget https://github.com/mupen64plus/mupen64plus-core/raw/master/tools/m64p_helper_scripts.tar.gz
$ tar xvf m64p_helper_scripts.tar.gz
$ ./m64p_get.sh
$ CFLAGS="-g3 -fsanitize=undefined -fsanitize=address" ./m64p_build.sh OPTFLAGS="" INSTALL_STRIP_FLAG="" -j5
$ cd test
$ rm ~/.config/mupen64plus/mupen64plus.cfg
$ (cd ./test/ && ./mupen64plus --gfx ./mupen64plus-video-glide64mk2.so --datadir `pwd` ../zeldaoot.v64  ) 

../../src/r4300/interpreter_special.def:275:18: runtime error: signed integer overflow: -1860206572 + -551095242 cannot be represented in type 'int'
../../src/r4300/interpreter_special.def:289:18: runtime error: signed integer overflow: -2147463642 - 2147360282 cannot be represented in type 'int'
```

This is especially interesting because the calculation ADDU and SUBU (275 and 289) are for unsigned and not for signed calculations. Misusing signed integers with underflows can lead to misoptimizations in newer GCC versions. For example endless loops. I think @cxd4 can enlighten us more because he already experienced it himself when he wrote some benchmarks for his rsp plugin

comex:
Partially fixes #87, but there are probably more potential overflows lurking in there; I didn't go over it too thoroughly.

FYI: In C, integer overflow when adding and multiplying signed integers is undefined, as is shifting any integer by too high a width, and left-shifting a signed integer if it changes the sign.  GCC started performing optimizations based on these a decade ago, but those optimizations don't usually apply in any particular situation, so usually you just get away with it.

But on my system, with clang-802.0.42, Mario 64 actually fails to boot in the interpreter without this patch: it fails the ROM checksum test performed early on (at 0x800001AC).  (Not sure exactly which operation was responsible; I compiled with -fsanitize=undefined, fixed everything that made it crash under that and some nearby code, and now Mario 64 works.  However, all of the changes here should be correct/necessary.)

### DIFF
--- a/src/device/r4300/mips_instructions.def
+++ b/src/device/r4300/mips_instructions.def
@@ -102,7 +102,7 @@ DECLARE_INSTRUCTION(RESERVED)
 DECLARE_INSTRUCTION(LB)
 {
     DECLARE_R4300
-    const uint32_t lsaddr = irs32 + iimmediate;
+    const uint32_t lsaddr = (uint32_t) irs32 + (uint32_t) iimmediate;
     int64_t *lsrtp = &irt;
     ADD_TO_PC(1);
     *r4300_address(r4300) = lsaddr;
@@ -116,7 +116,7 @@ DECLARE_INSTRUCTION(LB)
 DECLARE_INSTRUCTION(LBU)
 {
     DECLARE_R4300
-    const uint32_t lsaddr = irs32 + iimmediate;
+    const uint32_t lsaddr = (uint32_t) irs32 + (uint32_t) iimmediate;
     int64_t *lsrtp = &irt;
     ADD_TO_PC(1);
     *r4300_address(r4300) = lsaddr;
@@ -127,7 +127,7 @@ DECLARE_INSTRUCTION(LBU)
 DECLARE_INSTRUCTION(LH)
 {
     DECLARE_R4300
-    const uint32_t lsaddr = irs32 + iimmediate;
+    const uint32_t lsaddr = (uint32_t) irs32 + (uint32_t) iimmediate;
     int64_t *lsrtp = &irt;
     ADD_TO_PC(1);
     *r4300_address(r4300) = lsaddr;
@@ -141,7 +141,7 @@ DECLARE_INSTRUCTION(LH)
 DECLARE_INSTRUCTION(LHU)
 {
     DECLARE_R4300
-    const uint32_t lsaddr = irs32 + iimmediate;
+    const uint32_t lsaddr = (uint32_t) irs32 + (uint32_t) iimmediate;
     int64_t *lsrtp = &irt;
     ADD_TO_PC(1);
     *r4300_address(r4300) = lsaddr;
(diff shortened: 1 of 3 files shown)


comex:
Added another fix.

src/device/r4300/mips_instructions.def
```diff
@@ -723,28 +723,28 @@ DECLARE_INSTRUCTION(SLLV)
 DECLARE_INSTRUCTION(DSLL)
 {
     DECLARE_R4300
-    rrd = rrt << rsa;
+    rrd = (uint64_t) rrt << (rsa & 0x3F);
```
bsmiles32: sa is only 5 bit here, not 6 ?

src/device/r4300/mips_instructions.def
```diff
@@ -758,7 +758,7 @@ DECLARE_INSTRUCTION(SRLV)
 DECLARE_INSTRUCTION(DSRL)
 {
     DECLARE_R4300
-    rrd = (uint64_t) rrt >> rsa;
+    rrd = (uint64_t) rrt >> (rsa & 0x3F);
```
bsmiles32: same here sa is 5 bits not 6 ?

src/device/r4300/mips_instructions.def
```diff
@@ -723,28 +723,28 @@ DECLARE_INSTRUCTION(SLLV)
...
-    rrd = rrt << (rrs32 & 0x3F);
+    rrd = (uint64_t) rrt << (rrs32 & 0x3F);
     ADD_TO_PC(1);
 }
 
 DECLARE_INSTRUCTION(DSLL32)
 {
     DECLARE_R4300
-    rrd = rrt << (32 + rsa);
+    rrd = (uint64_t) rrt << ((32 + rsa) & 0x3F);
```
bsmiles32: I would prefer (32 + (rsa & 0x1f)) as you did with DSRA32

src/device/r4300/mips_instructions.def
```diff
@@ -793,7 +793,7 @@ DECLARE_INSTRUCTION(SRAV)
 DECLARE_INSTRUCTION(DSRA)
 {
     DECLARE_R4300
-    rrd = rrt >> rsa;
+    rrd = rrt >> (rsa & 0x3F);
```
bsmiles32: same here sa is 5 bit ?

bsmiles32:
Thanks for your contribution. I had a quick look on your PR and except for the shift masks which seems to differ from the official documentation, it looks OK.

comex:
Apologies for the incorrect masks; fixed.

comex:
Er, except I guess the masks on rsa aren't needed in the first place since it's extracted from a 5-bit field to start with.  Oops.  The only actual issue with the shift instructions is the signed left shifting in DSLL and DSLLV.

richard42:
I don't think all of the casting 32-bit integers to (uint32_t) will change the output in any way. Adding is sign-agnostic

PR #297, opened 2017-05-28
Fix M64P_GL_SWAP_CONTROL with SDL>=1.3 and no video extension

comex:
It originally mapped to SDL_GL_SWAP_CONTROL, which was removed in SDL 1.3; whoever wrote this decided to replace it with SDL_RENDERER_PRESENTVSYNC.  However, SDL_RENDERER_PRESENTVSYNC is not a GL attribute: it's a setting used when creating a 2D renderer, which this code doesn't use at all.  If passed to SDL_GL_(Get|Set)Attribute, it'll be interpreted as the SDL_GLattr value which happens to have the same numeric value, 4, namely SDL_GL_BUFFER_SIZE.

This changes the code to use the functions SDL_GL_(Get|Set)SwapInterval instead.


bsmiles32:
I'll trust you on that, and revert if somebody complains :)

PR #301, opened 2017-05-30
Avoid usage of global variable g_delay_si.

bsmiles32:
One less global variable used inside the emulation core.


dankcushions:
built and tested on RPI3/ARM + mario64 - fine here :)

Issue #294, opened 2017-05-27
Save states are locking up the emulator


fzurita:
Some times, on the latest commit as of right now, the save states being generated are locking up the emulator when loaded. I have tried different RSP and video plugins. This didn't happen before, so something broke this recently. I have also only tried the ARM new dynarec.

This is specially bad with Goldeneye, where it is happening about half the time.

fzurita:
I have the feeling that we are not waiting for the correct interrupt when making a save state.

loganmc10:
I tested this on the x64 "old" dynarec and x86 new dynarec. It seemed to work fine on the x64 dynarec. I got it to freeze on the x86 new dynarec, but it took a lot of tries, definitely not 1/2

Here is the bad save state, it freezes the emulator if you try and load it.

[GoldenEye.zip](https://github.com/mupen64plus/mupen64plus-core/files/1033933/GoldenEye.zip)

fzurita:
Well, it's luck. Maybe I got unlucky and it froze for me more often. It seemed really bad in the intro of the game.

bsmiles32:
Could it be related to changes in commit a21299985fb9752244174c5b5083624735faa858 where I reworked the first triggering of VI interrupt ?

fzurita:
I can try reverting it and checking if it happens again.

fzurita:
Ok, I was able to reproduce the issue after reverted https://github.com/mupen64plus/mupen64plus-core/commit/a21299985fb9752244174c5b5083624735faa858

I'm going to try reverting all recent commits that changed interrupt handling code and see if anything fixes it.

fzurita:
Ok, this turns out that it's not a recent issue. I just wasn't looking for it before. I went back all the way to this commits before all the mayor refactors and it still happens:

https://github.com/mupen64plus/mupen64plus-core/commits/master?after=e8baea699fc6af8a4b91d8a038ee04ac92f78bbb+139

bsmiles32:
Could this be related to TLB mappings not being saved/restored correctly in the new_dynarec ?

loganmc10:
I managed to get GoldenEye to do this again with the x86 new dynarec. But it's really rare for me, we are talking maybe 1/100 save states (I just tap F5 and then F7 over and over again until it freezes). I couldn't get it to do it with the x64 dynarec, but I don't know for sure if it's not possible there, because like I said, it seems really rare.

fzurita:
[...]

PR #302, opened 2017-05-30
Be more selective on which interrupts we can load or save states

fzurita:
This seems to fix the issue encountered in https://github.com/mupen64plus/mupen64plus-core/issues/294.

I have no idea if this is the correct fix or what kind of side effects it could have, but after making this change, I'm unable to reproduce the above issue.

### DIFF
--- a/src/device/r4300/mi_controller.c
+++ b/src/device/r4300/mi_controller.c
@@ -103,7 +103,9 @@ int write_mi_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
 
         check_interrupt(r4300);
         cp0_update_count(r4300);
+        r4300->cp0.interrupt_unsafe_state = 1;
         if (*cp0_next_interrupt <= cp0_regs[CP0_COUNT_REG]) gen_interrupt(r4300);
+        r4300->cp0.interrupt_unsafe_state = 0;
         break;
     }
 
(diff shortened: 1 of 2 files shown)


loganmc10:
Is this still a problem? I've still never been able to reproduce this issue on Android

fzurita:
I don't know, I have this workaround included in the play store version of Mupen64plus AE so I don't see it there and that's the one I usually use.

I'll give it a shot soon.

fzurita:
I tried for a while and couldn't make it happen. It was very easily reproducible for me before. I'll close and re-open if the issue shows up again.

fzurita:
I just had the issue happen in Golden Eye. I guess it's still happening but much rarer.

loganmc10:
Do you still have the save state? I remember when I was testing this, I did have GoldenEye freeze when trying to load the save state once (on Android), but when I tried to open the save state using the cached interpreter, it worked, just not with the arm dynarec

fzurita:
I'm pretty sure the issue is due to TLB, on the save states where the issue happened, TLB address translation did not seem to be working correctly. the new dynarec has its own TLB implementation. 

I'll check if I still have the save state.

fzurita:
Here is the save state:

[golden_eye_freeze.sav.zip](https://github.com/mupen64plus/mupen64plus-core/files/1318490/golden_eye_freeze.sav.zip)

loganmc10:
Yeah I can open that savestate on my PC no problem, doesn't freeze. I think it's something up with the new dynarec

PR #306, opened 2017-06-02
INI cheat cleanup

loganmc10:
Mainly, remove Pokemon Snap cheats.

You can read about why these Pokemon Snap cheats were added in the first place here: https://github.com/mupen64plus/mupen64plus-core/commit/252fa6f1d42197b88456fae8e64b607cfaea183b

Basically, the tutorial doesn't unlock the controls until you take a picture. Most graphics plugins aren't accurate enough to recognize when you've captured a Pokemon, so the controls never unlock. The cheat forced the controls to unlock. Kind of a silly thing to do, since you still can't get it to recognize the Pokemon in your picture.

Anyway, accurate graphics plugins like Angrylion or GLideN64 allow you to take a picture, so the controls unlock. This hack isn't needed, and probably should have never been added in the first place. I played through the whole first level without this cheat using GLideN64, and everything works fine, including Professor Oak's picture check, etc...

The rest is mostly cosmetic, reordering the Zelda cheat from 1 to 0. Removing the commented out hack to produce shadows in Pilot Wings. Again, an accurate graphics plugin like GLideN64 shows the shadows. I also removed all the Mario Kart multiplayer timing hacks from everything but the main entry, which keeps it in line with the other ROMs.


loganmc10:
Here is a screenshot showing that the shadows work without that hack (I assume this is what it means by "land shadows"):
pilot_wings64-000

Issue #181, opened 2016-09-11
Top Gear Really 2 freezes


fzurita:
I'm heading a lot that Top Gear Rally 2 is freezing mid race or right after the race finishes. This is regardless of RSP or video plugins, or interpreter or dynarec mode.

This is similar to the Tarzan freezes that were just fixed.

loganmc10:
I tested this on Linux and it froze all the time with:

```
Core Warning: two events of type 0x80 in interrupt queue
```

Seemingly random, but quite often, could never get through a full lap.

Setting ```CountPerOp=1``` fixed the problem for me, did a few laps without freezing. Perhaps someone else could confirm. If that does fix it then it can be added to the ini

loganmc10:
Actually with ```CountPerOp=1``` I get a ```Core Warning: two events of type 0x80 in interrupt queue``` error and a freeze in the menu if I let it sit there. I get the same thing with CountPerOp=2

However, if I comment this out:
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/interrupt.c#L155

Then I still get ``Core Warning: two events of type 0x80 in interrupt queue```, but no freeze. It seems that hack should be specific to Perfect Dark.

Since I rambled a bit: I think the best way to fix this game is set CountPerOp=1 and make that interrupt hack specific to Perfect Dark

On a side note, @bsmiles32. I notice that you were working on this: https://github.com/bsmiles32/mupen64plus-core/commit/9b98afade1ea609dc738108709874339829629d0.

Is this (https://github.com/bsmiles32/mupen64plus-core/commit/a1b6df0800dd86936b8500cb7b731b10ee8d79ac) still possible? Can we do that and get rid of the Perfect Dark hack altogether?

bsmiles32:
These commits are old and I unfortunately don't remember exactly why I did them this way... Will have to rethink why. Sorry

loganmc10:
No problem, I submitted a PR to look at (https://github.com/mupen64plus/mupen64plus-core/issues/307) that implements that idea. So far my testing results are quite promising, and it fixes the crashes in this game without having to play with the CountPerOp settings. But obviously that kind of change will need quite a bit of testing to avoid regressions.

fzurita:
This one should be closed.

PR #307, opened 2017-06-02
Trust plugins to trigger SP interrupts

loganmc10:
This is based on work by @bsmiles32 in https://github.com/bsmiles32/mupen64plus-core/commit/a1b6df0800dd86936b8500cb7b731b10ee8d79ac, brought on by debugging https://github.com/mupen64plus/mupen64plus-core/issues/181

I've tested this ever so briefly on a few games like Super Mario 64 and World Driver Champ. I'm just posting this here to get some feedback and perhaps some testing.

I've also posed a few questions in comments in the code that I was a little unclear on, I'd appreciate if anyone has some insight on those lines.

### DIFF
--- a/src/device/r4300/interrupt.c
+++ b/src/device/r4300/interrupt.c
@@ -148,11 +148,6 @@ void add_interrupt_event_count(struct cp0* cp0, int type, unsigned int count)
 
     if (get_event(&cp0->q, type)) {
         DebugMessage(M64MSG_WARNING, "two events of type 0x%x in interrupt queue", type);
-        /* FIXME: hack-fix for freezing in Perfect Dark
-         * http://code.google.com/p/mupen64plus/issues/detail?id=553
-         * https://github.com/mupen64plus-ae/mupen64plus-ae/commit/802d8f81d46705d64694d7a34010dc5f35787c7d
-         */
-        return;
     }
 
     event = alloc_node(&cp0->q.pool);
--- a/src/device/rdp/rdp_core.c
+++ b/src/device/rdp/rdp_core.c
@@ -143,10 +143,6 @@ void rdp_interrupt_event(void* opaque)
 {
     struct rdp_core* dp = (struct rdp_core*)opaque;
 
-    dp->dpc_regs[DPC_STATUS_REG] &= ~DPC_STATUS_FREEZE;
-    dp->dpc_regs[DPC_STATUS_REG] |= DPC_STATUS_XBUS_DMEM_DMA
-        | DPC_STATUS_CBUF_READY;
-
     raise_rcp_interrupt(dp->r4300, MI_INTR_DP);
 }
 
--- a/src/device/rsp/rsp_core.c
+++ b/src/device/rsp/rsp_core.c
@@ -284,21 +284,11 @@ void do_SP_Task(struct rsp_core* sp)
         sp->regs2[SP_PC_REG] |= save_pc;
         new_frame();
 
-        cp0_update_count(sp->r4300);
-        sp->rsp_task_locked = 0;
-        if ((sp->regs[SP_STATUS_REG] & (SP_STATUS_HALT | SP_STATUS_BROKE)) == 0)
-        {
-            sp->rsp_task_locked = 1;
-            sp->r4300->mi.regs[MI_INTR_REG] |= MI_INTR_SP;
-        }
-        if (sp->r4300->mi.regs[MI_INTR_REG] & MI_INTR_SP) {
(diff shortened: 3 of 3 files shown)


loganmc10:
I've tested this with quite a few games now, both HLE and LLE graphics plugins. It seems pretty solid, I don't see any regressions. It does fix the crashing in Top Gear Rally 2, and removes the need for the Perfect Dark hack. I assume it also fixes other stability problems as well.

There are still a few items that I'm not sure if I'm getting right, does ```cp0_update_count(sp->r4300);``` need to appear before every ```add_interrupt_event()``` call?

Also, in the current version it schedules a DP interrupt with a value/counter of 1000. This is necessary for Pokemon detection in Pokemon Snap (probably other things as well). After this change, 1000 no longer worked, but 4000 does. Project64 uses 0x1000 (4096), which is where I got the idea for 4000. But is 4000 right? How are these numbers determined?

bsmiles32:
add_interrupt_event needs an updated COUNT register as all events are compared using it. That's why we make sure to call cp0_update_count before.
As for the other values, I really don't know where they come from. I guess it's purely empirical as everything timing related  is unfortunately.

loganmc10:
Ok understood. I've updated the PR to cleanup the code. I've tested this with quite a few games. Including the ones that have been more sensitive to RSP interrupt changes (Perfect Dark, Kirby, Banjo 1/2, Blast Corps, Pokemon Snap, etc...).

I've played several races in Top Gear Rally 2 and several levels in Perfect Dark, no more crashing.

bsmiles32:
@loganmc10 I also tested on my end and it seems to work correctly. I'm not sure why it works correctly though. I'll merge it now and revert if somebody complains.

fzurita:
I'll get this out to a wider audience and see how many people complain or are happy about it.

Gillou68310:
I don't think this implementation is correct, I don't even understand why it works because no SP interrupts are ever triggered when the RSP is done with a task.

On the real hardware the CPU and RSP run in parallel:
-The CPU starts the RSP
-The RSP process its task while the CPU is still running.
-When the RSP is done with its task it causes an SP interrupt on the CPU side.

On the emulator the CPU and RSP are obviously not running in parallel:
-The CPU starts the RSP
-The RSP process its task
-When the RSP is done with its task an SP interrupt is scheduled with a configurable delay.
-The CPU runs until the SP interrupt is triggered.

The SP interrupt configurable delay is an approximation of the time the RSP spends executing its task.

@loganmc10 If you see this ```Core Warning: two events of type 0x80 in interrupt queue``` it's because the CPU is starting the RSP before the last SP interrupt has been triggered. You could try reducing the configurable delay to see if it fixes the issue.

Now that I think about it we should probably clear ```SP_STATUS_BROKE``` and ```SP_STATUS_HALT``` bits when returning from ```doRspCycles``` and set them back in ```rsp_interrupt_event```.

loganmc10:
> I don't even understand why it works because no SP interrupts are ever triggered when the RSP is done with a task.

The RSP plugin sets the interrupt bit: https://github.com/mupen64plus/mupen64plus-rsp-hle/blob/master/src/hle.c#L165

So basically this is the difference between having "DelaySI" on or off. With the way it is now, there is no delay, the SP interrupt is raised the next time interrupts are checked. With the old implementation, we would check if the RSP plugin set the MI interrupt bit, if it did, we would unset it, and schedule the RSP interrupt for later.

The problem with the old implementation is that, like you mentioned, sometimes it would "double up" and schedule a new RSP task before the old one was finished, which caused issues in Top Gear Rally 2, Perfect Dark, and probably other games.

You are probably right, perhaps if we cleared HALT and BROKE after the RSP task, then it would know not to start a new task. I'll try that out sometime today.

loganmc10:
@Gillou68310 actually now that I really look at the RSP plugin, it does:

```
*hle->mi_intr |= MI_INTR_SP;
HleCheckInterrupts(hle->user_defined);
```

So it actually checks the interrupt itself as well, by having the emulator core also call a delayed interrupt, wouldn't we be calling 2 interrupts for the same RSP task?

Gillou68310:
IIRC the ```CheckInterrupts``` API points to an ```EmptyFunc```

Gillou68310:
I just checked and ```rsp_interrupt_event``` is never called!

loganmc10:
No I totally agree, see https://github.com/mupen64plus/mupen64plus-core/pull/371

Gillou68310:
Well I was not expecting such great results :-)

PR #308, opened 2017-06-02
Add note for NBA Showtime - NBA on NBC

loganmc10:
I know you've all been dying to play this title, so I thought I would add a note on how to get it to work. I'm not sure where else this kind of thing could go, so I put it here. Might be helpful in the future to someone that tries to get it working.


bsmiles32:
Not sure where to put that info either but I'm OK with that.

Issue #309, opened 2017-06-05
Space Invaders won't start


loganmc10:
```Core Error: MFC0 instruction reading un-implemented Random register```

Tried Pure Interpreter, Cached Interpreter, and Dynarec.

loganmc10:
I'm curious what the purpose of that check is, commenting out these lines:
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/mips_instructions.def#L1356-L1359

Allows the game to boot, and it all seems to be working properly for me.

loganmc10:
Simply commenting out ``` *r4300_stop(r4300)=1;``` allows the game to work, the error message is just printed once at the beginning of the game

loganmc10:
I tried to find where that line originated, it appears to be very old, probably from Mupen64. The error message originally read "lecture de Random", which translates to "Random reading".

In reading the r4300i datasheet (http://datasheets.chipdb.org/MIPS/R4300i_datasheet.pdf), it doesn't seem to forbid reading that register, it just says that it's read-only.

PR #310, opened 2017-06-05
Fix reading CP0_RANDOM_REG

loganmc10:
From the r4300i datasheet:
>  This register will decrement on every instruction executed. The values range between a low value determined by the TLB Wired register, and an upper bound of TLBENTRIES-1. The TLB Random register is used to specify the entry in the TLB affected by the TLBWR instruction

Right now, CP0_RANDOM_REG is only updated before a TLB Write (https://github.com/loganmc10/mupen64plus-core/blob/b3b5d5cae1ba97013a54ad8b9410a725b67449aa/src/device/r4300/mips_instructions.def#L1332-L1334) because that is the only time it is really useful.

However, the value can also be read from, so this change updates the value of CP0_RANDOM_REG before it is read.

Fixes https://github.com/mupen64plus/mupen64plus-core/issues/309

### DIFF
--- a/src/device/r4300/mips_instructions.def
+++ b/src/device/r4300/mips_instructions.def
@@ -1354,15 +1354,15 @@ DECLARE_INSTRUCTION(MFC0)
     switch(rfs)
     {
     case CP0_RANDOM_REG:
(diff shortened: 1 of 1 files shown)


bsmiles32:
Tested. Works great. Merging :)

Issue #311, opened 2017-06-05
International Track & Field 2000 won't start (TLB Exception)


loganmc10:
Game doesn't boot, this is printed over and over:

```
tlb exception !!! @ 20, 1, add:80021ec4
```

loganmc10:
I have a very similar issue with Midway's Greatest Arcade Hits Vol. 1 (won't boot):

```
tlb exception !!! @ 20, 1, add:80051714
```

I wonder if this is related to https://github.com/mupen64plus/mupen64plus-core/issues/221. They all have an ```add`` value of over 0x80000000

loganmc10:
Both of these games work with CEN64, so I assume it's not the same problem as #221

loganmc10:
Here is the info for the TLB entry that is written right before this starts happening (these numbers are in hex):


![image](https://user-images.githubusercontent.com/848146/27146246-e7029f92-50f5-11e7-9023-7071f7dd95b0.png)

fzurita:
Sounds like a similar issue to what I was having in Goldeneye when loading save states. The TLB mapping was giving a completely bogus memory address.

loganmc10:
Yeah, I don't know if it's the TLB or something else. I got rid of the current TLB implementation and replaced it with CEN64's (it's not terribly hard since the code structure is really similar), other games that use TLB like Conkers/Mario 64/Golden Eye still worked, but this game still didn't.

loganmc10:
Potentially a similar issue in Rat Attack!, which also doesn't start, however this is only printed once:

```
tlb exception !!! @ 0, 1, add:8009b8a8
```

loganmc10:
I believe I've located the condition that is unique to these games. The MIPS "load instructions" (inside https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/mips_instructions.def) have the conditional statement:

```
if (*r4300_address(r4300)) {
```
That statement is normally true, but in these games, it evaluates to false once, that is where the game locks up, specifically in the LB and LW instructions. I'm not sure what to do in the "else" case, but it seems like they aren't being handled properly.

bsmiles32:
This check on address is "misleading". It checks if address translation succeded or not. This is a "legacy" of liberal use of global variables.
I think at some point we should rework the TLB implementation. Maybe after the big memory refactorings.

loganmc10:
[...]

PR #312, opened 2017-06-08
Remove memd

bsmiles32:
Some more refactorings which allows to get rid of dword memory accessors. Some more will come in this PR or later PR depending on feedback. Please test and report regressions.

### DIFF
--- a/src/device/ai/ai_controller.c
+++ b/src/device/ai/ai_controller.c
@@ -164,7 +164,7 @@ void poweron_ai(struct ai_controller* ai)
     ai->delayed_carry = 0;
 }
 
-int read_ai_regs(void* opaque, uint32_t address, uint32_t* value)
+void read_ai_regs(void* opaque, uint32_t address, uint32_t* value)
 {
     struct ai_controller* ai = (struct ai_controller*)opaque;
     uint32_t reg = ai_reg(address);
@@ -184,11 +184,9 @@ int read_ai_regs(void* opaque, uint32_t address, uint32_t* value)
     {
         *value = ai->regs[reg];
     }
-
-    return 0;
 }
 
-int write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
+void write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
 {
     struct ai_controller* ai = (struct ai_controller*)opaque;
     uint32_t reg = ai_reg(address);
@@ -198,11 +196,11 @@ int write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
     case AI_LEN_REG:
         masked_write(&ai->regs[AI_LEN_REG], value, mask);
         fifo_push(ai);
-        return 0;
+        return;
 
     case AI_STATUS_REG:
         clear_rcp_interrupt(ai->r4300, MI_INTR_AI);
-        return 0;
+        return;
 
     case AI_BITRATE_REG:
     case AI_DACRATE_REG:
@@ -211,12 +209,10 @@ int write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
             ai->samples_format_changed = 1;
 
         masked_write(&ai->regs[reg], value, mask);
-        return 0;
+        return;
     }
 
     masked_write(&ai->regs[reg], value, mask);
-
-    return 0;
 }
 
 void ai_end_of_dma_event(void* opaque)
--- a/src/device/ai/ai_controller.h
+++ b/src/device/ai/ai_controller.h
@@ -77,8 +77,8 @@ void init_ai(struct ai_controller* ai,
 
 void poweron_ai(struct ai_controller* ai);
 
-int read_ai_regs(void* opaque, uint32_t address, uint32_t* value);
-int write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask);
+void read_ai_regs(void* opaque, uint32_t address, uint32_t* value);
+void write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask);
 
 void ai_end_of_dma_event(void* opaque);
 
--- a/src/device/memory/memory.c
+++ b/src/device/memory/memory.c
@@ -45,22 +45,20 @@
 #include <stddef.h>
 #include <stdint.h>
 
-typedef int (*readfn)(void*,uint32_t,uint32_t*);
-typedef int (*writefn)(void*,uint32_t,uint32_t,uint32_t);
+typedef void (*readfn)(void*,uint32_t,uint32_t*);
+typedef void (*writefn)(void*,uint32_t,uint32_t,uint32_t);
 
 
-static int readw(readfn read_word, void* opaque, uint32_t address, uint64_t* value)
+static void readw(readfn read_word, void* opaque, uint32_t address, uint64_t* value)
 {
     uint32_t w;
-    int result = read_word(opaque, (address & ~0x3), &w);
+    read_word(opaque, (address & ~0x3), &w);
     *value = w;
-
-    return result;
 }
 
-static int writew(writefn write_word, void* opaque, uint32_t address, uint32_t value, uint32_t wmask)
+static void writew(writefn write_word, void* opaque, uint32_t address, uint32_t value, uint32_t wmask)
 {
-    return write_word(opaque, (address & ~0x3), value, wmask);
+    write_word(opaque, (address & ~0x3), value, wmask);
 }
 
 
@@ -293,7 +291,7 @@ static void writemem_with_bp_checks(void)
     check_breakpoints_on_mem_access(*r4300_pc(&g_dev.r4300)-0x4, *r4300_address(&g_dev.r4300), 4,
             M64P_BKP_FLAG_ENABLED | M64P_BKP_FLAG_WRITE);
 
-    return g_dev.mem.saved_writemem[*r4300_address(&g_dev.r4300)>>16]();
+    g_dev.mem.saved_writemem[*r4300_address(&g_dev.r4300)>>16]();
 }
 
 void activate_memory_break_read(struct memory* mem, uint32_t address)
--- a/src/device/memory/memory.c
+++ b/src/device/memory/memory.c
@@ -58,30 +58,11 @@ static int readw(readfn read_word, void* opaque, uint32_t address, uint64_t* val
     return result;
 }
 
-static int readd(readfn read_word, void* opaque, uint32_t address, uint64_t* value)
-{
-    uint32_t w[2];
-    int result =
-    read_word(opaque, address    , &w[0]);
-    read_word(opaque, address + 4, &w[1]);
-    *value = ((uint64_t)w[0] << 32) | w[1];
-
-    return result;
-}
-
 static int writew(writefn write_word, void* opaque, uint32_t address, uint32_t value, uint32_t wmask)
 {
     return write_word(opaque, (address & ~0x3), value, wmask);
 }
 
-static int writed(writefn write_word, void* opaque, uint32_t address, uint64_t value)
-{
-    int result =
-    write_word(opaque, address    , (uint32_t) (value >> 32), ~0U);
-    write_word(opaque, address + 4, (uint32_t) (value      ), ~0U);
-
-    return result;
-}
 
 
 static void read_nothing(void)
@@ -90,17 +71,7 @@ static void read_nothing(void)
     *g_dev.r4300.rdword = (*g_dev.r4300.rdword << 16) | *g_dev.r4300.rdword;
 }
 
-static void read_nothingd(void)
-{
-    *g_dev.r4300.rdword = *r4300_address(&g_dev.r4300) & 0xFFFF;
-    *g_dev.r4300.rdword = (*g_dev.r4300.rdword << 16) | *g_dev.r4300.rdword;
-}
-
 static void write_nothing(void)
-{
-}
-
-static void write_nothingd(void)
 {
 }
 
@@ -113,15 +84,6 @@ static void read_nomem(void)
     read_word_in_memory();
 }
 
-static void read_nomemd(void)
-{
-    struct r4300_core* r4300 = &g_dev.r4300;
-
-    *r4300_address(r4300) = virtual_to_physical_address(r4300, *r4300_address(r4300),0);
-    if (*r4300_address(r4300) == 0x00000000) return;
-    read_dword_in_memory();
-}
-
 static void write_nomem(void)
 {
     struct r4300_core* r4300 = &g_dev.r4300;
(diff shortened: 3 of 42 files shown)


loganmc10:
I tested a number of games on x86 new dynarec and x64 dynarec. No issues, it all seems good. Didn't test ARM

loganmc10:
Here is a kind of interesting one though. I was testing International Track & Field 2000, which doesn't currently boot (https://github.com/mupen64plus/mupen64plus-core/issues/311)

Before this PR, it just sat in an endless loop. But with this PR, using the x86 new dynarec, it actually segfaults.

```
Thread 1 "mupen64plus" received signal SIGSEGV, Segmentation fault.
MTC0.lto_priv.339 () at ../../src/device/r4300/mips_instructions.def:1287
1287	        cp0_update_count(r4300);
(gdb) backtrace
#0  MTC0.lto_priv.339 () at ../../src/device/r4300/mips_instructions.def:1287
#1  0xe6445c11 in ?? ()
Backtrace stopped: Cannot access memory at address 0x3
```

Doesn't happen with the Cached Interpreter or the x64 dynarec. I'm not saying there is something wrong with the PR, since this game never worked before, I guess it's just "exposing the bad behavior" in a different way.

loganmc10:
Compiling this for Android I get:
```
/home/travis/android-ndk-r14b/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: /home/travis/mupen64plus-ae/app/build/intermediates/ndkBuild/release/obj/local/armeabi-v7a/objs/mupen64plus-core/./mupen64plus-core/src/device/r4300/new_dynarec/arm/linkage_arm.o: requires unsupported dynamic reloc R_ARM_MOVW_ABS_NC; recompile with -fPIC
/home/travis/android-ndk-r14b/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: warning: shared library text segment is not shareable
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

I'm compiling with fPIC now, I'll post my results in a bit. I assume the Raspberry Pi will have a similar issue, I'll test tonight, we may have to update the Unix makefile to add -fPIC for arm

loganmc10:
I still get that error even after doing this: https://github.com/loganmc10/mupen64plus-ae/commit/dc3fb9e3f0196c92edbf497f2db97f8982d954c8

I'm not really an Android pro, perhaps @fzurita can try getting it to compile. I'll try on my Raspberry Pi tonight

fzurita:
Hmmm, sounds like we are doing something that is not PIC complaint. I'm usually not good at deciphering those thing but @Gillou68310 had in the past.

bsmiles32:
Updated. I changed the way dynarec_{read,write}_aligned_dword is called and hopefully this should fix the "PIC" issue.

Gillou68310:
The current implementation will break code invalidation and tlb exceptions with the new dynarec.

One solution I can think of is to call write_rdram_new/write_nomem_new twice within the write_dword_new function, updating address/wword/wmask between the 2 calls.

We still need to find a workaround for read_nomemd_new.

loganmc10:
I just tested this PR on Android with a few games and it seems to be working, no more PIC errors and no crashes.

I assume the issues that Gilles is talking about relate to #313, which still doesn't work on ARM or x86 new dynarec.

bsmiles32:
Tried to solve merge conflicts.
@Gillou68310 I don't get why this PR would break code invalidation nor TLB exceptions. Maybe you are referring to my other PR313 (which is based on this one, and goes a bit further with the refactorings) ?

dankcushions:
build error on RPI3/ARM:
```
In file included from ../../src/device/r4300/new_dynarec/new_dynarec.c:1678:0:
../../src/device/r4300/new_dynarec/new_dynarec.c: At top level:
../../src/device/r4300/new_dynarec/arm/assem_arm.c:1011:13: warning: ‘regname’ defined but not used [-Wunused-variable]
 static char regname[16][4] = {
             ^
    CC  _obj/main/zip/unzip.o
    CC  _obj/asm_defines/asm_defines.o
    CXX _obj/osd/screenshot.o
awk -v dest_dir="../../src/asm_defines" -f ../../tools/gen_asm_defines.awk "_obj/asm_defines/asm_defines.o"
    CC  _obj/device/r4300/new_dynarec/arm/linkage_arm.o
../../src/device/r4300/new_dynarec/arm/linkage_arm.S:69:29: fatal error: asm_defines_gas.h: No such file or directory
 #include "asm_defines_gas.h"
                             ^
compilation terminated.
```

Gillou68310:
image

This it what the history looks like on this branch. What a mess!!!
Could you please revert your last 2 commits and rebase against master?

Sorry if I sound like a psycho ;-)

Gillou68310:
You cannot call a C functions because you need to keep control over the stack address in case of a tlb exception.
When a tlb exception occurs the stack address is restored with the value of the stack address right before the ```read/write_byte/hword/word/dword_new``` call.
Code invalidation is not impacted this was a mistake.

loganmc10:
I tested a bunch of games on the x86 new dynarec and they all seemed to work properly with this PR, but it does segfault on the next PR with the new dynarec

bsmiles32:
@Gillou68310 Thank you for your valuable input. I will rebase properly against master and rework it as to not break byte and hword accesses.

Gillou68310:
@bsmiles32 don't spend to much time on this because I'm planning to rework the tlb exception part for the new dynarec.

bsmiles32:
I rebased against master, and removed forcing the word access alignment up until readw/writew functions so that eventual ERROREPC will get the correct address. Please test and double check everything so we can move on with this one :)

loganmc10:
I tested about a dozen games using the x86 new dynarec, no issues that I can see

bsmiles32:
@Gillou68310 Does it looks OK for you ? I'd like to get that merged in the coming days/weeks so I can work on the next memory refactoring PR. Thanks

dankcushions:
builds and works on ARM/RPI3 :)

fzurita:
Any good speedups on ARM?

bsmiles32:
Thanks for testing 😀

dankcushions:
@fzurita not that I could notice. but i think it's always GPU that's a bottleneck on the RPI3.

ricrpi:
I tried timing the call to eglSwapBuffers() to see if the implicit glFlush() blocked. However on a Pi B I never saw it wait for the GPU. Although I guess with the more modern ARM cores, the CPU could get to the next VI before GPU has finished.

dankcushions:
Interesting! My go-to bottleneck area is mario 64 - the start of bobomb battlefield (immediately after the text) gets choppy with distorted sound. CPU usage isn't that high so I presume it's GPU, but maybe it's memory or something else. Other than that it seems locked at 30 for everything previous. I'd love to figure out for certain.

That's with a Pi 3, though. It never seems to max out a core in mupen, so I guess it has enough CPU grunt for the dynarec.

bsmiles32:
@fzurita provided that it works, you might have some more performance in the next PR (memory_refactorings).

fzurita:
Ok, I think I was confusing this PR with that one.

ricrpi:
Mupen64plus is mainly single threaded so maybe a PI 3 can't max out a quad core CPU.

bsmiles32:
@Gillou68310 : Is it okay if I merge this ?

Gillou68310:
> You cannot call a C functions because you need to keep control over the stack address in case of a tlb exception.
> When a tlb exception occurs the stack address is restored with the value of the stack address right before the read/write_byte/hword/word/dword_new call.

This is still true, the only reason why it's not crashing is probably because read/write dword never trigger a tlb exception. I started reworking the tlb exception code https://github.com/Gillou68310/mupen64plus-core/commit/feb52e7ff36909de97976a8f930a49578c96bc32 but I don't have much time to work on it these days so progress is really slow, sorry about that.

The main idea is to remove all read/write assembly routine and use C code instead. This should make refactoring easier in the future ;-)

loganmc10:
@bsmiles32 if it was me I would just merge this. I don't think it's right to let the new dynarec to hold up the emulator's progress (this is coming from a guy that uses Android for mupen64plus by the way).

The x86 new dynarec is already currently broken on Android. I think all you can do is do your best to make it work with the new dynarec, and just merge it. If someone can keep up with the new dynarec then awesome, if not then it'll have to "fall by the wayside"

This isn't meant to be a slight on anyone by the way. People are busy and we unfortunately don't have a lot of people knowledgeable with ARM assembly code to maintain it properly, especially since it reimplements things like the TLB which make it even more complicated

loganmc10:
I wasn't around when the new dynarec was developed so this might make me sound like a complete idiot, but I'm not sure why the "old dynarec" wasn't just ported to arm. Even a device like the Raspberry Pi is only usually around 70% CPU utilization with the new dynarec, so even if the older one was a bit slower, it would probably still work well.

If I ran the zoo, I would just port the old dynarec to arm and blow away the new dynarec, it would make the code 1000% more maintainable. I'd even be willing to learn arm assembly and fumble through that task if I knew that meant we could get rid of the new dynarec

bsmiles32:
@Gillou68310 Thanks for answering. I'll see if I can update the PR to address your concern, or wait after you're done with the TLB refactorings. I'm also not able to work much on this currently (and maybe for several more weeks).
@loganmc10 new dynarec is indeed difficult to maintain and slows refactorings, but unless we can replace it with something equivalent, I don't think removing it is the right way to go.
I have plans for unifying all the r4300 cores (pure_interpreter, cached_interpreter, and both dynarecs), but I wanted to finish all the memory refactorings first (eg PR 312, 313 and the fast_mem issue 277 for which I have a basic PoC locally, but I'm not getting the performance improvement I seek). The idea to unify them is to leverage LLVM infrastructure : describe mips instructions using LLVM IR, compile them so they can be called by the pure_interpreter or the cached_interpreter, or JIT them with additional optimization passes for the dynarec. This will effectively unify all r4300 cores and provide support for x86, x64, ARM and MIPS :) But that is for later, and maybe I should create an issue to open the discussion :).

bsmiles32:
Rebased against master, and tentative fix for ARM dynarec.
@Gillou68310 can you check if my last commit addresses your concerns please ?
I use the very same technique to call C functions as when it used to call {read,write}_memd[address >> 16], but now I call dynarec_{read,write}_aligned_dword.

Gillou68310:
@bsmiles32 I finally took some time to finish reworking the new dynarec tlb exception code https://github.com/Gillou68310/mupen64plus-core/tree/tlb_exception_rework

Once this is merged you should be able to integrate the remove_memd branch without too much trouble ;-)

@loganmc10 @fzurita This should be intensively tested (especially on ARM where I did almost no testing) before being merged!

Also I'm a bit concerned about the performance impact, could you guys try to make some measurements?

@fzurita This also fix https://github.com/mupen64plus/mupen64plus-core/issues/426

Gillou68310:
@bsmiles32 I finally took some time to finish reworking the new dynarec tlb exception code https://github.com/Gillou68310/mupen64plus-core/tree/tlb_exception_rework

Once this is merged you should be able to integrate the remove_memd branch without too much trouble ;-)

@loganmc10 @fzurita This should be intensively tested (especially on ARM where I did almost no testing) before being merged!

Also I'm a bit concerned about the performance impact, could you guys try to make some measurements?

@fzurita This also fix #426

fzurita:
Thanks, I'll try this as soon as I can.

fzurita:
It seems to be working ok in ARM. I'm trying CBFD. Performance seems identical from the FPS counter.

bsmiles32:
@Gillou68310 Thanks for working on this ! I'll wait for this to be merged and rebase this and hopefully the other memory PR after.

fzurita:
@Gillou68310 You should merge your tlb_exception_rework soon, since it fixes the x86 new dynarec freezing.

Gillou68310:
Ok I'll merge it then.

Gillou68310:
Done

bsmiles32:
Rebased against master. @Gillou68310 thank you very much for making progress on this, it indeed greatly simplify my work on the memory subsystem.

loganmc10:
Seems to work well on Linux (x64), no issues

bsmiles32:
A superset of this PR is available here : https://github.com/mupen64plus/mupen64plus-core/pull/313
Depending on how much work we want to merge, it might be preferable to directly merge the PR 313.

loganmc10:
I know the last time I tested all this, this PR worked with the new dynarec but 313 crashed. It might be different now after Gillou68310's work though. I'd probably merge this and then we can focus on doing a bit more testing on 313, including making sure the memory allocation isn't an issue.

fzurita:
Maybe known regressions should be taken care of before merging this.

See https://github.com/mupen64plus/mupen64plus-core/issues/385

bsmiles32:
@fzurita After @Gillou68310's work the only time I touched the new_dynarec in this PR (and PR 313) is for :
- extra spaces in linkage_arm.S and linkage_x86.S
- removing unneeded protypes in new_dynarec.c
- and replacing usages of {read,write}_{d,}word_in_memory in new_dynarec.c
So I don't think this PR has high chances of breaking more the new_dynarec than what it is now. Still I can understand why we might want delay this work a little bit.
But please, don't delay this work too much, because it is quite massive therefore resolving conflicts is always tedious and I've been trying to do this refactoring for almost 3 years now (I started with the now defunct PR https://github.com/mupen64plus/mupen64plus-core/pull/47)... So I will be very happy when all these memory refactorings will be merged :)

fzurita:
@bsmiles32 Loading save states is causing crashes after this last set of commits. At least with the arm new dynarec. I haven't tried anything else.

bsmiles32:
@fzurita Can you get backtrace of the crash ? Or bisect more precisely (may or may not be possible) ?

fzurita:
Yeah, I will try. I didn't have much time last night.

fzurita:
Ok, here is a backtrace:
```
********** Crash dump: **********
Build fingerprint: 'motorola/quark_verizon/quark:6.0.1/MCG24.251-5/9:user/release-keys'
pid: 8273, tid: 8353, name: ServiceStartArg  >>> paulscode.android.mupen64plusae.GameActivity <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
Stack frame #00 pc 0002911c  libmupen64plus-core.so: Routine set_rumble_reg at ./mupen64plus-core/src/device/controllers/paks/rumblepak.c:32
Stack frame #01 pc 00016a50  libmupen64plus-core.so: Routine savestates_load_m64p at ./mupen64plus-core/src/main/savestates.c:643
Stack frame #02 pc 00017210  libmupen64plus-core.so: Routine savestates_load at ./mupen64plus-core/src/main/savestates.c:1152
Stack frame #03 pc 000b45ac  libmupen64plus-core.so: Routine .E1 at linkage_arm.S:?
```

loganmc10:
I also just got a report about crashing save states, this is on Windows:

https://github.com/m64p/mupen64plus-GLideN64/issues/22

bsmiles32:
Can you test https://github.com/mupen64plus/mupen64plus-core/pull/439 ?

Also it was not related to this PR but on the one about SI refactorings.
Hope it fixes this issue as I will be away for a couple of days...
Sorry for the inconvenience.

PR #313, opened 2017-06-15
Memory refactorings

bsmiles32:
DON'T MERGE !
This is a WIP, with several refactorings concerning mainly the memory module :
- now the memory module only deals with "physical" mappings (virtual to physical translation is done in r4300)
- no more usage of global variables in this module
- single allocation of memory for all device memories. I allocate 512M, but only a fraction gets used (eg RAM, ROM, PIF_RAM, SP_MEM). This simplify the fast_mem_access function which gives a nice speed boost on the pure_interpreter. This also paves the way for the "fast memory" implementation.
- memory mappings are defined in the device module
- pifbootrom has been reworked to not use the device structure, but only r4300_core struct. This also paves the way for external PIFBootROM support. Not there yet, but closer.

edit: ~~It also breaks the cached_interpreter, don't know why. And~~ I didn't really bother with the new_dynarec so it should break it quite badly.

When I get some time, I will try to rework it and put it into shape. But early testers are welcome :)
Especially, I'd like to know if it improves the emulation speed on other computers without bad regressions.

### DIFF
--- a/src/api/frontend.c
+++ b/src/api/frontend.c
@@ -92,9 +92,9 @@ EXPORT m64p_error CALL CoreStartup(int APIVersion, const char *ConfigPath, const
     if (!main_set_core_defaults())
         return M64ERR_INTERNAL;
 
-    /* allocate memory for rdram */
-    g_rdram = malloc(RDRAM_MAX_SIZE);
-    if (g_rdram == NULL) {
+    /* allocate base memory */
+    g_mem_base = malloc(0x20000000);
+    if (g_mem_base == NULL) {
         return M64ERR_NO_MEMORY;
     }
 
@@ -121,9 +121,9 @@ EXPORT m64p_error CALL CoreShutdown(void)
     /* tell SDL to shut down */
     SDL_Quit();
 
-    /* deallocate RDRAM */
-    free(g_rdram);
-    g_rdram = NULL;
+    /* deallocate base memory */
+    free(g_mem_base);
+    g_mem_base = NULL;
 
     l_CoreInit = 0;
     return M64ERR_SUCCESS;
--- a/src/device/ai/ai_controller.c
+++ b/src/device/ai/ai_controller.c
@@ -195,7 +195,12 @@ void write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask
     {
     case AI_LEN_REG:
         masked_write(&ai->regs[AI_LEN_REG], value, mask);
-        fifo_push(ai);
+        if (ai->regs[AI_LEN_REG] != 0) {
+            fifo_push(ai);
+        }
+        else {
+            /* stop sound */
+        }
         return;
 
     case AI_STATUS_REG:
--- a/src/device/ai/ai_controller.c
+++ b/src/device/ai/ai_controller.c
@@ -164,7 +164,7 @@ void poweron_ai(struct ai_controller* ai)
     ai->delayed_carry = 0;
 }
 
-int read_ai_regs(void* opaque, uint32_t address, uint32_t* value)
+void read_ai_regs(void* opaque, uint32_t address, uint32_t* value)
 {
     struct ai_controller* ai = (struct ai_controller*)opaque;
     uint32_t reg = ai_reg(address);
@@ -184,11 +184,9 @@ int read_ai_regs(void* opaque, uint32_t address, uint32_t* value)
     {
         *value = ai->regs[reg];
     }
-
-    return 0;
 }
 
-int write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
+void write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
 {
     struct ai_controller* ai = (struct ai_controller*)opaque;
     uint32_t reg = ai_reg(address);
@@ -198,11 +196,11 @@ int write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
     case AI_LEN_REG:
         masked_write(&ai->regs[AI_LEN_REG], value, mask);
         fifo_push(ai);
-        return 0;
+        return;
 
     case AI_STATUS_REG:
         clear_rcp_interrupt(ai->r4300, MI_INTR_AI);
-        return 0;
+        return;
 
     case AI_BITRATE_REG:
     case AI_DACRATE_REG:
@@ -211,12 +209,10 @@ int write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
             ai->samples_format_changed = 1;
 
         masked_write(&ai->regs[reg], value, mask);
-        return 0;
+        return;
     }
 
     masked_write(&ai->regs[reg], value, mask);
-
-    return 0;
 }
 
 void ai_end_of_dma_event(void* opaque)
--- a/src/device/ai/ai_controller.h
+++ b/src/device/ai/ai_controller.h
@@ -77,8 +77,8 @@ void init_ai(struct ai_controller* ai,
 
 void poweron_ai(struct ai_controller* ai);
 
-int read_ai_regs(void* opaque, uint32_t address, uint32_t* value);
-int write_ai_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask);
+void read_ai_regs(void* opaque, uint32_t address, uint32_t* value);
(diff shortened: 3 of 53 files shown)


loganmc10:
I tried Super Mario and Smash Bros, neither booted with x64 dynarec or either interpreter, it didn't crash, but CPU was at 100%, here is what GDB had if I Ctrl+C'ed out of the program:

```
(gdb) backtrace
#0  SPECIAL (PC=3915365300, inst=<optimized out>) at ../..//su.c:1702
#1  run_task () at ../..//su.c:2006
#2  0x00007fffccc1e7c9 in DoRspCycles (cycles=4294967295) at ../..//module.c:375
#3  0x00007ffff05edead in do_SP_Task (sp=0x7ffff1939718 <g_dev+17834392>) at ../../src/device/rsp/rsp_core.c:297
#4  0x00007ffff05eb583 in r4300_write_aligned_word (r4300=0x7ffff0837580 <g_dev>, address=67371024, value=1, mask=<optimized out>) at ../../src/device/r4300/r4300_core.c:393
#5  0x00007ffff05d0eee in pifbootrom_hle_execute (r4300=0x7ffff0837580 <g_dev>) at ../../src/device/pifbootrom/pifbootrom.c:74
#6  0x00007ffff05f5e64 in main_run () at ../../src/main/main.c:1120
#7  0x00007ffff05d834d in CoreDoCommand (Command=<optimized out>, ParamInt=<optimized out>, ParamPtr=<optimized out>, ParamPtr=<optimized out>, ParamInt=<optimized out>, 
    Command=<optimized out>) at ../../src/api/frontend.c:225
#8  0x0000000000402e7f in main (argc=<optimized out>, argv=<optimized out>) at ../../src/main.c:798
```

I tested this using rsp-cxd4 in HLE video mode

loganmc10:
Also, malloc'ing 512MB of RAM will be pretty tight on some ARM devices. A lot of Android devices only have 2GB of RAM, my Shield Tablet has 2GB and it says only 487 is free right now (although I assume it would just close some old apps if an app requested memory, so it's probably not a problem). The Raspberry Pi also has 1GB of RAM.

Is there some way to have a "fallback mode"? Either if the malloc fails or if the user requests it. I'm really excited for fast memory access so I think it should go forward, but it might mean that performance actually suffers on low memory devices

bsmiles32:
Normally the kernel should lazy alloc it, so only pages that are really accessed are allocated. I guess it should be no problem.

bsmiles32:
Several fixes which re-enable the cached interpreter and improve pifbootrom hle implementation.
As far as I understand, only new dynarec needs fixing and that is mostly at commit 424938e.

loganmc10:
Crashing is fixed for me, tested several games on x64 dynarec and Cached/Pure Interpreter, seems to be working well now

loganmc10:
Just tried compiling x86 new dynarec, I get this error on linking:

```
_obj/device/r4300/new_dynarec/x86/linkage_x86.o:../../src/device/r:(.text+0x5c8): undefined reference to `write_mi'
```

I'm not sure if the error is a result of a commit from this PR or the previous one

bsmiles32:
New_dynarec requires some more work unfortunately. I will need help from Gillou on this, because I'm not familiar with how the new_dynarec does code invalidation, and TLB translation.
Still if you manage to hack around and get the pure or cached interpreters to run on your rpi I'm interested in your results 😀

loganmc10:
I've tested the Pure Interpreter inside my VMware machine. In my testing this PR saves about 2% CPU time in Pure Interpreter mode on average. Sometimes it was pretty much the same, I would just run the same scene in both versions and compare what I saw in "top", so the testing wasn't anything too scientific. I don't see any regressions anyway.

bsmiles32:
Good ! Thanks for testing ! 2% now, hope for some more with full fast mem 😀

bsmiles32:
This also validate the big allocation approach. So this is good.

loganmc10:
My dream would be to get to the point where the cached interpreter would be fast enough on mobile devices to get rid of the dynarec's altogether (but this is probably an unrealistic dream). I think having an arm dynarec without anyone with enough time or knowledge to maintain it is really going to hold things back.

bsmiles32:
I feel the same way ! Only one person knows how to work with the new dynarec... And its not me. So all my refactorings take much more time than needed. I have hope (and maybe a long term plan) to unify the pure,cached and both dynarec. But this will take more time than I have now.

bsmiles32:
Tried something to fix the new dynarec. Don't know if it will work, but it's a step closer I guess.

ricrpi:
What if only 128MB + ROM size is used and a catch for PIF access?

Also how does Linux handle lazy allocation if MAP_LOCKED is used with mmap() to try and reduce page faults?

bsmiles32:
Not really sure, but from what I undestand it allocates 512MB of virtual memory, then when accesses are made physical pages are allocated when needed. Of all of these 512MB only ram_size+rom_size+rsp_mem_size+pif_boot_rom_size+pif_ram_size+rounding to next physical page size will get physically allocated (approximately).
Note that for now no pagefault mecanism is in use (this is not yet the fast mem approach). This just make all N64 memories accessible through à single pointer, which gives a small speed boost on fast_mem_address due to the removed if/else logic at the cost of a small increase of memory usage.
The next step is indeed to change memory protection of region with mmio semantic in order to trigger a pagefault which will allow to call the appropriate mmio handler. But that's for later.
Still I'm interested to know if this big allocation works reliably on 32bit systems such as the rpi.

dankcushions:
build error with arm + RPI3:
```
   CC  _obj/asm_defines/asm_defines.o
    CXX _obj/osd/screenshot.o
nm _obj/asm_defines/asm_defines.o | awk -v dest_dir="../../src/asm_defines" -f ../../tools/gen_asm_defines.awk
    CC  _obj/device/r4300/new_dynarec/arm/linkage_arm.o
    LD  libmupen64plus.so.2.0.0
/usr/bin/ld: _obj/device/r4300/new_dynarec/arm/linkage_arm.o: relocation R_ARM_MOVW_ABS_NC against `dynarec_read_aligned_dword' can not be used when making a shared object; recompile with -fPIC
_obj/device/r4300/new_dynarec/arm/linkage_arm.o: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
Makefile:706: recipe for target 'libmupen64plus.so.2.0.0' failed
make: *** [libmupen64plus.so.2.0.0] Error 1
```

this might be a compilation options issue with -fPIC, but by default I get this.

loganmc10:
I briefly tested a few games with the x86 new dynarec and it seems to be working properly.

bsmiles32:
Updated with 2 fixes :
* one from remove_memd PR concerning the way I called dynarec_{read,write}_aligned_dword
* one for the missing write_mi function

Thank you guys for testing  :)

dankcushions:
builds! but segfaults for me. unfortunately i think the backtrace is not useful:
```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/arm-linux-gnueabihf/libthread_db.so.1".
[New Thread 0x478aa460 (LWP 15533)]
[New Thread 0x46ef3460 (LWP 15534)]
[New Thread 0x464ff460 (LWP 15535)]
[New Thread 0x45cff460 (LWP 15536)]
[New Thread 0x454ff460 (LWP 15537)]
[New Thread 0x44cff460 (LWP 15538)]
[New Thread 0x444ff460 (LWP 15539)]

Program received signal SIGSEGV, Segmentation fault.
0x61746144 in ?? ()
(gdb) bt
#0  0x61746144 in ?? ()
#1  0x67ade5e0 in ?? () from /opt/retropie/emulators/mupen64plus/lib/libmupen64plus.so.2
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```
went back to master and everything worked ok via gdb. that's all i've got, sorry :(

bsmiles32:
@dankcushions Thanks for testing.
I guess, this PR will be on hold until we can merge PR312 "remove_memd".

bsmiles32:
Rebased against master, includes the changes from PR remove_memd, cleaned up commits.

bsmiles32:
Rebased against master.

Finally, this PR should be ready for merging.

Gillou68310:
@bsmiles32 something in this PR broke the reset functionality

Gillou68310:
It was introduced by 1c01c2339bafb503e46cf776d995e51ca46c73b4

PR #317, opened 2017-06-21
Allow using compiler other than gcc

loganmc10:


bsmiles32:
OK, reverted all commits. Thank you for reporting this issue.
Really sorry about this. 3rd time in few days that I break master... :(

loganmc10:
No that was my bad, I didn't realize that make sets the CC variable

fzurita:
You could use a different environment variable other than CC

loganmc10:
It seems that you can still override CC by doing ```make CC=clsng all``` so that should suffice

PR #322, opened 2017-06-26
Fix asm_defines w/ mawk by setting LANG=C

comex:
Otherwise it silently exits with code 127, presumably because the input (binary) is not valid text in the system encoding.


charlemagnelasse:
Btw. the return code of the mawk call is 0 in my case.

comex:
Bleh, I thought I tested this but I must have messed something up, given

    BUGS
           mawk  cannot  handle ascii NUL \0 in the source or data files.  You can output NUL using printf
           with %c, and any other 8 bit character is acceptable input.

But I don't want to depend on `gawk` on platforms that might not have it, e.g. macOS.  I just updated the PR to instead pipe `strings` to `awk`.  Windows can stay the way it is: there, `strings` may not exist, but `awk` is gawk, directly from `mupen64plus-win32-deps`.  On Unix, `strings` should definitely exist with the same cross-compile prefix as `gcc`, so we can use that always.

loganmc10:
I ran into this same problem when testing Ubuntu 17.04, which uses mawk by default. This PR (using ```strings```) did indeed fix the problem.

PR #326, opened 2017-06-27
Updates to DisableExtraMem

loganmc10:
This PR does a number of things:

It moves the initialization of (g_)rdram_size from CoreStartup to main_run, with the rest of the config variables. @bsmiles32 I didn't see any need for this to be a global, am I correct in my understanding?

It allows DisableExtraMem to be set in mupen64plus.ini, and enables it for Iggy's Reckin' Balls, where it is required.

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -5914,6 +5914,7 @@ GoodName=Iggy's Reckin' Balls (E) [!]
 CRC=D692CC5E EC58D072
 Players=2
 SaveType=Controller Pack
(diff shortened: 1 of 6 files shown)


bsmiles32:
It was a global just to keep it close to g_rdram which is required to be a global. I'm okay with your changes, and anyway when we get to merge my other work about memory, g_rdram will be renammed to g_mem_base and g_rdram_size would also have been (re)moved like you did. So good work ! Will merge it later if other want to test it a bit more.

PR #327, opened 2017-06-28
Implement "Audio Signal"

loganmc10:
PR https://github.com/mupen64plus/mupen64plus-core/pull/307 actually broke a few games (NBA on NBC, Hydro Thunder).

Project64 has a setting/hack called "RSP Audio Signal" https://github.com/project64/project64/blob/master/Source/Project64-core/N64System/Mips/MemoryVirtualMem.cpp#L1148-L1152

This fixes games like Hydro Thunder and NBA on NBC. I tested it and it works here as well. It is enabled for: NBA on NBC, Hydro Thunder, Tarzan, and Rugrats in Paris (these games were taken from the Project64 RDB)

Of course we could also revert #307, but I think that was a positive change, it simplified the code and allowed us to remove the strange Perfect Dark hack that was in place, it also fixed freezing in Top Gear Rally 2.

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -3348,6 +3348,7 @@ CRC=D614E5BF A76DBCC1
 Players=1
 SaveType=Controller Pack
 CountPerOp=1
+AudioSignal=1
 
 [AF740B224E5DD0BD09F7254811559ADF]
 GoodName=Disney's Tarzan (E) [h1C]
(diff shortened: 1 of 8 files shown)


fzurita:
Because mupen64plus is not cycle accurate,  I think some hacks are acceptable to get all games running. I can't see any other way... I guess you have to pick which hacks are better than others.

loganmc10:
Yeah it does seem to be some kind of timing issue, running these games with RSP-CXD4 and ```SupportCPUSemaphoreLock``` enabled will also fix this problem. This approach is sort of the "HLE" approach that allows them to work with rsp-hle as well.

bsmiles32:
Tested and indeed fixes these games. Merging this but hope that a less hackish way of fixing this will be found :)

PR #330, opened 2017-07-04
Add hack for Rat Attack!

loganmc10:
Yes I know... another game specific hack. Project64 employs a similar hack (they disable TLB for the game).

I also took the opportunity to remove the global variable, and I used the same variable for the GoldenEye, and now Rat Attack, hack.

I'm open to suggestions on variables names, I was never very good at naming things...

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -11701,6 +11701,7 @@ CRC=20FD0BF1 F5CF1D87
 Players=4
 Rumble=Yes
 SaveType=Controller Pack
+CountPerOp=1
 
 [E0BB65C30C1185FD9997020A1994B07E]
 GoodName=Rat Attack (E) (M6) [f1] (NTSC)
--- a/src/device/device.c
+++ b/src/device/device.c
@@ -37,6 +37,7 @@ void init_device(struct device* dev,
     unsigned int emumode,
     unsigned int count_per_op,
     int no_compiled_jump,
+    int special_rom,
     /* ai */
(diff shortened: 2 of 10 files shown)


loganmc10:
I never tested this with the new dynarec, I don't know if it would need a different solution

loganmc10:
Got it working with the new dynarec

bsmiles32:
On my end, it indeed allow the game to boot, but I can't get in-game with the European version as I get a :
Core Error: NI() @ 0x800702C0
Core Error: opcode not implemented: 800702C0:D
The USA version seems OK, still in-game there seems to be some missing objects but I don't know if it's a core issue or a gfx issue.
Tested with Pure Interpreter, and GlideN64 plugin.

loganmc10:
The gfx issues are from GLideN64, it works with Glide64.

You can hold off on this for a bit, I'll see if I can track down a better fix, or at least one that also works for the E version

loganmc10:
CountPerOp=1 fixes the European version for me, I've updated the PR. Interestingly, CountPerOp=1 breaks the USA version in the same way the European version was broken. Another unfortunate timing situation it seems.

I looked at 1964 to see if they had fixed it, but they have TLB disabled for this game as well. The European version also crashes in PJ64, so it's a pretty common issue it seems.

bsmiles32:
Merging this, but I don't like the special rom approach much. It is yet another way to enable hacks. Hope we'll find a more generic solution to this issue.

Issue #331, opened 2017-07-04
Indiana Jones freeze at end of level


loganmc10:
Firstly, this game is very unstable unless CountPerOp=1 is set. It seems like there were some issues in the past with this, but I played through the first level twice with CountPerOp=1 and it worked well. With CountPerOp=2, it sometimes freezes right after you get control of Indy. There is also flickering that happens on CountPerOp=2.

The big issue though: At the end of the level, it goes to a black screen and freezes there. I tested with the x84 dynarec and Cached Interpreter.

Project64 has the same issue, but only with their CPU Re-compiler (not their Interpreter) https://github.com/project64/project64/issues/259.

Here is a save state during the ending cutscene of the first level if anyone wants to take a look: I would use the Cached Interpreter and set CountPerOp=1 if you're going to try it out

[Indiana Jones and the Infernal Machine (U) [!].zip](https://github.com/mupen64plus/mupen64plus-core/files/1122692/Indiana.Jones.and.the.Infernal.Machine.U.zip)

loganmc10:
What a whacky game. If I load that save state with CountPerOp=2, then I can get to the Trading Post/end of level. But if I play the whole level through with CountPerOp=2, it doesn't load.

This game might be hard to emulate without cycle accuracy, there may be some magical combination of CountPerOp/CountPerScanline that works though

AmbientMalice:
Also, the jeep physics are wrong without CF=1, and there are issue with the inflatable life-raft.

loganmc10:
I'm going to revisit this if HLE graphics are ever accomplished for this game, see comment in #332

PR #332, opened 2017-07-05
Fix Indiana Jones

loganmc10:
This PR is dependent on https://github.com/mupen64plus/mupen64plus-core/pull/330. Fixes https://github.com/mupen64plus/mupen64plus-core/issues/331

I would like someone to test this and review before it's merged, I don't mind if it sits here for a while.

This does a number of things:
It sets CountPerOp=1 and CountPerScanline=2200. This fixes random freezing and some graphical glitches

It also introduces ```pi->interrupt_mod```, basically, for Indiana Jones, it divides the pi interrupt timers by 2. This fixes the black screen/hanging at the end of the levels. I think this is acceptable, since the current value (0x1000 usually), seems like a guess/approximation.

I played through the first 2 levels using the x64 dynarec without issue. But like I said, I would like someone else to test/comment before it's merged. This was mostly the result of trial/error.

I tested the U rom, I did not test the PAL rom

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -5962,12 +5962,16 @@ GoodName=Indiana Jones and the Infernal Machine (U) [!]
 CRC=AF9DCC15 1A723D88
 Players=1
 SaveType=Eeprom 4KB
+CountPerOp=1
+CountPerScanline=2200
 
 [6F417D30D772F4420C9384E9BBB7BC01]
 GoodName=Indiana Jones and the Infernal Machine (E)
 CRC=3A6F8C6B 2897BAEB
 SaveType=Eeprom 4KB
 Players=1
+CountPerOp=1
+CountPerScanline=2200
 
 [A7781D441AF55C4FF8AFC68AB3A59313]
 GoodName=Indy Racing 2000 (U) [!]
--- a/src/device/device.c
+++ b/src/device/device.c
@@ -37,6 +37,7 @@ void init_device(struct device* dev,
     unsigned int emumode,
     unsigned int count_per_op,
     int no_compiled_jump,
+    int special_rom,
     /* ai */
     struct audio_out_backend* aout, unsigned int fixed_audio_pos,
     /* pi */
@@ -72,7 +73,7 @@ void init_device(struct device* dev,
     };
 
     init_r4300(&dev->r4300, &dev->mem, &dev->ri, interrupt_handlers,
-            emumode, count_per_op, no_compiled_jump);
+            emumode, count_per_op, no_compiled_jump, special_rom);
     init_rdp(&dev->dp, &dev->r4300, &dev->sp, &dev->ri);
     init_rsp(&dev->sp, &dev->r4300, &dev->dp, &dev->ri);
     init_ai(&dev->ai, &dev->r4300, &dev->ri, &dev->vi, aout, fixed_audio_pos);
--- a/src/device/device.h
+++ b/src/device/device.h
@@ -62,6 +62,7 @@ void init_device(struct device* dev,
     unsigned int emumode,
     unsigned int count_per_op,
     int no_compiled_jump,
+    int special_rom,
     /* ai */
     struct audio_out_backend* aout, unsigned int fixed_audio_pos,
     /* pi */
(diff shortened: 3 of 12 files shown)


AmbientMalice:
This is fantastic work, dude. I've been looking through the github over the past week, and I'm really surprised and pleased by how mupen64plus is coming along. Huge strides. I've become... increasingly displeased with PJ64, and it's great to see this emu stepping up to the plate.

AmbientMalice:
Would it be possible for you to upload the modified mupen64plus.dll? I'm having trouble building the emulator. It keeps complaining about "more than one input file specified" for mupen64plus-core.

loganmc10:
Yeah no problem, I can get it to you. It'll have to wait until Monday though when I have access to my Windows computer. Thanks a lot for testing.

fzurita:
This is pretty cool, now if we can get the new dynarec in ARM to stop crashing in this game.

loganmc10:
Ok @AmbientMalice here is a version to test. I can only build using MinGW, so it needs a few "support DLL's". I've attached the version I build at m64p.github.io with this Indiana Jones fix.

If you're going to test, I'd suggest opening mupen64plus-gui.exe, going to Settings->Plugin Settings, and click "Reset All Settings" at the bottom, just to make sure you're dealing with a fresh config.

[mupen64plus-Jones-Fix.zip](https://github.com/mupen64plus/mupen64plus-core/files/1135729/mupen64plus-Jones-Fix.zip)

AmbientMalice:
@loganmc10 Thanks a lot. I'm a huge admirer of this game, and it's great some effort is going into fixing it.

I'll give it a test and keep you posted. GLideN64 has a serious depth problem with this game, so I'll use z64gl from another build of mupen64plus. I'll just point out that this is possibly the most unstable "commercial" N64 game, notorious for stuff like the music suddenly looping rapidly and then the screen turning into "rainbow barf" before the console resets. Which means that random crashes don't necessarily mean the emulator is at fault, and makes troubleshooting a bit inherently unreliable.

loganmc10:
Yeah I quite enjoyed playing through the first 2 levels. It's got a pretty good story, especially compared to alot of games from that time.

That's a 64 bit build of mupen64plus so you'll need a 64 bit version of any plugin you use. I can provide a 32 bit version if you need.

AmbientMalice:
So I tried the build you posted, and I'm getting a freeze/crash at the end of the first level, per usual.

loganmc10:
Ok tomorrow I'll double check that I uploaded the right build. Can you try deleting any .eep/saves you may have? On Windows they should be in ```C:\you\AppData\Roaming\mupen64plus\save```

I'll try that build I posted tomorrow on my Windows PC to make sure, I'm really surprised it didn't work for you.

loganmc10:
A weird thing with this game, and probably a hint to what is wrong: If you delete you save (eep), and start the game with CountPerOp=2, and do the same thing with CountPerOp=1, the eep file it creates right when the game launches has a different md5sum. This game won't even boot with CountPerOp=1 (but it does create the eep file). However, CountPerOp=1 and CountPerScanline=2200 creates an eep file with a matching md5sum to the eep file created with CountPerOp=2. Basically it seems like the specific timing settings actually modifies what is written to the eep save.

AmbientMalice:
The game is extremely reluctant to boot on PJ64 with certain VI Refresh Rate settings.

I've played the game in the past on mupen64plus, and IIRC, disabling the expansion pack allowed you to get past one Indy's Trading Post crash, but not others. It's a very strange issue, and a casual observer might think it was some kind of anti-piracy technique.

loganmc10:
Yeah this seems really strange, when I use HLE graphics, it behaves differently (doesn't print out the DMA error like it does in LLE), but it seems like it still froze, it's hard to tell, but it definitely is sending different DMA data using HLE graphics.

There are other games that lock up using LLE graphics (Perfect Dark comes to mind), in both PJ64 and mupen64plus. Given that the ucode for this game is likely to be deciphered in a year or so (granted this is probably wishful thinking, but if Rogue Squadron gets done, Naboo/Indiana might not be far behind), I think I'm going to put this aside for now, and revisit if/when HLE graphics are accomplished for the game. Everything I am trying feels really hacky and probably not the right way to do it, maybe we'll get lucky with HLE!

AmbientMalice:
I think that's the right approach. Good stab at it, anyway.

PR #334, opened 2017-07-10
Fix write_cart_rom/read_cart_rom

loganmc10:
This fixes https://github.com/mupen64plus/mupen64plus-core/issues/311 (International Track & Field 2000 and Midway's Greatest Arcade Hits Vol. 1)

### DIFF
--- a/src/device/pi/cart_rom.c
+++ b/src/device/pi/cart_rom.c
@@ -33,6 +33,7 @@ void init_cart_rom(struct cart_rom* cart_rom,
 void poweron_cart_rom(struct cart_rom* cart_rom)
 {
     cart_rom->last_write = 0;
+    cart_rom->rom_written = 0;
 }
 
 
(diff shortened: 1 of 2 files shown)


loganmc10:
They don't work in the new dynarec either, but it looks like the new dynarec implements it's own version of the opcodes, so this would need to get implemented over there as well to work with the new dynarec.

bsmiles32:
There is already a similar logic in pi/cart_rom.c
I didn't know which game used that, now I know ;)

bsmiles32:
But it doesn't work with zero value because we use zero as sentinel value. See last_write

loganmc10:
Do you think there is a "common place" that is can be implemented? Or will it need to be implemented in each dynarec as well as the interpreter?

loganmc10:
Oh I misunderstood what you were saying.

Changing ```last_write```'s default value from 0 to -1 (I changed it to a signed int) solved the problem! the games boot! Such an easy fix. Much less hacky as well.

@bsmiles32 what would be a good default value besides 0? I'm not sure if changing it to a signed int will cause problems because of the lower max value. Is there some value that you know would be bogus that we could use?

bsmiles32:
I guess just fixing the read,write_cart_rom should be enough. No need to touch LW/SW.

bsmiles32:
Adding a separate rom_written flag instead of using 0 for last_write should be ok

loganmc10:
Right, okay I've updated the PR, this does fix both games in the dynarec and interpreter. Might fix other things as well, who knows! A lot cleaner now as well, thanks for helping with this

Issue #221, opened 2016-12-27
All Star Baseball 2001 not starting (TLB Exception)


loganmc10:
Nothing happens when I launch this game. I tested mupen64plus on Linux and mupen64plus-FZ, just sits at a black screen forever

loganmc10:
I seem to have the same problem with 2000 as well

fzurita:
To help narrow down the problem. Did you try to use interpreter? Did you try a different RSP, like CXD4? Did you try a different video plugin?

loganmc10:
Ya I tried everything, I tried Pure Interpreter with CXD4 and Glide64 and GLideN64, still nothing.

'99 works though, just not 2000 or 2001

loganmc10:
In case this is helpful to anyone, here is the debugging output with Cached Interpreter:
```
Core: init block A4000000 - A4001000
Core: init block 84000000 - 84001000
```
Here is the 64-bit dynarec:
```
Core: init block A4000000 - A4001000
Core: init block 84000000 - 84001000
Core: NOTCOMPILED: addr = a4000040 ops = 7f0b5b520360
Core: block recompiled (A4000040-A40004C0)
Core: NOTCOMPILED: addr = a4000778 ops = 7f0b5b520360
Core: block recompiled (A4000778-A4000880)
Core: NOTCOMPILED: addr = a4000880 ops = 7f0b5b520360
Core: block recompiled (A4000880-A400090C)
Core: NOTCOMPILED: addr = a400090c ops = 7f0b5b520360
Core: block recompiled (A400090C-A4000980)
Core: NOTCOMPILED: addr = a4000a40 ops = 7f0b5b520360
Core: block recompiled (A4000A40-A4000AD0)
Core: NOTCOMPILED: addr = a4000980 ops = 7f0b5b520360
Core: block recompiled (A4000980-A4000A40)
Core: NOTCOMPILED: addr = a4000ad0 ops = 7f0b5b520360
Core: block recompiled (A4000AD0-A4000B6C)
Core: init block 80000000 - 80001000
Core: init block A0000000 - A0001000
Core: NOTCOMPILED: addr = 80000000 ops = 7f0b5b520360
Core: block recompiled (80000000-800002B4)
Core: init block 80300000 - 80301000
Core: init block A0300000 - A0301000
Core: NOTCOMPILED: addr = 80300000 ops = 7f0b5b520360
Core: block recompiled (80300000-80300100)
[...]

PR #336, opened 2017-07-14
Ignore TLB write if TLB entry is unmapping itself

loganmc10:
Fixes https://github.com/mupen64plus/mupen64plus-core/issues/221

This was adapted from:
https://github.com/project64/project64/blob/master/Source/Project64-core/N64System/Mips/TLBclass.cpp#L128-L146

It allows the games mentioned in #221 to boot. I haven't done extensive testing to see if this breaks other games.

### DIFF
--- a/src/device/r4300/mips_instructions.def
+++ b/src/device/r4300/mips_instructions.def
@@ -1161,6 +1161,12 @@ DECLARE_INSTRUCTION(TLBR)
 static void TLBWrite(struct r4300_core* r4300, unsigned int idx)
 {
     uint32_t* cp0_regs = r4300_cp0_regs(&r4300->cp0);
+    uint32_t pc_addr = *r4300_pc(r4300);
+
+    if (pc_addr >= r4300->cp0.tlb.entries[idx].start_even && pc_addr < r4300->cp0.tlb.entries[idx].end_even && r4300->cp0.tlb.entries[idx].v_even)
+        return;
+    if (pc_addr >= r4300->cp0.tlb.entries[idx].start_odd && pc_addr < r4300->cp0.tlb.entries[idx].end_odd && r4300->cp0.tlb.entries[idx].v_odd)
+        return;
 
     if (r4300->emumode != EMUMODE_PURE_INTERPRETER)
     {


loganmc10:
By the way, once this, along with the other PR's I have waiting are merged, we are at "compatibility parity" with Project64 as far as I can tell. I don't know of any game that doesn't boot with mupen64plus that works in PJ64. This of course excludes the 64DD games.

There might be games that could use a better CountPerOp or CountPerScanline setting, PJ64's RDB receives a little more love and attention compared to mupen64plus's INI, but those are easy things to play with.

gizmo98:
@loganmc10 
Just a question. Shouldn't all ported Project 64 fixes show up under credits? I have seen someone on reddit complaining about license violation.

loganmc10:
Well I'm not a GPL license expert first off. I think the "credits" thing is possible, but it's not a license requirement, more of a nice gesture? I try to be transparent/open about where I'm getting this stuff from, so I mention it in the PR's, which is my way of giving credit.

As far as actual licenses is concerned: the code wasn't copied, I looked at the code and how it worked, and re-implemented it here, so I don't think any copyright was violated.

The person that seems to have the issue with licenses (and it is 1 person), seemed to take issue with a PR I made in mupen64plus-rsp-hle. Basically zilmar copied mupen64plus-rsp-hle into PJ64 for his Android version, and didn't retain any of the copyright notices. bsmiles32 brought it to his attention, and he added the copyright notices. Then, he (zilmar) made a few minor bug fixes, which were ported back (https://github.com/mupen64plus/mupen64plus-rsp-hle/pull/40, https://github.com/mupen64plus/mupen64plus-rsp-hle/pull/43)

Anyway I asked bsmiles32 about it and he didn't seem concerned about it, since the code that was copied was so minor, we're talking a basically moving a few brackets around and moving some lines of code around.

So no, I wouldn't be worried about any copyrights, since nothing here was actually copied.

gizmo98:
@loganmc10 
Seems legit. I didn't unterstand the complaint.

loganmc10:
Personally when it comes to emulators I think copyright is a bit of a silly argument. We are all trying to emulate the same system, so the code is going to look very similar, it's also going to act very similar.

For instance (https://copyleft.org/guide/comprehensive-gpl-guidech5.html):
> As an introductory matter, it is important to note that literal copying of a significant portion of source code is not always sufficient to establish that a second work is a derivative work of an original program. Conversely, a second work can be a derivative work of an original program even though absolutely no copying of the literal source code of the original program has been made. This is the case because copyright protection does not always extend to all portions of a program’s code, while, at the same time, it can extend beyond the literal code of a program to its non-literal aspects, such as its architecture, structure, sequence, organization, operational modules, and computer-user interface.

So really it could be argued that whoever "got there first" (UltraHLE, 1964, whatever older N64 emulators existed) and implemented a certain opcode or function owns the copyright to that function, and anyone who implements it after is creating a "derivative work". If you look at cen64's code it's going to look pretty similar to PJ64 or mupen64plus, but nobody is too concerned about copyright, since we all of us (or most of us) are just trying to have a little fun with a hobby.

loganmc10:
The important thing is that the code is protected by the GPL, so some company can't (or shouldn't) come along, steal the code, and try to sell a closed source emulator based on it. The copyright is important because that is the person that controls the decision around the licenses.

For instance, some guy claiming to be from LG just asked gonetz if he would re-license GLideN64 for them so they can use it without releasing the source of their application. GLideN64 has it's own strange license situation so it's unclear whether gonetz would even really have that authority, but assuming he did, he could re-license it to something other than GPL if he wanted. That's the "power" of being a copyright holder.

gizmo98:
I always thought it is common practice to share code so all emulators benefit. Snes emulators benefit from each other. Libretro cores benefit more or less from upstream. Upstream benefits sometimes from libretro and so on. 

And there is this guy who is complaining who was first as if there is a competition.

This LG guy is really funny. I also do not really now who is the right copyright holder of this piece of code. Reicast for example asked all contributors if relicensing of their contributed code is possible.

> Contributions to the reicast project (excluding mine). Copyright belongs to the contributors, we use them via GPL.


https://github.com/reicast/reicast-emulator/issues/455.

loganmc10:
I tested this with about 15 games (first 1-2 minutes of each game) and I didn't see any regression, given that this is already implemented successfully in another N64 emulator I think it's pretty safe.

bsmiles32:
Sorry for the delay, I'm away from home with limited connectivity. That's why I did only merged "safe" PR. I will tackle the remaining one later. Can you suggest an order for me to merge your work ? With all your work on the audio side things are getting very interesting ! Maybe you could also ask @richard42 write permission. Thanks for your dedication.

loganmc10:
The order shouldn't matter, they are all independent of each other. If a merge conflict is created after one is merged I can sort it out.

> Maybe you could also ask @richard42 write permission. 

I'd rather not be merging my own PR's, it's nice to know at least one other person looked at it. After the current batch of PR's is merged there isn't much I was planning on working on anyway, besides maybe trying to get Indiana Jones to work. I'm open to suggestions though.

> Thanks for your dedication

Thanks, I enjoy it, it's pretty satisfying to solve these problems, it's almost more fun than actually playing the games lol

PR #337, opened 2017-07-16
[WIP] Transfer pak connection to frontend

loganmc10:
This was my attempt to connect the transfer pak code to the end users.

It introduces a new input plugin API function (GetGBCartInfo). So basically the core can query the input plugin to determine what files the user has chosen for the GB ROM/RAM files.

I also changed some of the log output that is generated to "verbose".

Thoughts? I know that I would also have to update the API docs and version numbers. I'll do that if this gets a "thumbs up"

@bsmiles32 one thing I noticed, it doesn't seem to matter what "plugin" I have chosen in the input plugin, the transfer pak always works if I have a valid path to a ROM/RAM chosen

See also https://github.com/mupen64plus/mupen64plus-input-sdl/pull/45

### DIFF
--- a/doc/emuwiki-api-doc/Mupen64Plus-Plugin-Parameters.mediawiki
+++ b/doc/emuwiki-api-doc/Mupen64Plus-Plugin-Parameters.mediawiki
@@ -263,7 +263,7 @@ The Mupen64Plus-Input-SDL plugin uses a separate config section for each simulat
 |-
 |plugin
 |M64TYPE_INT
-|Specifies which type of expansion pak is in the controller: 1=None, 2=Mem pak, 5=Rumble pak
+|Specifies which type of expansion pak is in the controller: 1=None, 2=Mem pak, 4=Transfer pak, 5=Rumble pak
 |-
 |mouse
 |M64TYPE_BOOL
@@ -280,6 +280,14 @@ The Mupen64Plus-Input-SDL plugin uses a separate config section for each simulat
 |AnalogPeak
 |M64TYPE_STRING
 |An absolute value of the SDL joystick axis >= AnalogPeak will saturate the N64 controller axis value (at 80).  For X, Y axes. For each axis, this must be greater than the corresponding AnalogDeadzone value
+|-
+|GBRomFile
+|M64TYPE_STRING
+|Filepath pointing to a GB Cart ROM file. Used if plugin = 4 (Transfer Pak).
+|-
+|GBSaveFile
+|M64TYPE_STRING
+|Filepath pointing to a GB Cart RAM (Save) file. Used if plugin = 4 (Transfer Pak).
 |-
 |}
 
--- a/doc/emuwiki-api-doc/Mupen64Plus-v2.0-API-Versioning.mediawiki
+++ b/doc/emuwiki-api-doc/Mupen64Plus-v2.0-API-Versioning.mediawiki
@@ -112,3 +112,5 @@ This is the most complicated interface, because it involves 3 components: the vi
 ** add (optional) RenderCallback function to input plugin. This function is called by the core after rendering on screen text (OSD) and before the graphics plugin swaps the buffers. The purpose of this function is to enable the input plugin to draw on screen content, for example buttons in a touch input plugin. If this function is not present the core will ignore it and on screen rendering by the input plugin will be disabled.
 * '''CONFIG_API_VERSION''' version 2.3.1:
 ** add new functions "ConfigExternalOpen()" "ConfigExternalClose()" "ConfigExternalGetParameter()" that allows plugins to leverage the core INI parser to read config files.
+* '''INPUT_API_VERSION''' version 2.0.2:
+** add new function "GetGBCartInfo". This function is called by the core to obtain the filepath to GB ROM and RAM (save) files.
--- a/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Plugin-API.mediawiki
+++ b/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Plugin-API.mediawiki
@@ -224,6 +224,9 @@ These functions are present in all of the plugins.
 |-
 |<tt>void SDL_KeyUp(int keymod, int keysym);</tt>
 |Pass a SDL_KEYUP-style message to the input plugin
+|-
+|<tt>void GetGBCartInfo(int Controller, const char **RomFile, const char **SaveFile);</tt>
+|Pass GB Cart filepaths to emulator core.
 |}
 
 === Remove From Older Input API ===
--- a/src/api/m64p_plugin.h
+++ b/src/api/m64p_plugin.h
@@ -245,6 +245,7 @@ typedef void (*ptr_InitiateControllers)(CONTROL_INFO ControlInfo);
 typedef void (*ptr_ReadController)(int Control, unsigned char *Command);
 typedef void (*ptr_SDL_KeyDown)(int keymod, int keysym);
 typedef void (*ptr_SDL_KeyUp)(int keymod, int keysym);
+typedef void (*ptr_GetGBCartInfo)(int Controller, const char **RomFile, const char **SaveFile);
 typedef void (*ptr_RenderCallback)(void);
 #if defined(M64P_PLUGIN_PROTOTYPES)
 EXPORT void CALL ControllerCommand(int Control, unsigned char *Command);
@@ -253,6 +254,7 @@ EXPORT void CALL InitiateControllers(CONTROL_INFO ControlInfo);
 EXPORT void CALL ReadController(int Control, unsigned char *Command);
 EXPORT void CALL SDL_KeyDown(int keymod, int keysym);
 EXPORT void CALL SDL_KeyUp(int keymod, int keysym);
+EXPORT void CALL GetGBCartInfo(int Controller, const char **RomFile, const char **SaveFile);
 EXPORT void CALL RenderCallback(void);
 #endif
 
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -100,22 +100,6 @@ void* g_rdram = NULL;
 
 struct device g_dev;
 
-/* Gameboy roms to load in transfer pak */
-/* TODO: allow ui to pass the gb rom file to the core */
-char* g_gb_rom_files[GAME_CONTROLLERS_COUNT] = {
-#if 0
-    "./pkmb.gb",
-    "./mtennis.gbc",
-    "./pd.gbc",
-    "./pkmc.gbc",
-#else
-    NULL,
-    NULL,
-    NULL,
-    NULL
-#endif
-};
-
 int g_gs_vi_counter = 0;
 
 /** static (local) variables **/
@@ -170,16 +154,6 @@ static char *get_flashram_path(void)
 {
     return formatstr("%s%s.fla", get_savesrampath(), ROM_SETTINGS.goodname);
 }
-
-static char *get_gbsav_path(unsigned int controller)
-{
-    /* gb save files names are suffixed with the controller number
-     * to avoid multiple controllers to write to the same save file.
-     */
-    const char* name = namefrompath(g_gb_rom_files[controller]);
-    return formatstr("%s%s.%u.sav", get_savesrampath(), name, controller);
-}
-
 
 /*********************************************************************************************************
 * helper functions
@@ -1032,6 +1006,7 @@ m64p_error main_run(void)
     sra_storage = (struct storage_backend){ sra.data, sra.size, &sra, save_file_storage };
     eep_storage = (struct storage_backend){ eep.data, (ROM_SETTINGS.savetype != EEPROM_16KB) ? PIF_PDT_EEPROM_4K : PIF_PDT_EEPROM_16K, &eep, save_file_storage };
 
+    int gb_init[GAME_CONTROLLERS_COUNT];
     /* setup game controllers data */
     for(i = 0; i < GAME_CONTROLLERS_COUNT; ++i)
     {
@@ -1040,18 +1015,19 @@ m64p_error main_run(void)
         mpk_storages[i] = (struct storage_backend){ mpk.data + i * MEMPAK_SIZE, MEMPAK_SIZE, &mpk, save_file_storage };
         rumbles[i] = (struct rumble_backend){ &channels[i], rvip_exec };
 
-        if (g_gb_rom_files[i] != NULL)
+        const char* gbsav_path = NULL;
+        const char* gbrom_path = NULL;
+        if (input.getGBCartInfo)
+            input.getGBCartInfo(i, &gbrom_path, &gbsav_path);
+        if (gbrom_path != NULL && gbsav_path != NULL && strcmp(gbrom_path, "") != 0)
         {
-            char* gbsav_path = get_gbsav_path(i);
-            char* gbrom_path = strdup(g_gb_rom_files[i]);
-
             gb_carts_rom[i].data = NULL;
             gb_carts_rom[i].size = 0;
-            gb_carts_rom[i].filename = gbrom_path;
+            gb_carts_rom[i].filename = strdup(gbrom_path);
 
             gb_carts_ram[i].data = NULL;
             gb_carts_ram[i].size = 0;
-            gb_carts_ram[i].filename = gbsav_path;
+            gb_carts_ram[i].filename = strdup(gbsav_path);
 
             if (init_gb_cart(&gb_carts[i],
                              &gb_carts_rom[i], init_gb_rom,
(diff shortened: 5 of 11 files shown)


bsmiles32:
Thanks for tackling this !
I intentionally left it for later because it would have required a change in the API. The way I envisioned it was just to add a load GB cart command to the core.
I'm not a big fan of moving the ROM/ram loading logic to input plugin, I would prefer to have it in the core/frontend (eg same as N64 ROM loading logic). But I welcome other opinions on this, especially from frontend writers.

Changing the log to verbose is the right thing to do (or maybe even placing them inside #if/#endif?)

For the automatic tpak enablement when having a gb_cart loaded see src/plugin/emulate_game_controller_via_input_plugin.c

Also, I don't have access to my laptop ATM that's why I don't merge stuff for now. Sorry for delay.

loganmc10:
> Also, I don't have access to my laptop ATM that's why I don't merge stuff for now. Sorry for delay.

No problem there, I'm in no rush.

The input plugin made more sense to me, since it is meant to emulate the controller. The input plugin already chooses which Pak is enabled. In my mind, which GB cart is plugged into the transfer pak is part of "how is my controller set up".

Anyway it's not really a deal breaker either way, it could also be done in the core. Something like:
```CoreDoCommand(M64CMD_LOAD_GB_ROM, int ControllerNum, char * filepath)```, and the same for the RAM/Save file, was that what you were envisioning?

**Comment by loganmc10** [comment]
In case it helps anyone, this is what I was envisioning the "user experience" would be like with this kind of setup:

https://youtu.be/hiA0ZSBK6WI

If it was handled by the core, I would probably add a "Load GB ROM File" and "Load GB RAM File" option somewhere in the GUI, which would execute ```CoreDoCommand(M64CMD_LOAD.......```. The one issue I see with this is that it wouldn't be persistent, the user would need to do that every time they launched the emulator if they wanted to use the transfer pak.

**Comment by fzurita** [comment]
Doing it through CoreDoCommand does seem simpler to me. Less API change and should be backwards compatible with old input plugins.

I don't see an issue with always having to call "CoreDoCommand". This can be easily automated by the front-end.

**Comment by loganmc10** [comment]
I think the persistence thing is an issue. Picture the "worst case scenario" where there is a transfer pak plugged into each controller. That means you need to load 8 files. If you're using mupen64plus-ui-console, that means we'd probably have to modify it to take a command-line argument like:

```./mupen64plus --gb-rom-1 ./romfile1.gb --gb-ram-1 ./ramfile1.sav --gb-rom-2 ./romfile2.gb --gb-ram-2 ./ramfile2.sav --gb-rom-3 ./romfile3.gb --gb-ram-3 ./ramfile3.sav --gb-rom-4 ./romfile4.gb --gb-ram-4 ./ramfile4.sav ./pokemonstadium.n64```

And you'd have to do that every time you launch the game.

Even with a GUI that can store it's own settings. You'd still have to have a "Load GB ROM" and "Load GB RAM" button, that asked the user for the controller number and filename. Then the GUI would need to store the controller number and filename.

It's a controller specific setting, it makes sense to me that it should be stored with the controller settings.

**Comment by fzurita** [comment]
Why would you have to load it with UI console? Wouldn't CoreDoCommand be enough?

**Comment by loganmc10** [comment]
Maybe I don't understand, what is calling CoreDoCommand if not UI console? You would have to pass those parameters to ui-console when it launches so that it can call CoreDoCommand with the proper values.

**Comment by fzurita** [comment]
With the way mupen64plus-ae does it, CoreDoCommand can be called at any time, not just at startup.

**Comment by loganmc10** [comment]
Yeah but ui-console doesn't have that capability. With a gui frontend you can call it anytime, but in this particular case it would need to be called before the game starts anyway, it wouldn't do anything if you called it while the game was running.

**Comment by fzurita** [comment]
Right, I guess it could also be part of the UI console config. There are definitely a lot of options for this...

**Comment by loganmc10** [comment]
Anyway I'm open to other ways of doing it, but I can't think of a really clean way to implement it using CoreDoCommand that would work well with ui-console. It seems like either you'd have to pass it a ton of commandline parameters, or you'd have to add 8 config options to either ui-console or the core, and at that point it seems like you might as well just add it to the controller configs.

**Comment by bsmiles32** [comment]
I took the liberty to cherry-pick the log level part of your PR and merge it right away. It will decrease the "noise" while we review the other part of the PR.

I stil don't have the definite answer yet on how to proceed, but I will give more perspective on the initial work :
- each gb cart can have various extra devices inside of them such as a RAM, RTC, rumbling device or accelerometer
- we don't have de db for gb cartridges so some heuristic based on GB header is used to know what kind of device is present inside the cartridge
- The gb cart loading code provides this heuristics and only ask the caller (via some callbacks) to initialize the corresponding backend when needed (storage backend for ROM and RAM, rtc backend for rtc if present)
- If multiple tpak are connected, save files should be differents to avoid corrupting them

I'd like to delegate all the ROM,RAM and RTC initialization to the ui.
Pak selection used to be dictated by the input plugin, but in my book this is mistake that is due to the zilmar spec : emulation code pertain to the core, plugins should only take care of plateform specificities. Otherwise we get different level of support, and/or copy pasting horrors as is currently happening in gfx plugins. That is also why I crafted the "backends" which only take care of platform stuff and don't use/require emulation code.

Will try to thing more about it, and come back with suggestions later.

**Comment by Mastergatto** [comment]
> Pak selection used to be dictated by the input plugin, but in my book this is mistake that is due to the zilmar spec : emulation code pertain to the core, plugins should only take care of plateform specificities. 

Forgive me if I join the discussion like this, but if the pak selection moves to the core, then could this spell trouble for [mupen64plus-input-raphnetraw](http://www.raphnet.net/programmation/mupen64plus-input-raphnetraw/index_en.php)? This input plugin allows the real pak accessories to be used directly in mupen64plus through N64 pad and USB adapter.

**Comment by bsmiles32** [comment]
Thanks for joining the discussion and bringing this use case. Hopefully we can find a way to also have it work. Something like if we use emulated controller let the core do the pak selection, otherwise we're in "passthrough" mode and everything is delegated to the plug-in. Reworking the input handling is on my to-do list, but I guess it will be for after the next release.

**Comment by Mastergatto** [comment]
The passthrough mode sounds like a good idea, but this could mean that some changes are needed also on the other side. I hope it won't be too much an issue. 

Also, keep up the good work 👍

**Comment by loganmc10** [comment]
@bsmiles32 I'd like come to some kind of solution on this. I don't mind it being handled by the core, would something like this be acceptable:

```
CoreDoCommand(M64CMD_SET_GB_ROM, int controllerNum, &filePath);
CoreDoCommand(M64CMD_SET_GB_RAM, int controllerNum, &filePath);
```

And then just let the frontend handle that? We can just specify that it needs to be done before the ROM is opened.

I'm not sure how the raphnet plugin would be affected, but it doesn't seem like this system would work any differently than the system I have currently proposed in that respect.

**Comment by bsmiles32** [comment]
@loganmc10 : I've been thinking about it lately and maybe I have an intermediate solution which doesn't require any changes in the API and leave the input plugin alone.
GB carts setups will be stored inside the mupen64plus.cfg file under a game-specific section :
[Perfect Dark (E) !]
tpak-{0,1,2,3}-rom=file path to gb rom
tpak-{0,1,2,3}-ram=file path to gb ram

Numbers indicate which controller (0 to 3) it applies to.
Here I used the GoodName, because its user-friendly but gameid (4 letters) should also be an option.


With that the core will be fully responsible for loading the gb cart.
Frontends just need to let user edit these configuration variables. To help that we should add in our gamedb which games are compatible with the tpak, so that frontends can know if they have to check for a transferpak configuration variables.

Maybe we could also allow for multiple gb carts per controller and allow with a single key press to toggle between gb carts. Could be useful for pokemon stadium which is compatible with several gb carts.

Hope it makes sense. Comments welcome.
---

Also (not strictly related, but still intersting), I've been reworking the si subsystem to extract controllers and paks logic. In there I have a proof of concept with pak switching (some kind of caroussel) which could be extended for gb carts when there are multiples. This is still very WIP and may not be how I will submit it, but it shows where I want to go with all this handling of n64 accessories.

https://github.com/bsmiles32/mupen64plus-core/tree/pif_and_other_things

**Comment by loganmc10** [comment]
I don't think writing game specific settings to mupen64plus.cfg is a great idea, seems a little messy. But what's more, I don't think it's a good idea to expect frontends to write directly to that file (outside the API I mean, right now all writes/reads go through the API).

I don't think this is really urgent, so I wouldn't really look to any intermediate solution. If pak switching is implemented, wouldn't it make sense to just extend CoreDoCommand like I mentioned?

```
CoreDoCommand(M64CMD_SET_GB_ROM, int controllerNum, &filePath);
CoreDoCommand(M64CMD_SET_GB_RAM, int controllerNum, &filePath);
```
If the frontend executes that command, it would trigger a pak switch/load. The frontend can be in charge of keeping track of with pak is for what game, etc...

**Comment by bsmiles32** [comment]
@loganmc10 Thanks for commenting, I'll try to better argue/explain my point :)

I suggested storing tpak's configuration inside mupen64plus.cfg because it is the file that the core uses. Besides, it stores will store user's specific GB carts path, so I think it has is place inside this file.
Yes it is game-specific, but usually compatible GB carts are game-specific as well so it's not like it's a hack or something.
Frontends can (and should !) access these variables via the Config API so I guess everything is already in place for it to work.

Benefits of this approach is that once what I proposed is implemented in the core, all ui can benefit from that without needing updating. Yes user will have to manually edit the mupen64pls.cfg in this case, but eventually ui will be updated and will allow users to directly edit these variables from inside the ui.
Also it will be much easier to implement both pak switching and gb cart switching inside the core than hope that frontends will implement that properly (especially for the ui-console which is not really an interactive interface... eg once you run the game, we expect no user input (no gb cart path) until the end of execution).

Hope to have cleared some of your concerns. Still open to suggestions and comments as I can have missed something here.

**Comment by loganmc10** [comment]
> Frontends can (and should !) access these variables via the Config API so I guess everything is already in place for it to work.

Sorry yes I realized I had a bit of a "brain freeze" there.

I guess it just seems a bit "clunky" to me, but it makes sense.

> Frontends just need to let user edit these configuration variables. To help that we should add in our gamedb which games are compatible with the tpak, so that frontends can know if they have to check for a transferpak configuration variables.

I'm still a little weary of the per-game thing, and I'm not sure how letting the frontend know which ROM's support tpaks would help. Picture a traditional desktop GUI. If a user downloads that GUI and the first thing they want to do is play Pokemon Stadium with a GB cart, they'll go to the configuration area and set everything up before they ever open the ROM. The frontend won't know what ROM they want until the game has already started. Or are you picturing the frontend having some kind of "dropdown" with a list of games that support the tpak? I doubt many frontends would go to all that work, it's probably better to just make it a general thing, and if the user wants to play a different game, they modify the config to point to the correct GB cart

**Comment by loganmc10** [comment]
I would just do it like this:

```
[TransferPak]
tpak-{1,2,3,4}-rom=file/path/to/gb.rom,file/path/to/gb2.rom
tpak-{1,2,3,4}-ram=file/path/to/gb.ram,file/path/to/gb2.ram
```

Just separate the files by commas to allow switching, and number them 1-4 (just because computers start from 0 doesn't mean we have to make the humans do it ;) )

Anyway I've come around to the mupen64plus.cfg option, I'm just not sure about the game specific thing, maybe someone else can offer their opinion

**Comment by loganmc10** [comment]
I just think given the relatively low development activity related to the frontends, it's unlikely that sort of system would see widespread adoption. Even with a more generic approach, a frontend could still choose to manage transfer pak association per game itself (store the per game configurations itself, and then modify the [TransferPak] section of mupen64plus.cfg right before the game starts).

ui-console could adopt your per-game system for instance, and then copy the relevant info into the [TransferPak] section when it knows what ROM is about to be opened.

I think that gives the frontends the flexibility to implement a really good per-game system if they want, but there could still be an easier to implement generic solution as well.

**Comment by bsmiles32** [comment]
> I just think given the relatively low development activity related to the frontends, it's unlikely that sort of system would see widespread adoption.
Maybe, but even in that case, if we document these config variables, users will be able to edit them by hand and it will work. That way we don't force frontend writers to update their code if they don't want, while still letting user enjoy this feature.

Letting the frontend manage the game-specific-ness of the configuration as you suggests is indeed a viable option, but will leave each frontend to write it's own implementation of this, whereas if it's in the core, you get it (almost) for free.

**Comment by tony971** [comment]
This is one application where I wouldn't mind per-game info in mupen64plus.cfg. The list of games that support transfer packs is common knowledge. There isn't any game that will accidentally be included when it shouldn't be or vice versa. It's something that will be populated once and forgotten (Hacks excluded, but they should know what they're doing).

**Comment by fzurita** [comment]
For Mupen64plus-AE it doesn't really matter where it's kept, since it's actively developed. Keeping it in the Mupen64plus.cfg file does mean that the front end does not need to be updated.

**Comment by richard42** [comment]
Sorry for coming so late to the party. This is a tricky architectural design issue. A lot of points have already been brought up in the previous comments, so I'll start by addressing some of them. Personally I really don't like the idea of adding per-game setting to the main config file. It reminds of the disaster that Microsoft created (and the world pays for) with the Windows NT registry. I would rather keep the scope of the config file limited to the current configuration of the emulator. I see how it would be advantageous for certain use cases, and the per-game setting even wouldn't be used by most people at the beginning, but I think it's a start down the wrong path.

Regarding the command-line options for the paths with the ui-console front-end, I don't see that as a problem. The 8-path case is really an extreme situation, and in most cases the user would only enter 2 paths. For people who are okay with typing, this is not a big deal. For people who can't handle typing, they're not using the ui-console front-end anyway.

I'm okay with adding filepaths for currently connected transfer paks to the core section or maybe the input sections, but I'm not convinced that this is clearly the right approach either. One problem with putting the paths in the input plugin controller sections is that the auto-config logic might blow them away if the controller is changed.

Stepping back and thinking about the whole system, you could make the case that in order to most closely mimic the operation of the real N64, the API should be changed so that the core can make a function call to get the actual contents (byte buffers) of the GB ram and rom data from the input plugin. This might be the best option for supporting the real hardware with a plugin like Raphnet. If we did this, the SDL-Input plugin would have the logic for determining the filenames and reading the files. It could even be enhanced to support per-game file paths in its own config sections, but then again could suffer from data loss if the sections were deleted by the auto-config logic. There are tradeoffs.

Another possibility is to put this GB RAM/ROM paths in the mupen64plus.ini file, which already has all of the per-game sections. The downside here is that this is supposed to be a global file and not contain user-specific configuration data.

Here's another question: for the transfer pak support, does the input plugin need to add more keyboard/joystick mapping options, to be able to select mem, rumble, or transfer pak?

Getting back to the ram/rom path issue, here is another possible solution. Extend the Core-FrontEnd API by adding functions (DoCoreCommand) for setting the paths. These should be usable at any time, even while the game is running. Add command-line options to the ui-console front-end so that the user can specify these. But then also extend the ui-console so that it will take a path to a directory containing ram/rom images with specially formatted filenames. The format can be something like

<GameName>-<ControllerID>.ram/.rom

Where GameName can be a GoodName or CRC or MD5 or gameid or whatever you want, and ControllerID is a number from 1 to 4. Then the front-end could look for any applicable files after loading the ROM and automatically set the proper paths before running the emulation. On unix systems you could even use symbolic links to make multiple games or controllers use the same files.

**Comment by bsmiles32** [comment]
@richard42 thanks for joining the discussion.
Given that game-specific config is not popular, I think the best way to go then is extend the API with something like that :
- the UI register a set of callbacks that initialize the various components of a GB cart (eg ROM, RAM, RTC,... )

```
DoCoreCommand(M64CMD_SET_GB_CART_LOADER, sizeof(gb_cart_loader), &gb_cart_loader);

struct gb_cart_loader {
voir* cb_data; // allow UI to carry some state to the following callbacks
int (*init_rom)(void* cb_data, int controller_num, const char** rom_filename);
int (*init_ram)(void* cb_data, int controller_num, const char** ram_filename);
int (*init_rtc)(void* cb_data, int controller_num, time_t* rtc_date); // can be added in a future revision if needed ?
};
```

- the core calls these callbacks appropriately depending on what hardware is present inside the GB cart when a tpak is connected for the first time (eg not when switching between mempak,rumblepak and transferpak) and when a new GB cart is connected to tpak. This might require an extra per-controller button to trigger this action. Maybe we can repurpose the {Mem,Rumble}pak switch buttons : 1 for Pak switching (mem -> rumble -> tpak -> none -> mem -> ....) depending on what paks are compatible with the game, 1 for GB cart switching when a tpak is connected.


Notes:
- That way the UI doesn't need to know how to parse a GB cart, the core does it
- All the persistence of GB paths is up to the UI, so we keep core config game agnostic
- If we don't need to change between multiple GB carts, GB cart loading is done only once at first tpak connection (eg potentially bother the user once). Still we support changing GB cart en live. Interactive UI (eg desktop GUI) can popup a file picker dialog to select appropriate files. Non interactive UI (eg ui-console) can select between several pre-selected GB carts in a carousel fashion. Or any other scheme that the UI wants to implement.
- if we need to extend the gb_cart_loader because one crazy hardware shows up (GB camera ?), we just need to add a new field (say init_camera) and add proper struct size check inside the core command handler.
- I would prefer if the input plugin is left untouched (except maybe for the extra per-controller GB cart changing button) because :
- I plan, in the future, to replace input plugins with controller_input_backends (which only abstract gamepad buttons and joystick state - no paks, rumble or GB cart logic) directly backed into the core. And for input plugins that needs more than that (eg raphnet) there will be a joybus_device interface which allow for "raw" pif commands processing.
- Moving this GB cart logic to the input plugin will lead to different level of support for the different input plugins. Also it doesn't make sense for all of them : for instance the raphnet plugin wouldn't need that at all because the real N64 controller already handle paks accesses for us (eg it's not emulated, it's the real deal)

What do you think ?

loganmc10:
Seems good to me, I'm a little confused what the purpose of init_rtc(). What is the frontend supposed to do there? Isn't it something that the core can handle? What information would the frontend be privy to that the core isn't?

bsmiles32:
I think RTC has an internal counter to keep track of time. Core or UI can setup this counter to start from a previous time or give the current time. For the first iteration we can omit it I guess.

richard42:
I'm fine with this proposal.

PR #339, opened 2017-07-18
Use count_per_op instead of 2

loganmc10:
Someone can correct me if I'm wrong, but I'm fairly certain that is "2" is supposed to represent the "count_per_op" setting


loganmc10:
Given that this is supposed to be a random number, and CEN64 literally just returns a random number:

https://github.com/tj90241/cen64/blob/0d0e042817a73950cc5350719c0481be149d5b45/vr4300/cp0.c#L261

It's probably not a big deal, but I think using count_per_op is more "correct"

PR #341, opened 2017-07-20
Proper fix for Twisted Edge audio

loganmc10:
This mostly reverts https://github.com/mupen64plus/mupen64plus-core/pull/289

This was taken from https://github.com/mamedev/mame/commit/aee15107876b7ce7f9b298931b4c4710f38a12ee so credits to @Happy-yappH

The author states that this behavior is actually the result of a hardware bug. I'm not aware of any game besides Twisted Edge that runs into this, but at least this is a generic solution instead of being game specific.

In practice is does the same thing, so there is no real difference, other than a more proper solution, and perhaps this also fixes some audio issue is another game, I'm not sure.


fzurita:
Wow, this is some pretty good research you have been doing lately.

PR #342, opened 2017-07-20
[WIP] Add support for streaming audio buffer

loganmc10:
This PR proposes a change to the API, so I haven't updated the docs yet, I'll do that if the API change gets a thumbs up.

This fixes audio in Boss Games (Stunt Racer, World Driver Championship, Top Gear Rally, Twisted Edge Snowboarding).

You can read my explanation of the problem/solution here:
https://github.com/mupen64plus/mupen64plus-core/issues/200#issuecomment-316780301

Basically in this PR, every time AI_LEN_REG is read, we update the contents of the audio buffer. These games update the audio data while the DMA is happening, so we can't just copy the memory at the beginning of the DMA and send it to the audio plugin.

Because of this, I created a new audio API function (PlayAudio), which just takes a buffer and plays the audio in it, instead of having to read from DMEM.

You can see the change I made to the audio plugin here:
https://github.com/m64p/mupen64plus-audio-sdl2/commit/79fef8d055e78664e13f9855e34fc98631dd7ac7

These games don't seem to play nice with audio speed limiters/"sync to audio" settings, so if you're going to test it, you'll probably need to test it with that sort of thing disabled.

I'm hoping some people can test this, I can provide a binary build if needed.

(technically this is actually the more accurate way to emulate the AI, but since it will add some audio latency because the audio plugin won't start playing the audio until the end of the DMA, I've only enabled it for these Boss games)

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -6641,6 +6641,7 @@ Players=4
 SaveType=Controller Pack
 FixedAudioPos=1
 CountPerScanline=1600
+StreamAudio=1
 
 [CE9AE0AA6DBBF965B1F72BC3AA6A7CEF]
 GoodName=King Hill 64 - Extreme Snowboarding (J) [b1]
@@ -13383,6 +13384,7 @@ CRC=9510D8D7 35100DD2
 SaveType=Controller Pack
 Players=4
 Rumble=Yes
+StreamAudio=1
 
 [14DDF19382385774F253049ABEAF01D2]
 GoodName=Summer64 Demo by Lem (PD) [b1]
@@ -14260,6 +14262,7 @@ CRC=7F43E701 536328D1
 Players=4
 Rumble=Yes
 SaveType=Controller Pack
+StreamAudio=1
 
 [48D6C194AE106018DDFC3486A8B347F7]
 GoodName=Top Gear Rally (E) [b1]
@@ -14277,6 +14280,7 @@ CRC=0E596247 753D4B8B
 Players=4
 SaveType=Controller Pack
 Rumble=Yes
+StreamAudio=1
 
 [6F7030284B6BC84A49E07DA864526B52]
 GoodName=Top Gear Rally (U) [!]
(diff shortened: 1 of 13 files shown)


loganmc10:
Some of these games could use a bit of tweaking, World Driver Championship seems to be more smooth with CountPerOp=1, but I'll leave that kind of thing for another PR

AmbientMalice:
Great work, man.

fzurita:
This is really good work. Is the amount of latency that it adds noticeable? If it's in the milliseconds range and it's a more accurate way of emulating audio, why not have it always enabled?

loganmc10:
I haven't measured the latency, but the "chunks" are 16384 bytes, the sample rate is 22000Hz (in World Driver Championship) if I remember correctly, and there are 4 bytes per sample. So I think that works out to roughly 1/5 of a second if I did the math right (could be wrong)

In driving games like these I don't think audio latency is as noticeable, platformers or fighters might make it more noticeable.

In most games it will make absolutely no difference since the audio buffer doesn't change mid-DMA, if there are other games with bad audio we can test this setting to see if it helps. It just seemed like there wasn't a need to introduce any latency if it wasn't going to improve anything

fzurita:
For games that are not affected, wouldn't the buffer sizes be a lot smaller, so the latency would be a lot less?

But yeah, 16384 bytes at 22Khz is ~0.18 seconds.

loganmc10:
Yeah that's true, most games use a much smaller buffer, like 2k or so.

fzurita:
Which is like 2ms at 22KHz. I don't think a human would perceive that.

fzurita:
Maybe it's the case that the only games that use such large buffer are the one that use the streaming technique.

loganmc10:
Thanks for asking these questions and getting me to think! I've thought up a much simpler solution to this problem that shouldn't affect the latency in a noticeable way. I created a new PR so I can keep this code around if I ever want to look at it again.

https://github.com/mupen64plus/mupen64plus-core/pull/343

PR #343, opened 2017-07-21
Send audio data to plugin as the DMA drains

loganmc10:
This is an improved version of https://github.com/mupen64plus/mupen64plus-core/pull/342. Please read that for details on the problem that this fixes.

This version is much simpler and doesn't require any changes to the API. Instead of sending all the audio data to the plugin at the beginning of the DMA, we send it piece by piece as the DMA happens. Then, at the end of the DMA, we send whatever is leftover.

This should be a "global solution" that doesn't add any noticeable latency, while still fixing the audio in these games.

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -6646,7 +6646,6 @@ GoodName=King Hill 64 - Extreme Snowboarding (J) [!]
 CRC=519EA4E1 EB7584E8
 Players=4
 SaveType=Controller Pack
-FixedAudioPos=1
 CountPerScanline=1600
 
 [CE9AE0AA6DBBF965B1F72BC3AA6A7CEF]
@@ -14680,7 +14679,6 @@ CRC=E688A5B8 B14B3F18
 SaveType=Controller Pack
 Players=2
 Rumble=Yes
-FixedAudioPos=1
 CountPerScanline=1600
 
 [4F0E2AF205BEEB49270154810660FF37]
@@ -14699,7 +14697,6 @@ CRC=BBC99D32 117DAA80
 Players=2
 Rumble=Yes
 SaveType=Controller Pack
-FixedAudioPos=1
 CountPerScanline=1600
 
 [0B0DF8EC747BF99F3A55A3300CE8BC0D]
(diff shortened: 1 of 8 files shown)


loganmc10:
I tested this, along with https://github.com/mupen64plus/mupen64plus-core/pull/341 with a good number of games. I didn't hear any regressions.

AmbientMalice:
Out of curiosity, does this help with Body Harvest, at all?

loganmc10:
Do you want to hear something really weird about Body Harvest?

When I'm playing Body Harvest, the Anti-Virus on my PC goes crazy, maxing out the CPU, and it slows down the game to the point where it gets choppy. Other N64 games don't do this. It doesn't seem to happen on Project64. But besides that it works perfectly (like on Linux it sounds perfect).

What issues are you having with Body Harvest exactly? Actually I noticed that Body Harvest triggers the same issue that Twisted Edge had with audio. You know how Twisted Edge has no sound in Project64? Well the same thing that causes Twisted Edge to have no sound happens in Body Harvest, but it happens intermittently, and it is fixed by https://github.com/mupen64plus/mupen64plus-core/pull/341

loganmc10:
@AmbientMalice if you or anyone else wants to try out these audio fixes I've attached a version of https://m64p.github.io with the audio fixes applied

[mupen64plus-audio-fix.zip](https://github.com/mupen64plus/mupen64plus-core/files/1173407/mupen64plus-audio-fix.zip)

twinaphex:
Have you tested if this PR is safe with savestate loading? I ask because I have experienced plenty of crashing problems so far when attempting to load a savestate. Sometimes it works without crashing, other times it crashes on accessing the buffer.

loganmc10:
I tested a bunch of games, no issue with save states. Save states don't really work reliably with World Driver Championship or Stunt Racer. I assume it's because the value of ```rsp_task_locked``` isn't stored within the save state.

loganmc10:
I combined this PR with https://github.com/mupen64plus/mupen64plus-core/pull/341 since they would have created merge conflicts with each other, and they are mildly related

loganmc10:
I fixed the merge conflicts that arose

bsmiles32:
I think I did it also using the github editor, but you beat me to it :)
Just a matter of waiting Travis to finish and I merge it.

PR #347, opened 2017-07-25
Fix timing in DK64

loganmc10:
I'm hoping a couple people can test this out before it gets merged.

I believe this improves the timing in DK64 because:

During the intro, right after the rap sequence, there is a "Press Start" screen. In Project64, and currently in mupen64plus, Donkey Kong jumps into the water, climbs up a hill, and then jumps into the water again, missing the vines.

With this change, he correctly climbs up the hill, then jumps across the vines and into the cannon. This is the correct sequence for the intro of the game.

I've never really played this game much, so I'm hoping some people can test this out in various stages of the game. The settings are CountPerOp=3 and CountPerScanline=1750. You should be able to just modify mupen64plus.cfg to test this, no need to recompile anything

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -3658,7 +3658,8 @@ CRC=11936D8C 6F2C4B43
 SaveType=Eeprom 16KB
 Players=4
 Rumble=Yes
-CountPerOp=1
+CountPerOp=3
+CountPerScanline=1750
 ; Bone displacement fix
 Cheat0=806128E2 0000
 
@@ -3682,7 +3683,8 @@ GoodName=Donkey Kong 64 (J) [!]
 CRC=053C89A7 A5064302
 SaveType=Eeprom 16KB
 Players=4
-CountPerOp=1
+CountPerOp=3
+CountPerScanline=1750
 ; Bone displacement fix
 Cheat0=806170A2 0000
 
@@ -3712,7 +3714,8 @@ GoodName=Donkey Kong 64 (U) [!]
 CRC=EC58EABF AD7C7169
 SaveType=Eeprom 16KB
 Players=4
-CountPerOp=1
+CountPerOp=3
+CountPerScanline=1750
 ; Bone displacement fix
 Cheat0=80619632 0000
 


loganmc10:
By the way, there are other settings that will allow DK to catch the vine in the intro, CountPerOp=2 and CountPerScanline=1100 also works.

The other issue with this game is the timed sections. There are sections where you have 60 seconds to complete a task for instance.

With CountPerOp=2 and CountPerScanline=1100, the timer for the 60 second section took 75 seconds.

With CountPerOp=3 and CountPerScanline=1750, the timer for the 60 second section took 50 seconds.

Basically it seems like if we could do a "CountPerOp=2.5", the timing would be pretty close to perfect, but this is probably as close as we are going to get with our current timing system.

loganmc10:
hm actually I just tested the current default settings, and although DK doesn't hit the vine in the intro, the timed section is almost exactly 60 seconds. It's probably more important that the timing in these timed sections is correct, rather than the intro, so I'll close this. What a strange game

fzurita:
I think we need to have a "variable count per op" setting. Where count per op is different depending on the instructions being executed.

fzurita:
Did the timing for this game change at all when you made count per scanline dynamic?

fzurita:
I have the feeling that the game is controlling dk through the timing of some interrupt and changing count per op or count per scanline changes that timing.

loganmc10:
> Did the timing for this game change at all when you made count per scanline dynamic?

Changing the count per scanline does modify DK's path in the intro, but the change you are referring to only changed it from 1500 to around 1545, not enough to probably make any noticeable difference.

> Maybe changing sp_delay_time or add_interrupt_event in some places will have some effect on the timing in the intro.

You are probably right about this, it could be the audio interrupt or any other interrupt, but yes it is probably measuring the time between some interrupt. It seems like if you get the timing right during the intro, it's wrong at the timed in-game sections though. It'll probably be hard to get both of them right

fzurita:
From your initial look at this problem, it seemed that count per op was the only thing that affected the timed sections? So if we set that to 2 and only change the interrupt timing, maybe it will work?

loganmc10:
I can't remember exactly. In my experience modifying CountPerOp was a big difference and CountPerScaline was a smaller change, at least for the intro scene.

Changing CountPerScaline is basically just changing the VI interrupt timing, so in they respect it would be no different than trying to modify the audio interrupt timing or something. But DK may be measuring the time between VI interrupts, and in that case you would change that interrupt timing by changing CountPerScaline

fzurita:
Ok

fzurita:
I tried modifying the interrupt times, it didn't affect the game controlled dk movement.

Interestingly, modifying the ai interrupt time made the game run faster or slower.

PR #348, opened 2017-07-26
Fix Indiana Jones

loganmc10:
This is my latest attempt to get Indiana Jones working.

CountPerOp=1 fixes graphical and stability issues
CountPerScanline=2200 allows the game to boot with CountPerOp=1

DisableExtraMem=1 disables the expansion pak. This seems to fix the hang at the end of the level, right before the Trading Post. The expansion pak brought the game to 640x480, but I think disabling it is worth it to make the game playable, at least until a better fix is found.

I'm going to test this over the next few days to make sure I got this right. These are all just settings you can set in mupen64plus.cfg, so anyone can test this out. If you're going to test I would suggest deleting any save (eep) files associated with the game.

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -5962,12 +5962,18 @@ GoodName=Indiana Jones and the Infernal Machine (U) [!]
 CRC=AF9DCC15 1A723D88
 Players=1
 SaveType=Eeprom 4KB
+CountPerOp=1
+CountPerScanline=2200
+DisableExtraMem=1
 
 [6F417D30D772F4420C9384E9BBB7BC01]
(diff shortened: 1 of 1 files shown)


AmbientMalice:
Disabling the expansion pack causes the music to sound strange, makes one of the levels inaccessible (it gets skipped), and in my playtesting on m64p in the past, the game will proceed to freeze at the end of later missions. If not the second, then the third.

loganmc10:
Hm yep it appears you're right. The audio sounds fine to me, it froze at the end of the 2nd level. Setting "DelaySI" allowed it to complete the 2nd level, but then I went back and tested the 1st level with DelaySI and it froze....

PR #351, opened 2017-07-28
Add SI_STATUS_DMA_BUSY

loganmc10:
This fixes the error:
```
Core Warning: two events of type 0x8 in interrupt queue
```
in some games. I've noticed it in Diddy Kong Racing and Mischief Makers (it really spams this message in Mischief Makers).

Anyway they are doing this because they don't know the SI DMA is busy. I added the busy bit, now these errors don't show up anymore.

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -8706,6 +8706,7 @@ GoodName=Mischief Makers (E) [!]
 CRC=418BDA98 248A0F58
 Players=1
 SaveType=Eeprom 4KB
+DelaySI=0
 
 [5690D74157C6623E2928A6F0353EF4AF]
 GoodName=Mischief Makers (E) [b1]
@@ -8717,6 +8718,7 @@ GoodName=Mischief Makers (U) [!]
 CRC=0B93051B 603D81F9
 Players=1
 SaveType=Eeprom 4KB
+DelaySI=0
 
 [CCF012DF82022D4797CE4CC5405E084F]
 GoodName=Mischief Makers (U) [b1]
--- a/src/device/si/si_controller.c
+++ b/src/device/si/si_controller.c
@@ -33,8 +33,10 @@
 enum
 {
     /* SI_STATUS - read */
+    SI_STATUS_DMA_BUSY  = 0x0001,
+    SI_STATUS_RD_BUSY   = 0x0002,
+    SI_STATUS_DMA_ERROR = 0x0008,
     SI_STATUS_INTERRUPT = 0x1000,
-
 };
 
 static void dma_si_write(struct si_controller* si)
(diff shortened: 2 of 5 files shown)


loganmc10:
I also added SI_STATUS_RD_BUSY, which is only applied for ```dma_pi_write()``` (a write to SI_PIF_ADDR_WR64B_REG). Perhaps someone could confirm that is correct before this is merged.

rcp.h says that SI_STATUS_RD_BUSY means "IO read busy". It's a little clear to me what that means, but I assume is means "when dram is being read from", which would be the case for ```dma_pi_write()```

loganmc10:
Actually now I'm not sure, wouldn't the "IO" be the data from the PIF? In that case the bit should be set during ```dma_si_read``` @bsmiles32 do you know how this is supposed to work?

bsmiles32:
I don't really know how SI_STATUS_RD_BUSY should be interpreted.
While these changes fix the "two event of type 0x8 in interrupt queue" spamming, it breaks button presses on my end in Mischief Maker.

loganmc10:
Hm ok thanks for testing, I'll do some more testing and research

loganmc10:
@bsmiles32 do you know why 0x900 was chosen as the interrupt delay time? Or what games need/benefit from the SI delay?

Reducing the 0x900 number fixes Mischief Makers. In my reading, the peak DMA transfer rate for the RSP is 8 bytes per cycle. So I have a couple questions:

What is the 0x900 number supposed to represent? What is it counting?

If it is supposed to be counting ops or cycles, is ```PIF_RAM_SIZE / 8``` a good approximation? It is a very small number (0x8), I can test this on a few games, I'm just wondering if you had any insight into how the current system came to be.

loganmc10:
Even setting it to 0x100 makes it work, which is the number in the comment, but I'm not sure why it's commented

loganmc10:
Shortening the interrupt timing broke Body Harvest, so I'm assuming that's not a safe thing to do.

I've modified the PR to get rid of RD_BUSY, since I don't see any benefit of using it. I also added DelaySI as a INI param and disabled it for Mischief Makers. This fixes the input. Does this seem ok? Diddy Kong Racing seems to perform as normal. I'm not aware of any other game that had those error messages.

bsmiles32:
Don't know about the 0x900 value nor the delay si behavior. I guess it's pure try and guess what worked for selected titles (like for any timing stuff in m64p).
I'm not sure what speed the SI DMA has, so a test ROM is what we need here.

The updated PR looks OK to me, but some more testing is welcome before I merge it.

loganmc10:
I guess this was tried once before:
https://github.com/mupen64plus/mupen64plus-core/commit/252d09f30cfd53a5be6f069e124e6051d03041b4

And reverted for this same reason:
https://github.com/mupen64plus/mupen64plus-core/commit/b77d5512db99650ab435f05daae02c629081c542

I'll do a fair bit of testing for this change to make sure it doesn't cause issues in any other games

loganmc10:
I tested this with about 60 games. I made sure the input worked in all of them. I also made sure that input and saving works in Diddy Kong Racing and Mischief Makers.

It all looks good to me, no issues. If we ever run across another game like Mischief Makers that has issues, disabling DelaySI will probably do the trick, and now it can just be added to the ini file.

PR #356, opened 2017-08-09
Get rid of "alternate VI timing"

loganmc10:
I noticed a while ago (https://github.com/mupen64plus/mupen64plus-core/pull/196#issuecomment-298761293) that this "alternate timing" hack only makes a difference when the result of the calculation makes vi->regs[VI_CURRENT_REG] = 0

I got the idea for "< 40" from this comment: https://github.com/tj90241/cen64/blob/master/vi/controller.c#L83-L87

Sometimes ```vi->next_vi - cp0_regs[CP0_COUNT_REG]``` can actually be a negative number in Pokemon Puzzle League, so we've already passed the VI interrupt when this math is being done, which screws up the calculation.

Anyway I've tested this with a number of games and I don't see any regressions, however it is modifying the VI interrupt system, which is already a little "fragile" as it is, so I will probably do some more testing over the next few days.

### DIFF
--- a/src/device/vi/vi_controller.c
+++ b/src/device/vi/vi_controller.c
@@ -29,9 +29,6 @@
 #include "main/main.h"
 #include "plugin/plugin.h"
 
-/* XXX: timing hacks */
-enum { NTSC_VERTICAL_RESOLUTION = 525 };
-
 unsigned int vi_clock_from_tv_standard(m64p_system_type tv_standard)
 {
     switch(tv_standard)
@@ -61,13 +58,12 @@ unsigned int vi_expected_refresh_rate_from_tv_standard(m64p_system_type tv_stand
 }
 
 void init_vi(struct vi_controller* vi, unsigned int clock, unsigned int expected_refresh_rate,
-             unsigned int count_per_scanline, unsigned int alternate_timing,
+             unsigned int count_per_scanline,
              struct r4300_core* r4300)
 {
     vi->clock = clock;
     vi->expected_refresh_rate = expected_refresh_rate;
     vi->count_per_scanline = count_per_scanline;
-    vi->alternate_timing = alternate_timing;
     vi->r4300 = r4300;
 }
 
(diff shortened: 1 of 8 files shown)


src/device/vi/vi_controller.c
```diff
@@ -89,13 +85,19 @@ int read_vi_regs(void* opaque, uint32_t address, uint32_t* value)
     {
         /* XXX: update current line number */
         cp0_update_count(vi->r4300);
-        if (vi->alternate_timing)
-            vi->regs[VI_CURRENT_REG] = (vi->delay - (vi->next_vi - cp0_regs[CP0_COUNT_REG])) % (NTSC_VERTICAL_RESOLUTION + 1);
+        int remaining_vi_duration = vi->next_vi - cp0_regs[CP0_COUNT_REG];
+
+        if (remaining_vi_duration < 40)
```
bsmiles32: Naming this magic const and adding a comment like in cen64 would help understanding this snippet.

loganmc10:
I tested this with a pretty good number of games and saw no regressions, more testing or comments is always welcome of course

loganmc10:
I think my previous fix was incorrect. Just doing:

```
if (vi->regs[VI_CURRENT_REG] >= NTSC_VERTICAL_RESOLUTION)
    vi->regs[VI_CURRENT_REG] -= NTSC_VERTICAL_RESOLUTION;
```

Fixes the issue and it makes a lot more sense to me. VI_CURRENT_REG can't be higher than the number of scan lines (as I understand it). So it we are higher than NTSC_VERTICAL_RESOLUTION, we wrap around.

I need to test a bit more, and I need to test with PAL as well (I assume it needs a different value), but this feels a lot more correct to me.

loganmc10:
Ok I tested both NTSC and PAL versions of Pokemon Puzzle League and they work. I also tested a good number of others games and they work as well. This seems pretty good to me.

Issue #145, opened 2015-11-30
Donkey Kong 64 - Bone Displacement


Isotarge:
I'm reporting this long standing issue on behalf of the DK64 RTA & TAS communities.

I have completed a thorough technical analysis of the symptoms [here](https://docs.google.com/document/d/15XNghhrc-thXJpbKzWWuHXl-NXdzcaDLEf7poFVruh8)
Savestates to reproduce the issue are available [here](https://www.mediafire.com/folder/x5u53bpu42cex/Bone_Displacement_States)

Let's squish this bug!

Gillou68310:
Wow if only every issues could be described that way!!!

As a fan of DK64 I totally have interest into fixing this issue, be sure that this is right on top of my TODO list. However I may not have time to look at it until next week.

Having a concrete example of the correct behaviour is definitely gonna help finding the root issue. Both situation can know be compared.

> My current theory as to why this fixes the problem is that it allows the loop processing the bone data for each object to completely finish before being interrupted

You might be right with this because reducing the cycle count per emulated instruction reduces the issue. I'll start with that.

Gillou68310:
Quick question, how can I find the pointer to the object data for DK's character in your savestate (st0)?
Here's the pointer list of currently loaded objects at 0x7FBFF0:

0x7FBFF0	803904e0
0x7FBFF4	80391030
0x7FBFF8	80397210
0x7FBFFC	803a5620
0x7FC000	803a9760
0x7FC004	803c4120
0x7FC008	803d8f20
0x7FC00C	803eb8c0
0x7FC010	803eba50
0x7FC014	803ebbe0
0x7FC018	803ebd70
0x7FC01C	803ebf20
0x7FC020	803ec0b0
0x7FC024	80419fe0
0x7FC028	8043cc00
0x7FC02C	804851e0

Isotarge:
Ah yes, I probably should have mentioned this somewhere. A pointer to the player object can always be found at 0x7FBB4C

Gillou68310:
Thanks I'll check this ;-)

theballaam96:
This issue seems to have died down a bit, and it'd be awesome if it were to be solved. Hopefully it can be picked up again :)

Isotarge:
[...]

PR #357, opened 2017-08-10
[WIP] Switch FGR to 32bit mode when jumping to an exception handler.

loganmc10:
First of all, credit for this goes to @Gillou68310, he's the one that actually wrote this code.

You can read the original discussion about this problem here:
https://github.com/mupen64plus/mupen64plus-core/issues/145#issuecomment-249221815

Basically, this is a fix for the "Bone Displacement" issue in DK64, as well as some flickering graphics in Factor5 games, see below:
29176429-190830d0-7da9-11e7-9bab-1289473d5ca1

indiana_jones-028

I tested this with the save states that @Isotarge provided and it does indeed seem to fix the DK bone displacement issue. I also tested Rogue Squadron and Indiana Jones (with CountPerOp=2) and it fixes the odd graphics in those games.

I tested save states as well and they seem to be functioning normally.

Here is a 64-bit Windows build of mupen64plus+GLideN64 if anyone would like to test this on Windows:
[mupen64plus-GLideN64-FGR-fix.zip](https://github.com/mupen64plus/mupen64plus-core/files/1218190/mupen64plus-GLideN64-FGR-fix.zip)

TODO:
I'm hoping @Gillou68310 can comment on this with his thoughts.
More work needs to be done to integrate this with the new dynarec.

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -3663,8 +3663,6 @@ SaveType=Eeprom 16KB
 Players=4
 Rumble=Yes
 CountPerOp=1
-; Bone displacement fix
-Cheat0=806128E2 0000
 
 [428067DC7DB42DFC977A775F0A6E55B1]
 GoodName=Donkey Kong 64 (E) [b1]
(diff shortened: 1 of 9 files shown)


AmbientMalice:
Being able to run some of these games at CF=2 would be a great help for low powered devices.

loganmc10:
See https://github.com/mupen64plus/mupen64plus-core/pull/368

PR #358, opened 2017-08-12
Fix RevA (U) Rogue Squadron

AmbientMalice:
Game needs CF=1. Will fix various issues, including low framerate.


loganmc10:
Thanks, I didn't even know there was a Rev A, do you know the differences? (Just curious)

AmbientMalice:
I'm not sure about any under the hood modifications, but they did place the secret Naboo Starfighter in the hangar, instead of hiding it behind a cheat code. (They hid that in the original version using some extremely complex code obfuscation.)

data/mupen64plus.ini
```diff
@@ -13107,6 +13107,14 @@ GoodName=Star Wars - Rogue Squadron (U) (M3) [t1]
 CRC=165CBDA2 05A62246
 RefMD5=47CAC4E2A6309458342F21A9018FFBF0
 
+[2E458D7CC355D7918493B0E0362C9A20]
+GoodName=Star Wars - Rogue Squadron (U) (M3) (V1.1) [!]
+CRC=C7F30CFA ECB0FA36
+Players=1
```
bsmiles32: Just adding the GoodName, CRC and
RefMD5=47CAC4E2A6309458342F21A9018FFBF0
should also work and avoid duplication.

loganmc10: @AmbientMalice in case that didn't make sense (I know the INI file works different than PJ64), you can just remove:
```
Players=1
SaveType=Eeprom 4KB
Rumble=Yes
CountPerOp=1
```
and add 
```
RefMD5=47CAC4E2A6309458342F21A9018FFBF0
```
and it will copy the settings from the current U entry for Rogue Squadron

AmbientMalice:
I did the same thing for the E version of Rogue Squadron.

PR #360, opened 2017-08-12
Rewrite PIF emulation.

bsmiles32:
Main changes in this partial rewrite are:
- channel processing is done in 2 separate steps. First is the channel
setup done when 0x1 is written in the "pif status byte". The second step
is the effective channel processing which allow each channel device to
answer. This step is done just before DMA reads. This is in accordance
with Patent 6394905. (Fig5B - Pif Macro). This allowed to greatly
simplify pif channel processing (no more strange bytes to parse in pif
ram and no duplicated code for that).
- "pif status byte" is treated as a bitfield with each flag indicating
which action to take.
- command processing interface has been reworked to separate the
transmitted data (Tx,TxBuf) from the the received data (Rx, RxBuf).
- reworked the code to allow (later) to have custom pif channel process
callback. This should allow to implement "raw" processor to let an
external device (such as a Raphnet controller) do the channel
processing.
- Validate pif commands
- Fixed eeprom write command which missed the status byte.

I tested several games without noticing any regression, but deeper testing is always welcome. Also, as implied by the patent SI DMA read seems to take longer than SI DMA write because PIF/Channel communication is done just before the DMA. This might be why we had to use the DelaySI hack.

As is, this PR might break the Raphnet input plugin, so more work is needed before merging it.


loganmc10:
When I boot Indiana Jones with this change I see:
```
Core Warning: Unexpected command format 21 fe 00
```
It still seems to work as normal. It's kind of a strange game. If CountPerOp=1 and CountPerScanline=1500 it won't boot at all, but it boots at CountPerOp=1 and CountPerScanline=2200, and it freezes at the end of each level.

loganmc10:
> DMA read seems to take longer than SI DMA write because PIF/Channel communication is done just before the DMA. This might be why we had to use the DelaySI hack

I'm not sure I understand, are you saying the DelaySI hack shouldn't be needed after this PR? I tested Body Harvest and it still seems to require it to boot.

bsmiles32:
DelaySI is still needed, but after reading carefully the patent it seems that dma read take longer because it waits for pif/channel communication to finish before doing the read. I didn't test but maybe having different timings for dma read (longer) and for dma write (shorter) could replace the DelaySI hack.
Also thanks for reporting the issue with Indiana Jones, I will check that.

loganmc10:
Just to be clear, I don't think the Indiana Jones issue was caused by this PR, there just wasn't any print out before, the game has always had issues.

bsmiles32:
Updated this PR with several new changes:
- Updated the controller input backend to split pak detection from controller connection test.
- Consolidated all input plugin compatibility code inside one module (eg. controller_input, rumble and pif processings functions).
- Added PIF channel printing debug
- When input plugin request RawData, we use the readController, controllerCommand functions instead of the built-in functions.

bsmiles32:
@loganmc10 Indiana Jones and Yoshi's Story have this strange 0x21 0xfe 0x00 command format Warning. From my little debugging, it seems that it's caused by a "bogus" CONT_PAK_READ request which has this extra 0x21 which screws PIF channels setup and cause the warning.

bsmiles32:
@raphnet In this PR I reworked how PIF emulation is done This has the potential to break existing plugins.
Could you take a look at this PR to see if breaks or not your plugin please ?

loganmc10:
I tested input and saving in Indiana Jones and Yoshi's Story, they both still work fine, so nothing has changed, it's nice to at least get a printout now when it's doing something strange.

raphnet:
@bsmiles32 As far as I can tell, everything still works fine. Tested the rumble pak and controller pak using a single player and a multi-player adapter. (Games used: Perfect Dark, 007 - The world is not enough, Mario Kart 64)

bsmiles32:
Thanks both of you for testing this ! I think it is in a mergeable state, but I will let @richard42 decides if he wants it to be part of the coming release.

loganmc10:
@bsmiles32 thanks for your explanation of how the PIF is supposed to work. I believe I figured out a way to remove the DelaySI hack:
https://github.com/loganmc10/mupen64plus-core/commit/fd304b064067edaf557d01bed3bb482b50b40c6c

Basically we set DMA busy for a write, RD busy for a read, and then delay the 2nd half of the process until right before the interrupt is signaled. I tested a number of games, including Body Harvest (which currently requires DelaySI to boot) and Mischief Makers (which requires DelaySI to be off for the input to work), they both work with this new method.

By the way, it's unclear to me why an interrupt is triggered in this case:
https://github.com/loganmc10/mupen64plus-core/blob/fd304b064067edaf557d01bed3bb482b50b40c6c/src/device/si/pif.c#L373-L374

I removed it and tested a few games, I didn't see any issue. After my change, ```process_pif_ram``` is called at the end of the DMA event, so it doesn't really make sense to me to immediately call another interrupt.

The way it currently works, ```process_pif_ram``` is always called right before ```schedule_si_interrupt```, so it seems like it would be redundant anyway.

So I didn't remove that interrupt scheduling, but I think it should be removed.

loganmc10:
@bsmiles32 I believe I found a bug in this PR. In Top Gear Rally, I get a "Controller Pak not found" message when the game boots. Doesn't happen in the current master. My DelaySI change doesn't make a difference either.

loganmc10:
Also by the way, when starting Pokemon Puzzle League I now see:

```
Core Warning: PIF - channel reset - not implemented !
Core Warning: PIF - channel reset - not implemented !
Core Warning: PIF - channel reset - not implemented !
Core Warning: PIF - channel reset - not implemented !
```

Doesn't seem to cause any issues though.

bsmiles32:
@loganmc10 Thanks for testing this PR more thoroughly. I was able to reproduce the issues you mentionned in this PR, but couldn't with my other WIP branch https://github.com/bsmiles32/mupen64plus-core/tree/pif_and_other_things . So I guess fixing these won't be too difficult.

bsmiles32:
- Updated against master
- Implements the pif channel reset byte (should fix 0xfd in Pokemon Puzzle league)
- Reduce the number of channels to 5 according to PIF pinouts
- Add a "hack" to discard bogus pak access commands. This is what fixes the non pak detection in Bomberman 64 and Top Gear Rally.

Please test and report so we can merge that and some more refactorings can be pushed :)

src/device/si/pif.c
```diff
@@ -80,6 +186,11 @@ void init_pif(struct pif* pif,
 {
     size_t i;
 
+    for(i = 0; i < PIF_CHANNELS_COUNT; ++i) {
+        init_pif_channel(&pif->channels[i], &pif_channel_devices[i]);
+
```
loganmc10: Extra space here

loganmc10:
Looks good to me, I tested using a controller to make sure rumble still works as well.

Narann:
Coverity have found some memory overflow. @bsmiles32 do you have access to coverity?

I put it here anyway:

```
New defect(s) Reported-by: Coverity Scan
Showing 2 of 2 defect(s)


** CID 173760:  Memory - corruptions  (OVERRUN)
/src/device/si/pif.c: 251 in setup_channels_format()


________________________________________________________________________________________________________
*** CID 173760:  Memory - corruptions  (OVERRUN)
/src/device/si/pif.c: 251 in setup_channels_format()
245            default: /* setup channel */
246   
247                /* HACK?: some games sends bogus PIF commands while accessing controller paks
248                  * Yoshi Story, Top Gear Rally 2, Indiana Jones, ...
249                  * When encountering such commands, we skip this bogus byte.
250                  */
>>>    CID 173760:  Memory - corruptions  (OVERRUN)
>>>    Checking "i + 1UL < PIF_RAM_SIZE" implies that "i" is 63 on the false branch.
251                if ((i+1 < PIF_RAM_SIZE) && (pif->ram[i+1] == 0xfe)) {
252                    ++i;
253                }
254                else {
255                    i += setup_pif_channel(&pif->channels[k++], &pif->ram[i]);
256                }

** CID 173759:  Control flow issues  (MISSING_BREAK)
/src/device/si/game_controller.c: 180 in process_controller_command()


________________________________________________________________________________________________________
*** CID 173759:  Control flow issues  (MISSING_BREAK)
/src/device/si/game_controller.c: 180 in process_controller_command()
174        uint8_t cmd = tx_buf[0];
175   
176        switch (cmd)
177        {
178        case PIF_CMD_RESET:
179            standard_controller_reset(cont);
>>>    CID 173759:  Control flow issues  (MISSING_BREAK)
>>>    The above case falls through to this one.
180        case PIF_CMD_STATUS: {
181            PIF_CHECK_COMMAND_FORMAT(1, 3)
182   
183            /* set device type and status */
184            rx_buf[0] = (uint8_t)(PIF_PDT_GAME_CONTROLLER >> 0);
185            rx_buf[1] = (uint8_t)(PIF_PDT_GAME_CONTROLLER >> 8);
```

loganmc10:
I'm having a hard time understanding what it's issue with the first thing is, but if the fallthrough in PIF_CMD_RESET is intentional, we probably just need to add a comment like this: https://github.com/snes9xgit/snes9x/commit/d441856303cfbcd19ca9c2cd73bdc8a8f412394a

That satisfies GCC, I assume it would satisfy Coverity as well, though I'm not sure

bsmiles32:
The fall through for reset is intentional because the reset command also report device status.
The other warning I don't understand.

richard42:
okay I see from here that it's intentional. I think the first warning about memory corruption is incorrect. as long as the setup_pif_channel only reads 1 byte at most from the &pif->ram[i] pointer, it should be fine.

fzurita:
So I finally got around to testing this. It seems to have broken the Android input plugin. I need to figure out what's up...

fzurita:
I'm trying to trace back why it's not working. It seems like ```setup_pif_channel``` in pif.c is never being called. I'll try to keep tracing back further.

PR #363, opened 2017-08-15
Round PI_WR_LEN_REG up to nearest even number

loganmc10:
This fixes AI Shogi 3 and other games.

Before:
before

After:
after

See:
https://github.com/project64/project64/blob/master/Source/Project64-core/N64System/Mips/Dma.cpp#L202-L203

### DIFF
--- a/src/device/pi/pi_controller.c
+++ b/src/device/pi/pi_controller.c
@@ -129,7 +129,7 @@ static void dma_pi_write(struct pi_controller* pi)
         return;
     }
 
-    longueur = (pi->regs[PI_WR_LEN_REG] & 0xFFFFFF)+1;
+    longueur = (pi->regs[PI_WR_LEN_REG] & 0xFFFFFE)+2;
     i = (pi->regs[PI_CART_ADDR_REG]-0x10000000)&0x3FFFFFF;
     longueur = (i + longueur) > pi->cart_rom.rom_size ?
                (pi->cart_rom.rom_size - i) : longueur;


src/device/pi/pi_controller.c
```diff
@@ -129,7 +129,7 @@ static void dma_pi_write(struct pi_controller* pi)
         return;
     }
 
-    longueur = (pi->regs[PI_WR_LEN_REG] & 0xFFFFFF)+1;
+    longueur = (pi->regs[PI_WR_LEN_REG] & 0xFFFFFE)+2;
```
richard42: Are you sure this is correct? I think the logic would make more sense if it were:
`longueur = ((pi->regs[PI_WR_LEN_REG] + 1) & 0xFFFFFE)`

loganmc10: Your method would round down. Before this PR, WR_LEN_REG would typically be an odd number, and we would add 1 to make it even. But, some games, like AI Shogi 3, write WR_LEN_REG as an even number, so we would add 1 and make it odd, no good.

The way it is now, if you WR_LEN_REG is 11 (for instance), it'll & it with FFFFFE, which makes 10, then add 2, which makes 12, so it's the same behaviour as before.

If WR_LEN_REG is 12, before, we would make it 13, but now we & it with FFFFFE, which makes 12, and then add 2, which makes 14. So we are "rounding" 13 up to 14. Your method would round 13 down to 12.

richard42: That is so odd. It just surprises me that the correct behavior would be to round down and then add 2. Perhaps it is supposed to round up to a 32-bit length? `((pi->regs[PI_WR_LEN_REG] + 3) & 0xFFFFFC)`

loganmc10: Well I can't really say what the correct behavior might be, I just know that this works. 1964 has an if statement like:

If number is odd: add 1

Which does the same thing, but I don't know how a real N64 deals with this situation

Issue #316, opened 2017-06-20
South park restarts (x86 fix not PIC complaint)


fzurita:
I'm not sure when these issues were introduced but he are a couple:

South Park 64: Game randomly resets, easy to reproduce, just start the game and try to kill something. It sounds game timing related.

Turok 1: This issue was reported to me. Getting shot at and dying causes the game to get stuck in a black screen.

loganmc10:
South Park: I could reproduce consistently with x86 new dynarec, but never with Cached Interpreter or x64 dynarec. So I'm assuming it's something to do with the new dynarec, I'll try and find the originating commit when I get time

Turok: Couldn't reproduce (used x86 new dynarec), I allowed myself to be shot and killed, I re-spawned and could keep playing, maybe a graphics plugin issue?

fzurita:
Yeah, Turok issue could be.

loganmc10:
The South Park issue can be hard for me to reproduce sometimes, it doesn't happen when I hit an enemy, it just happens sometime shortly after the game starts, but not always.

The earliest commit I could reproduce the issue on was c84be36e1f5a795d79e2883b2dd8010cde924a83, but it could be older than that.

loganmc10:
Ok I'm fairly confident the South Path issue originated with ec41a85876067993c9493d00dae879c2118bfed6 (PR #269). That is the first commit that I can reproduce the issue with. But again, only with the new dynarec (seems to affect arm and x86)

Gillou68310:
This should fix the issue:
https://github.com/Gillou68310/mupen64plus-core/commit/c0ace1d9716343ba09c149d7e0b047981093f6c8

loganmc10:
Yep that seems to fix it (tested x86 new dynarec)

fzurita:
Cool, that was fast.

bsmiles32:
Merged the changes by @Gillou68310 given the positive feedback.
However I fail to understand the core issue this solves, and why it was necessary modify mips_interpreter.def and x86{,_64}/dynarec.c files which where apparently not affected by this random reset.
@Gillou68310 : Could you elaborate more on this please :) ?

Gillou68310:
The main idea of my commit is to mask the address with 0xFFFFFFFC after tlb exception checks.
This way CP0_BADVADDR_REG will have the real address instead of the masked one in case of a tlb exception.
I'm actually surprised that it was only impacting the new dynarec!

fzurita:
[...]

PR #364, opened 2017-08-15
Get x86 new dynarec compiling again (when using PIC)

fzurita:
This allows the x86 new dynarec to compile again. I just switched the order of operations that the compiler was complain against and it allows it to build.

My x86 assembly skill is not strong at all, but you would think that the order of additions should not matter. Can anyone double check this and make sure that south park doesn't restart on its own anymore?

This is for this issue: https://github.com/mupen64plus/mupen64plus-core/issues/316

### DIFF
--- a/src/device/r4300/new_dynarec/x86/linkage_x86.asm
+++ b/src/device/r4300/new_dynarec/x86/linkage_x86.asm
@@ -488,7 +488,7 @@ read_byte_new:
     mov     ecx,    [find_local_data(g_dev_r4300_address)]
     and     ecx,    3
     xor     ecx,    3
-    movzx   eax,    BYTE [find_local_data(g_dev_r4300_new_dynarec_hot_state_rdword) + ecx]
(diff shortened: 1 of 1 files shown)


loganmc10:
I tested a few games in the new dynarec with this on Windows. No issues.

However, on PC it doesn't use PIC, so I can't say for sure how this would work on Android x86. I notice that the find_local_data() is different depending on the PIC setting

fzurita:
I'll dig out my Nexus player later tonight to make sure there are no issues there. I would really like if @Gillou68310 could comment since he made the original change.

fzurita:
I didn't see any issue in Android X86. Although, all the colors are currently messed up with GLideN64, sounds like something broke somewhere in the video plugin.

bsmiles32:
Merging this as it fixes the compilation and is trivial to revert if anything. Still if @Gillou68310 wants to comment on it, be my guest :)

PR #368, opened 2017-08-21
Disable exception handling if FR bit is set for some games

loganmc10:
I think this is a better solution compared to https://github.com/mupen64plus/mupen64plus-core/pull/357

If we were emulating the COUNT register properly, I don't think these issues would be happening. The previous solution was quite elaborate and applied to every game.

This still solves the bone displacement in DK, as well as the graphics in the Factor5 games, but it is much simpler and game specific.

I'll do some more testing in these games over the next few days. Hopefully somebody can figure out a better way to emulate the COUNT register, then maybe we could get rid of these many hacks.

Any comment from @Gillou68310 would be welcome as well.


Gillou68310:
Is it necessary to be a game specific hack? Did you notice any regression with other games?

Another idea would be to delay the interrupt until the FR bit is unset, instead of skipping the exception.

This hack https://github.com/mupen64plus/mupen64plus-core/pull/357 is incomplete and should not be used.

loganmc10:
I didn't notice any regressions in these games. I made it a game specific hack because I don't see any documentation stating that this is how the N64 worked. It's hard to say whether this might cause regressions in other games or not, since it might take some obscure situation to trigger the regression. I thought "better safe than sorry"

Gillou68310:
Ok for me ;-)
Still it would be interesting to test if delaying the interrupt is a possible solution too.

loganmc10:
I tried adding:

```
 if (cp0_regs[CP0_STATUS_REG] & CP0_STATUS_FR)
{
        uint32_t type = r4300->cp0.q.first->data.type;
        remove_interrupt_event(&r4300->cp0);
        add_interrupt_event(&r4300->cp0, type, 100);
        return;
}
```

Right before the switch statement in gen_interrupt. Rogue Squadron just locks up at boot. I might have done it wrong. Delaying the interrupt does seem like a decent solution, except that the interrupts might get out of order right? Like if we delay one, but not the others, then it seems to me that things would get out of order, unless there is a way to shift the whole interrupt queue. If there is I'm not aware of it.

Gillou68310:
In order to shift the whole interrupt queue you can use the ```translate_event_queue``` function.
Anyway let's use your current implementation for now we can come back to this later.

loganmc10:
Ok I think this can be merged then @Narann

theboy181:
@loganmc10 How can I reproduce this bone displacement in latest PJ64 ?

loganmc10:
https://github.com/project64/project64/issues/791

The first post has a link to PJ64 save states that will reproduce the issue.

You also need to remove the cheat based fix in the RDB https://github.com/project64/project64/blob/master/Config/Project64.rdb#L1481

Then you should be able to reproduce it

theboy181:
I get some weird blue flashing and audio issue until I pause, but no bone issues.. I have and had the cheat code in the RDB disabled too. Can you test? This method doesn't appear to work for me.

loganmc10:
One of the save states gave me the blue flashing, and the other gives the bone displacement as soon as you move. Actually I've only ever tried it in mupen64plus, the save states Isotarge provided are probably different for PJ64 and mupen64plus. I'll try PJ64 tomorrow. Did you try both save states?

theboy181:
I didn't see a second save state for PJ but the one link at the top does do a flashing blue screen and audio issue, but I assume that you would reproduce that from a save every time its loaded. I would like to test the bone displacement though.. where is the state?

loganmc10:
Sorry yeah I guess PJ64 just has that 1 state. @Isotarge might be a better person to ask about reproducing it, all I know is what was posted on that thread. There is also a pretty detailed Google doc explaining the issue

theboy181:
Thanks for the help, anyway. I think that the bug was squashed in PJ couple of months ago. Hope I'm not wrong.

Narann:
So, is it ok for merging? :)

Gillou68310:
> Project64 has made some progress here, their latest builds do not seem to suffer from bone displacement issues, even without the RDRAM patch in the RDB. The commit responsible for fixing this issue is possibly project64/project64@c7f8957

Well I think we should consider backporting the fix from project64 to see if it fixes the issue for us.

Narann:
PJ64 fix look nice, anyone would like to try it?

Gillou68310:
I'll take a look

Gillou68310:
Ok I backported the fix from project64:
https://github.com/Gillou68310/mupen64plus-core/commit/2cbd9bec9ce035d94f30da1a004097e6df5e08db

@loganmc10 could you test it?
Don't forget to remove the Bone displacement fix cheat code before testing ;-)

loganmc10:
Yes I tested DK with Isotarge's savestate, as well as Rogue Squadron and Indiana Jones, the issues are gone in all games, good work!

Gillou68310:
Cool :-)

loganmc10:
@Gillou68310 I assume you plan to submit a PR for that? Can I close this?

Gillou68310:
Sure
Here's the PR https://github.com/mupen64plus/mupen64plus-core/pull/390

theboy181:
Nice to see these type of bug fixes get across the scene so quickly.

PR #369, opened 2017-08-21
CountPerScanline=2200 for WDC

loganmc10:
During a race in World Driver Championship, you hear "garbage" sometimes, like a random sound you shouldn't hear. Setting CountPerScanline=2200 eliminates this for me, although it would be nice is someone else could test.

The in-race clock still seems to be pretty accurate with CountPerScanline=2200, it doesn't run too fast, but again, if someone could verify that'd be nice


loganmc10:
Actually I'm going to revisit this I think

PR #370, opened 2017-08-22
Base count_per_scanline on N64 clockrate

loganmc10:
Right now we default to 1500, which is a pretty close guess.

It's still not great though, for instance, the European version of Pokemon Puzzle League still has audio stuttering, but the USA version is fine. Increasing the count_per_scanline to 1600 fixes the European version of the game.

This change moves away from this "guesswork" system, and calculates the count_per_scanline based on the vi clockrate.

It works pretty well, the audio now works in both the USA and European version of Pokemon Puzzle League. In Resident Evil 2, the audio and video now stay in sync (with CountPerOp=1), without any changes to the scanline setting.

More testing is welcome. I've been testing quite a few games and see no regressions so far.

EDIT: For the curious, in NTSC games, this generally increases the count_per_scanline setting to ~1545, for PAL games it's ~1590

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -1552,7 +1552,7 @@ GoodName=Biohazard 2 (J) [!]
 CRC=7EAE2488 9D40A35A
 SaveType=SRAM
 Players=1
-CountPerScanline=2200
+CountPerOp=1
 
 [13D7834291A311077B84B9A2816AF6FC]
 GoodName=Birthday Demo for Steve by Nep (PD) [b1]
@@ -2854,14 +2854,12 @@ GoodName=Conker's Bad Fur Day (E) [!]
 CRC=373F5889 9A6CA80A
 SaveType=Eeprom 16KB
 Players=4
-CountPerOp=3
 
 [00E2920665F2329B95797A7EAABC2390]
 GoodName=Conker's Bad Fur Day (U) [!]
 CRC=30C7AC50 7704072D
 SaveType=Eeprom 16KB
 Players=4
-CountPerOp=3
 
 [DB7A03B77D44DB81B8A3FCDFC4B72D8C]
 GoodName=Cruis'n Exotica (U) [!]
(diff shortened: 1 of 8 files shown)


loganmc10:
@AmbientMalice I tested this change with Indiana Jones and I was able to play through the first 2 levels (including the trading post). I'll probably try it again tomorrow to be sure, I've been burned by that game before when I thought I had it working.

It only boots with CountPerOp=2, so I'm not sure how that might affect the rest of the game, the 1st 2 levels seemed normal to me.

AmbientMalice:
In my experience, =2 causes the game's physics to behave oddly. The game will temporarily freeze when Indy inflates the life raft in the third level, and the jeep will make a constant clunking noise and lurch around when you drive it.

Would you mind uploading a compiled build for me to test when I have time?

loganmc10:
It seems I spoke too soon again. I tested again on a different laptop today and I got through the 1st level, but it froze at the end of the second level. It must be some difference in what happens during the level or something. Very strange game...

Enabling DelaySI makes the first level work consistently for me, but breaks the 2nd level. Disabling DelaySI breaks the 1st level but allows me to complete the 2nd level. I wonder if shortening the DelaySI value would help. Anyway that is another problem for another day.

mudlord:
Have you tested timing dependant games like DK64?

loganmc10:
Yes, it's still the same, DK still misses the vines in the intro sequence. The song at the beginning still stays roughly in sync, but gets a bit worse towards the end. I assume more accurate emulation of the COUNT register will be needed to fix those issues.

The actual modification is rather small (1500 -> ~1545 in NTSC), but it's enough to fix some audio issues in a few games. Since the audio timing is dependent on the vi clock rate (https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/ai/ai_controller.c#L65) it makes sense to me to also use the vi clockrate here, instead of using the arbitrary 1500 number.

mudlord:
And COUNT register emulation would mean emulating cache behaviour so.......

loganmc10:
I haven't come across any regressions in this change and I believe it can be merged @bsmiles32 @richard42 @Narann

loganmc10:
In case anyone is ever reading through this stuff in the future... This also happened to fix: https://github.com/gonetz/GLideN64/issues/1549

PR #371, opened 2017-08-22
Have emulator signal RSP interrupts again

loganmc10:
This partially reverts https://github.com/mupen64plus/mupen64plus-core/pull/307

@Gillou68310 brought to my attention that those changes might not be quite right. The main difference here is that we unset TASKDONE/HALT/BROKE at the end of the RSP task, and set them when the interrupt is triggered. This should prevent the "doubling up" of RSP tasks that caused us issues before.

I tested Top Gear Rally 2, which used to freeze when an overlapping SP interrupt was called, it still seems fine. I also tested other "troublesome" games like Pokemon Snap, Kirby, World Driver Championship, they all seem to work well, but I'm hoping for some testing on this.

This also allows us to remove the "audio signal" hack that was added.

### DIFF
--- a/src/device/rsp/rsp_core.c
+++ b/src/device/rsp/rsp_core.c
@@ -113,11 +113,7 @@ static void update_sp_status(struct rsp_core* sp, uint32_t w)
 
     /* clear / set signal 0 */
     if (w & 0x200) sp->regs[SP_STATUS_REG] &= ~SP_STATUS_SIG0;
-    if (w & 0x400) {
-        sp->regs[SP_STATUS_REG] |= SP_STATUS_SIG0;
-        if (sp->audio_signal)
-            signal_rcp_interrupt(sp->r4300, MI_INTR_SP);
-    }
+    if (w & 0x400) sp->regs[SP_STATUS_REG] |= SP_STATUS_SIG0;
 
     /* clear / set signal 1 */
     if (w & 0x800) sp->regs[SP_STATUS_REG] &= ~SP_STATUS_SIG1;
@@ -158,13 +154,11 @@ static void update_sp_status(struct rsp_core* sp, uint32_t w)
 void init_rsp(struct rsp_core* sp,
               struct r4300_core* r4300,
               struct rdp_core* dp,
-              struct ri_controller* ri,
-              uint32_t audio_signal)
+              struct ri_controller* ri)
 {
     sp->r4300 = r4300;
     sp->dp = dp;
     sp->ri = ri;
-    sp->audio_signal = audio_signal;
 }
 
 void poweron_rsp(struct rsp_core* sp)
@@ -271,6 +265,8 @@ void do_SP_Task(struct rsp_core* sp)
 {
     uint32_t save_pc = sp->regs2[SP_PC_REG] & ~0xfff;
 
+    uint32_t sp_delay_time;
+
     if (sp->mem[0xfc0/4] == 1)
     {
         if (sp->dp->dpc_regs[DPC_STATUS_REG] & DPC_STATUS_FREEZE) // DP frozen (DK64, BC)
@@ -290,11 +286,13 @@ void do_SP_Task(struct rsp_core* sp)
         sp->regs2[SP_PC_REG] |= save_pc;
         new_frame();
 
-        if (sp->r4300->mi.regs[MI_INTR_REG] & MI_INTR_DP) {
+        if (sp->r4300->mi.regs[MI_INTR_REG] & MI_INTR_DP)
+        {
             cp0_update_count(sp->r4300);
-            add_interrupt_event(&sp->r4300->cp0, DP_INT, 4000);
+            add_interrupt_event(&sp->r4300->cp0, DP_INT, 1000);
             sp->r4300->mi.regs[MI_INTR_REG] &= ~MI_INTR_DP;
         }
+        sp_delay_time = 1000;
 
         protect_framebuffers(sp->dp);
     }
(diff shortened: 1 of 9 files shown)


Gillou68310:
When returning from an audio task we were also previously clearing the SP_STATUS_YIELDED flag, probably because an audio task can't be yielded.

loganmc10:
I saw that, if an audio task can't be yielded, then wouldn't the SP_STATUS_YIELDED flag never be set to begin with? As long as there isn't some bug in the RSP plugin, I would think it isn't necessary to clear it. In PR #307 that was removed, and I haven't seen or heard any issues to do with that.

Gillou68310:
Let's remove it!

loganmc10:
By the way, this PR, combined with https://github.com/mupen64plus/mupen64plus-core/pull/370, allows Perfect Dark to boot in LLE for me. The weird thing is, each PR on it's own doesn't seem to fix the issue, but using both at the same time allows it to work in LLE (it hasn't worked in LLE forever, it doesn't work in PJ64 either). So hopefully this is a good sign that we are moving towards more accurate timing

src/device/rsp/rsp_core.c
```diff
@@ -306,27 +304,45 @@ void do_SP_Task(struct rsp_core* sp)
...
     {
         sp->regs2[SP_PC_REG] &= 0xfff;
         rsp.doRspCycles(0xffffffff);
         sp->regs2[SP_PC_REG] |= save_pc;
+
+        sp_delay_time = 0;
     }
 
     sp->rsp_task_locked = 0;
     if ((sp->regs[SP_STATUS_REG] & (SP_STATUS_HALT | SP_STATUS_BROKE)) == 0)
```
Gillou68310: @loganmc10 I noticed that this is triggered in two different situations:
1) RSP is locked in an idle loop
2) RSP set the MI_INTR_SP flag via an MTC0 instruction

I think only the second case should schedule an SP interrupt!
In the first case we should allow the CPU to run for some time without triggering an SP interrupt. Maybe we can create a new dummy interrupt type?

I'll try to propose an implementation based on this commit.

loganmc10: If you unset the INTR_ON_BREAK flag then it won't raise an interrupt, maybe that is a solution in that case? I don't think you would need a second "dummy interrupt" timer

loganmc10: Also, you're sure that the RSP exits with halt=0 if it sets MI_INTR_SP in MTC0? I don't recall seeing that in cxd4's RSP

Gillou68310: yep tested with wdc

Gillou68310: Well actually I like the dummy interrupt idea because it will be persisted into savestates contrary to the rsp_task_locked flag!

loganmc10: oh yeah if you can get rid of the rsp_task_locked flag that would be nice, savestates don't work as you mention. If you figure it out I would submit a different PR though, it's sort of a separate issue from this PR.

Gillou68310: Yep totally different issue it's related to my own commit https://github.com/mupen64plus/mupen64plus-core/pull/257/commits/e11685ccc16c49658917a74c9a7d6ebb3020a5ae

Gillou68310: @loganmc10 could you test this https://github.com/Gillou68310/mupen64plus-core/commit/87f9186fb3007f5f4446354442344a7363155219 for regressions?

loganmc10: I tested WDC, Stunt Racer and Gauntlet. They all ran, save states worked sometimes, but still not reliably. Also, there was some graphical issues I saw in Stunt Racer, not present before this change (see the blue box under "Soda Mtn"). I noticed you changed the SP_LOCKED time to 2000. I assume fiddling with that number might solve the issue, I tried making it 1000 but then the games no longer got past the title screen.

stunt_racer_64-001

Gillou68310: Damn this is touchy! Using 1500 fixes the flickering blue box.

loganmc10:
Hold off on this for now, @Gillou68310 I just experienced
```
Core Warning: two events of type 0x100 in interrupt queue
Core Warning: two events of type 0x80 in interrupt queue
```
In Perfect Dark while the N64 logo is spinning, and then game froze. This is the same problem we had before. Interestingly, it also freezes in the same spot using LLE. I could only get it to do it once, but it does happen more frequently in LLE (it happens in LLE with or without this PR). I need to look into this a bit more.

loganmc10:
Ok I fixed that problem by doing this:
https://github.com/mupen64plus/mupen64plus-core/pull/371/commits/8bbbde9c94465a2df2f4af1cc733a97fce747fb4

I tested Blast Corps and DK64 in LLE, which are both games that go berserk without this DP freeze hack in place, and they both still work. Perfect Dark works again in HLE, but still not LLE (it never has, the DP freeze hack is to blame, disabling it makes Perfect Dark work in LLE).

So I think this is good to merge. That whole DP freeze thing is an unfortunate hack since the RSP and CPU don't run in parallel in the emulator, but it does seem to work for everything except Perfect Dark LLE

AmbientMalice:
@loganmc10 You should probably disable the DP freeze hack for Perfect Dark like libretro has done.

loganmc10:
I'm really trying to get rid of game specific hacks. We know the cause of that problem and it's very easy to reproduce, so it's a "prime candidate" for a proper fix. I'm worried that adding a game specific hack would remove all motivation to find a proper fix, and that game emulates very well in HLE, so that is a pretty good "workaround" for now.

loganmc10:
Ok I pushed a generic fix for Perfect Dark in LLE, I tested the first level in HLE and LLE, got all the way through and into the 2nd level (I think it used to freeze when you hit "Accept" at the end of the level, happened in PJ64 too, but I haven't experienced that in a while)

Gillou68310:
> Hold off on this for now, @Gillou68310 I just experienced
> 
> Core Warning: two events of type 0x100 in interrupt queue
> Core Warning: two events of type 0x80 in interrupt queue
> 
> In Perfect Dark while the N64 logo is spinning, and then game froze. This is the same problem we had before. Interestingly, it also freezes in the same spot using LLE. I could only get it to do it once, but it does happen more frequently in LLE (it happens in LLE with or without this PR). I need to look into this a bit more.

@loganmc10 could you try reducing the SP interrupt delay to see if it helps with this issue?

loganmc10:
I fixed it with this commit: https://github.com/mupen64plus/mupen64plus-core/pull/371/commits/49f9d387ebd08e7f8306ab1dd58536a4246ef7b8

Reducing the interrupt timer might fix it, but I don't really like counting on that. There could be other games that have the same issue but need an even shorter timer, or the 3rd level in Perfect Dark, etc... I'd rather have a "bullet proof" fix

loganmc10:
This has also been tested quite a bit and can be merged I believe @Narann @bsmiles32 @richard42

Narann:
Merge conflict after merging #370

loganmc10:
Merge conflict has been resolved

PR #373, opened 2017-08-23
Remove Rat Attack hack

loganmc10:
I think it was a mistake for me to submit this hack earlier. It's not good to add an extra if statement every time ```virtual_to_physical_address``` is called just so this game can (sort of) work.

It seems like some kind of strange timing issue, so hopefully as we get more accurate timing this problem will sort itself out.

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -11695,7 +11695,6 @@ CRC=20FD0BF1 F5CF1D87
 Players=4
 Rumble=Yes
 SaveType=Controller Pack
-CountPerOp=1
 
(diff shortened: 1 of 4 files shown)


loganmc10:
This will "break" Rat Attack, although it never worked 100% with this hack. I think removing this is a good idea and hopefully with better timing in the future this game will start to work. It can be merged

Gillou68310:
@loganmc10 you forgot to remove the hack from the new dynarec

loganmc10:
Whoops sorry https://github.com/mupen64plus/mupen64plus-core/pull/387 @Narann

Issue #154, opened 2016-04-01
Paper Mario (U) flash written incorrectly


roothorick:
The libretro guys observed this issue a good two years ago: https://github.com/libretro/mupen64plus-libretro/issues/63 Seeing the same symptoms in M64Py 2.3; from a quick skim of the commit log I didn't see any attempt to directly address it.

By all appearances ingame: when game-save data is written, a copy is written in all subsequent slots i.e. saving in slot 2 replicates the data to 3 and 4. The errant duplicates appear after returning to the title screen and reopening the load menu. Trying to copy saves has the same effect, starting from the slot copied to. Bizarrely, deleting a save doesn't behave the same way.

Seen claims that Project64 shows the same issue. Makes me wonder if the game had an unusual flash RAM setup that nobody noticed.

loganmc10:
@roothorick I finally figured out a solution to this problem:

https://github.com/mupen64plus/mupen64plus-core/pull/375

PR #375, opened 2017-08-25
Fix duplicate saving in Paper Mario

loganmc10:
This fixes https://github.com/mupen64plus/mupen64plus-core/issues/154

Right now, if the PI controller gets a strange request, we basically just ignore it and hope for better luck next time.

What this change does: If the rom address is too large, or if the rom address is good, but the (rom_address + length) would put us outside the ROM's size, then we copy what fits, and fill the rest with 0's.

I'll do some more testing in the next day or two to make sure this doesn't cause any regressions in other games.

### DIFF
--- a/src/device/pi/pi_controller.c
+++ b/src/device/pi/pi_controller.c
@@ -116,47 +116,32 @@ static void dma_pi_write(struct pi_controller* pi)
         return;
     }
 
-    /* XXX: why need special treatment ? */
-    if (pi->regs[PI_CART_ADDR_REG] >= 0x1fc00000) // for paper mario
-    {
-        /* mark DMA as busy */
-        pi->regs[PI_STATUS_REG] |= PI_STATUS_DMA_BUSY;
-
-        /* schedule end of dma interrupt event */
-        cp0_update_count(pi->r4300);
-        add_interrupt_event(&pi->r4300->cp0, PI_INT, 0x1000); /* XXX: 0x1000 ??? */
-
-        return;
-    }
-
     longueur = (pi->regs[PI_WR_LEN_REG] & 0xFFFFFE)+2;
-    i = (pi->regs[PI_CART_ADDR_REG]-0x10000000)&0x3FFFFFF;
-    longueur = (i + longueur) > pi->cart_rom.rom_size ?
-               (pi->cart_rom.rom_size - i) : longueur;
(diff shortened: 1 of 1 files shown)


loganmc10:
My testing of this PR hasn't revealed any issues with any other games

Narann:
Good job @loganmc10 !

This is not a tiny change, I would wait for @bsmiles32 and @Gillou68310 to check this before merge.

bsmiles32:
Not sure what is the real behavior on the console, but your changes seems ok. I tested with Paper Mario, did not notice anything strange, but also I wasn't able to reproduce the issue before, so I will just trust you that it indeed fixes the saving issue.

Merging this. Good work !

loganmc10:
@bsmiles32 somebody in the libretro thread reported that it doesn't happen in the PAL version, I'm assuming you were probably testing that version.

bsmiles32:
Indeed :)

src/device/pi/pi_controller.c
```diff
@@ -116,47 +116,32 @@ static void dma_pi_write(struct pi_controller* pi)
...
     {
-        dram[(dram_address+i)^S8] = rom[(rom_address+i)^S8];
+        int32_t diff = pi->cart_rom.rom_size - rom_address;
+        if (diff < 0) diff = 0;
+
+        for (i = 0; i < diff; ++i)
+        {
+            dram[(dram_address+i)^S8] = rom[(rom_address+i)^S8];
+        }
+        for (i = diff; i < longueur - diff; ++i)
```
richard42: I think this should be `for (; i < longueur; ++i)`

loganmc10: hm I think you are right, I don't know what I was envisioned when I wrote that. I tested Paper Mario and it's still fixed. Do you want me to submit another PR?

richard42: no it's okay I'll just push a fix from here.

Fad1000:
> This fixes #154
> 
> Right now, if the PI controller gets a strange request, we basically just ignore it and hope for better luck next time.
> 
> What this change does: If the rom address is too large, or if the rom address is good, but the (rom_address + length) would put us outside the ROM's size, then we copy what fits, and fill the rest with 0's.
> 
> I'll do some more testing in the next day or two to make sure this doesn't cause any regressions in other games.

What am I supposed to do? #

PR #376, opened 2017-08-29
[WIP] Fix FBInfo

loganmc10:
This makes the FBInfo API work as well as it does with Mupen64. It's still not perfect (doesn't work with Banjo-Kazooie or Pokemon Snap, but neither does Mupen64).

I believe there are some writes (maybe reads) in the PI controller that need to be tracked using ```pre_framebuffer_write/read```, but I can't get it working quite right. It seems to work in 1964 using GLideN64, at least the puzzle transition in Banjo does. The camera in Pokemon Snap doesn't seem to work in 1964 so I can't see if the red dot is there.

I'll keep working on this to see if I can get Pokemon Snap and Banjo working

***TODO***:

- Get it working with new dynarec
- Waiting for GLideN64 to fix FBInfo on it's side, so this can be properly tested
- Potentially need to add some ```pre_framebuffer_write/read``` to PI DMA
- Would like to have a way to communicate to the graphics plugin that FBInfo has been fixed. My idea is to write a 1 to ```fb->infos[0].addr``` before the first call to ```gfx.fBGetFrameBufferInfo``` this can be a signal to the GFX plugin, a sort of "version number", telling it that the FBInfo implementation has changed/is working.

### DIFF
--- a/src/device/rdp/fb.c
+++ b/src/device/rdp/fb.c
@@ -34,45 +34,56 @@ void poweron_fb(struct fb* fb)
 {
     memset(fb, 0, sizeof(*fb));
     fb->once = 1;
+    fb->dirty_page_counter = 0;
 }
 
 
 static void pre_framebuffer_read(struct fb* fb, uint32_t address)
 {
-    size_t i;
+    if (!fb->infos[0].addr) return;
+
+    size_t i, j;
 
     for(i = 0; i < FB_INFOS_COUNT; ++i)
     {
         if (fb->infos[i].addr)
         {
-            unsigned int start = fb->infos[i].addr & 0x7FFFFF;
+            unsigned int start = fb->infos[i].addr;
             unsigned int end = start + fb->infos[i].width*
                                fb->infos[i].height*
                                fb->infos[i].size - 1;
-            if ((address & 0x7FFFFF) >= start && (address & 0x7FFFFF) <= end &&
-                    fb->dirty_page[(address & 0x7FFFFF)>>12])
+            if (address >= start && address <= end)
             {
+                for (j = 0; j < fb->dirty_page_counter; ++j)
+                {
+                    if (address >= fb->dirty_page[j] && address < fb->dirty_page[j] + 0x1000)
+                        return;
+                }
                 gfx.fBRead(address);
-                fb->dirty_page[(address & 0x7FFFFF)>>12] = 0;
+                if (fb->dirty_page_counter == FB_DIRTY_PAGES_COUNT)
+                    fb->dirty_page_counter = 0;
+                fb->dirty_page[fb->dirty_page_counter++] = address;
             }
         }
     }
 }
 
-static void pre_framebuffer_write(struct fb* fb, uint32_t address)
+static void pre_framebuffer_write(struct fb* fb, uint32_t address, uint32_t length)
 {
+    if (!fb->infos[0].addr) return;
+
     size_t i;
 
     for(i = 0; i < FB_INFOS_COUNT; ++i)
     {
         if (fb->infos[i].addr)
         {
-            unsigned int start = fb->infos[i].addr & 0x7FFFFF;
+            unsigned int start = fb->infos[i].addr;
             unsigned int end = start + fb->infos[i].width*
                                fb->infos[i].height*
                                fb->infos[i].size - 1;
-            if ((address & 0x7FFFFF) >= start && (address & 0x7FFFFF) <= end)
-                gfx.fBWrite(address, 4);
+            if (address >= start && address <= end)
+                gfx.fBWrite(address, length);
         }
     }
 }
@@ -87,7 +98,7 @@ void read_rdram_fb(void* opaque, uint32_t address, uint32_t* value)
 void write_rdram_fb(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
 {
     struct rdp_core* dp = (struct rdp_core*)opaque;
-    pre_framebuffer_write(&dp->fb, address);
+    pre_framebuffer_write(&dp->fb, address, 4);
     write_rdram_dram(dp->ri, address, value, mask);
 }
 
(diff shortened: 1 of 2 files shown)


AmbientMalice:
It would be ideal if the stability issue in Pokemon Snap could be fixed because that's one of the use cases where this feature would be invaluable.

loganmc10:
Ok I figured out the issue. I moved unprotect/protect framebuffers to when VI_ORIGIN_REG is read, this is where 1964 does it:

https://github.com/LegendOfDragoon/1964-X/blob/7374fb0e5f175ecc2f2acfa089d0814bc7d99ae1/src/core/R4300i.cpp#L3474-L3485

Besides that, the only issue was the "dirty page" tracking, FBRead's should be ignored if the same 4K block has already been read. That system was broken, so I re-wrote it.

I tested various games, Banjo, Mario Kart, Dr Mario, they all look good now in GLideN64.

Could use some more testing, in GLideN64 you set "DisableFBInfo" to false to use this.

The red dot in Pokemon Snap is still missing. It only appears if CopyColorfromRDRAM is enabled, but then the screen flickers. Pokemon Snap doesn't work in 1964, and it freezes in Mupen64 when using FBInfo, so there is no other case to test against to see if it would work, so I assume it's working as it should now

fzurita:
Cool, nice work. Is this working with the new ARM dynarec? From my inspection of the code a while back, it seemed like it completely bypassed the mechanism that would allow FBInfo to work.

loganmc10:
lol I just realized that all my change did was completely disable FBInfo, so I need to keep digging... To be honest I don't really see the benefit of this API. It doesn't actually increase compatibility, or allow previously un-rendered effects to render.

It's just a speed thing, to try and notify the graphics plugin which parts of which framebuffers have been modified by the CPU. I haven't seen any real concrete evidence that it provides a noticeable speed boost. I'm starting to think we should just remove it, GLideN64 can render all these effects without the help of this API.

The only real benefit I could possibly see is that it could reduce the amount of data copied over via glReadPixels, since it can just read a chunk of the framebuffer instead of the whole thing, but even that is disabled by default:

> Read color buffer by chunks. When enabled, plugin follows FBInfo specification: "Notify the dll that the frame buffer memory is beening read at the given address. ... DLL should copy 4KB block content back to RDRAM frame buffer." Plugin will read only chunks of data, explicitly requested by emulator. It may reduce overheads if only few 4KB chunks have to be read. If CPU wants to modify whole buffer, read by chunks will be much slower. Since CPU usually needs the whole color buffer, that option is off by default.

fzurita:
This API is good for cases where async copies to RDRAM break certain game effects. Synchronous copies to RDRAM cause a very large slow down.

loganmc10:
But how does this API make a difference? The API just uses synchronous copies to copy the data, whether it's a chunk, or the whole thing (which is the default).

I guess I got the Mario Kart screen working, so I can test the difference in speed there, but I don't really see how it could make things faster

fzurita:
It makes things faster because it will only make the video plugin do a glReadPixels when the data is actually needed and not every frame.

loganmc10:
Well I feel really stupid/not so stupid. I've been testing 1964, since I heard that's the only emulator that supports this, but I didn't realize that FBInfo is disabled by default on the emulator side. When I enable "Frame Buffer R/W", then I have the exact same issue in Banjo and Pokemon Snap that I've been having here.

So as of right now I don't think there are any issues with this implementation, it seems like FBInfo in GLideN64 is broken or something

loganmc10:
> It makes things faster because it will only make the video plugin do a glReadPixels when the data is actually needed and not every frame.

Thanks for explaining, hopefully it can get sorted out in the GLideN64 side.

As for your question about the new dynarec, no it probably won't work right now. It probably won't require much to make it work, I'll just have to wrap some alternate code inside ```ifdef NEW_DYNAREC``` that uses read_rdram_new/write_rdram_new instead of the regular one. Once it's fixed on GLideN64's side and I can actually test and make sure everything is working I'll do that part.

AmbientMalice:
@loganmc10 Be aware that vanilla 1964 has a broken implementation that slows to a crawl in games like Mario Kart.

loganmc10:
Yeah that shouldn't happen here. I've got it so it doesn't do duplicate reads, that is the part that was missing in 1964.

AmbientMalice:
>To be honest I don't really see the benefit of this API. It doesn't actually increase compatibility, or allow previously un-rendered effects to render.

It allows the plugin to render above native resolution while rendering CPU framebuffer effects. This is extremely problematic in games like Jet Force Gemini, and of course Pokemon Snap. You just want to read the red pixels from the centre of the screen. You don't want to copy the entire framebuffer and paste it on the screen.

mudlord:
which would be expensive, especially at 1080p and above.

AmbientMalice:
Would it be possible to get a build with this enabled for testing? I'd like to see how things are working with GLideN64 now.

loganmc10:
@AmbientMalice gonetz posted a zilmar-spec build here: https://github.com/gonetz/GLideN64/commit/58636fcee6894cf68a43655099d26c25e2de09b4#commitcomment-25299034

the mupen64plus implementation isn't really ready for testing yet. The Banjo puzzle transition still doesn't work, and Mario Kart crashes, both issues that 1964 doesn't have. Now that we have a better implementation in GLideN64, I'm going to work on improving the mupen64plus side of things

loganmc10:
I've messed around with this quite a bit, looking at 1964's implementation. I can't get the puzzle transition in Banjo Kazooie to work properly.

I'm pretty much out of ideas so I'm going to close this. Someone else can give another stab at it in a different PR if desired. The implementation in GLideN64 still has some issues, Mario 64 freezes right at the intro for instance, so there are other issues as well. I think some PR recently also broke FBInfo with the dynarec, it only seems to work with the interpreter now (crashes with dynarec for me)

Issue #377, opened 2017-08-30
Travis CI broken


loganmc10:
@charlemagnelasse maybe you could take a look? It's complaining about some apt dependencies

charlemagnelasse:
@loganmc10 did you try to restart the build for this pull request?
@richard42  or @bsmiles32  can you please restart the build using the https://travis-ci.org/mupen64plus/mupen64plus-core/builds/246039941

It is shown here how to do that: https://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit#answer-17624403

loganmc10:
I can restart them, the problem still exists

charlemagnelasse:
Thanks for testing. It looks like their servers are still broken. You can try the workaround from #380. It replaces the apt-get calls with the apt travis ci addon. The installed packages are still the same.

PR #380, opened 2017-09-01
Add missing dependencies

BanzaiMan:
Resolves https://github.com/mupen64plus/mupen64plus-core/issues/377.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ compiler:
   - clang
 before_install:
   - sudo apt-get update -qq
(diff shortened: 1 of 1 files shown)


loganmc10:
I don't know why it doesn't install the dependencies automatically, but it does seem to fix it since the Travis build succeeded, thanks

charlemagnelasse:
It is a rather bad idea to install stuff which are dependencies for packages that are required by packages used by mupen64plus - *BUT NOT* actual dependencies for mupen64plus-core

charlemagnelasse:
Please reject this pull request. https://github.com/mupen64plus/mupen64plus-core/pull/381 shows that apt-get on their servers is broken and currently makes problems. Adding these packages is not required to work around this problem.

Issue #304, opened 2017-05-31
Pokemon Puzzle League New Dynarec assert


loganmc10:
```
mupen64plus: ../../src/device/r4300/new_dynarec/new_dynarec.c:3335: alu_assemble: Assertion `s2h>=0' failed.
```

loganmc10:
This was on a Raspberry Pi

Gillou68310:
@bsmiles32 could you also assign me to this issue? ;-)

loganmc10:
I just tested this on Android (arm) and x86 and couldn't reproduce it. I suppose it's possible that I was using a debug build on my Raspberry Pi, and that allowed the assert to be triggered, but I don't really use my Raspberry Pi anymore so I can't be sure. In any event, it seems to work fine in a release build

Gillou68310:
I worked on this issue yesterday and there is really a bug here, so this should stay open!

loganmc10:
Oh ok my bad, I guess an assert does get triggered, it must be arm specific then, I didn't run across it in x86 with asserts enabled

Gillou68310:
Yep it's arm specific ;-)

Gillou68310:
@fzurita @loganmc10 I can not reproduce this issue with latest master. Could you guys double check?

fzurita:
Yep, not crashing any more. Unintended bug fix?

Gillou68310:
lol yeah

Gillou68310:
ok closing this

Issue #288, opened 2017-05-24
Indiana Jones x86 new dynarec crash


loganmc10:
Right after starting the first level:

```
../../src/device/r4300/new_dynarec/new_dynarec.c:4005: load_assemble: Assertion `th>=0' failed.
```

loganmc10:
x64 old dynarec seems to be fine

fzurita:
Happens with ARM as well.

Gillou68310:
@bsmiles32 could you also assign me to this issue? ;-)

Gillou68310:
https://github.com/mupen64plus/mupen64plus-core/pull/383

@loganmc10, @fzurita could you guys check my fix for regressions ;-)

fzurita:
Sure! Probably some time tonight.

Issue #305, opened 2017-06-01
Worms Armageddon new dynarec crash


loganmc10:
On x86 new dynarec:

```
../../src/device/r4300/new_dynarec/new_dynarec.c:9733: new_recompile_block: Assertion `(branch_regs[k].is32&(1LL<<(f_regmap[hr]&63)))== (regs[k+2].was32&(1LL<<(f_regmap[hr]&63)))' failed.
```

Gillou68310:
@bsmiles32 could you also assign me to this issue? ;-)

bsmiles32:
@Gillou68310 my pleasure 😀

Gillou68310:
@loganmc10  is this still an issue with #871?

loganmc10:
Fixed, closing so I dont forget

Issue #286, opened 2017-05-23
FIFA 99: Players are floating (x64 Dynarec)


loganmc10:
Works fine with the Cached Interpreter. When I tried the x86 new dynarec, it crashed, still need to figure out where it crashed.

![fifa_99-003](https://cloud.githubusercontent.com/assets/848146/26377665/7548574e-3fce-11e7-8493-bebd5d92529d.png)
![fifa_99-007](https://cloud.githubusercontent.com/assets/848146/26377700/a08d2f38-3fce-11e7-9968-9695e8e97977.png)

loganmc10:
Here is the assert in the new dynarec @Gillou68310 :

```
mupen64plus: ../../src/device/r4300/new_dynarec/new_dynarec.c:6356: cjump_assemble: Assertion `s1h>=0' failed.
```

Gillou68310:
Thanks for opening the issue I don't have time to look at it right now but @bsmiles32 could you assign this issue to me so I can remember it?

loganmc10:
This problem doesn't occur in the x86 "old" dynarec.

Enableding ```INTERPRET_ADDU``` fixes the problem in the x86 dynarec. I'm not really an assembly code guy, but it looks like the problem is with the ADDU instruction.

DaveGravel:
Here I get this issue without the new dynarec.
If I enable the new dynarec it look to work correctly.
You can see here in this video, I use the 64bits version:
https://youtu.be/_P8pzxo0YFk?t=90

Gillou68310:
@loganmc10 should be fixed by #871

loganmc10:
Yes looks good, I'll close this so I don't forget

PR #383, opened 2017-09-01
New dynarec fixes [WIP]

Gillou68310:
https://github.com/mupen64plus/mupen64plus-core/issues/288

### DIFF
--- a/src/device/r4300/new_dynarec/new_dynarec.c
+++ b/src/device/r4300/new_dynarec/new_dynarec.c
@@ -3169,6 +3169,8 @@ static void alu_assemble(int i,struct regstat *i_regs)
           assert(s1l>=0);
           assert(s2l>=0);
           if(th>=0) {
+            assert(s1h>=0);
+            assert(s2h>=0);
             #ifdef INVERTED_CARRY
             if(opcode2[i]&2) emit_sub64_32(s1l,s1h,s2l,s2h,tl,th);
             #else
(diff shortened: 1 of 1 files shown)


loganmc10:
Just tested this (including the WDC fix). WDC/Stunt Racer and Indiana Jones are working now, however Smash Bros no longer boots.

Gillou68310:
Thanks could you check which one of the 2 commits broke smash bros? I guess it would be the first one. Did you tested on arm or x86?

loganmc10:
I do that in a bit, I tested x86 (windows)

Gillou68310:
Ok my fix for Indi on x86 is wrong. I can't force the shift value to be allocated to ECX because it's likely to break the register mapping when in a delay slot. I need to find another workaround!

Gillou68310:
However my fix for ARM should be ok!

Gillou68310:
@loganmc10 I updated my fixes, could you check again for regressions?

loganmc10:
Smash Bros works now. Pokemon Puzzle League and both Banjo games now fail at the assert(0) in MULTDIV. Pokemon Puzzle League used to fail at a different assert (https://github.com/mupen64plus/mupen64plus-core/issues/304), but I believe the Banjo games used to work.

Gillou68310:
Oups sorry I forgot to remove it. You can safely remove it if you want to do more tests ;-)

loganmc10:
Ok yeah removing that assert allows those 3 games to work normally.

I just ran across an issue in Conkers though:

```
Assertion failed!
File: ../../src/device/r4300/new_dynarec/x86/assem_x86.c, Line 777

Expression: rm<8
```

Doesn't happen in the current master, this is with x86

Gillou68310:
Thanks I'll investigate the issue!

Gillou68310:
@loganmc10 I updated the fix for WDC. It should fix the issue in CBFD.

loganmc10:
Yep looks good to me now. I did all my testing on x86, not arm

fzurita:
Ok, sorry for the delay. I just tried Indiana Jones with ARM Android, it's not crashing where it used to crash any more. I also tried several other games as sanity checks with no issue.

This looks good to me,

Gillou68310:
Thanks for testing!
I just pushed a fix for Glover BTW ;-)

loganmc10:
Does that fix perhaps also fix https://github.com/mupen64plus/mupen64plus-core/issues/304? I notice it modifies the assert that is failing in that issue

Gillou68310:
I didn't check but last time I invatigated this issue I think the problem was elsewhere. I need to double check!

fzurita:
Nice work, I'll try to test it soon.

fzurita:
I can confirm that the glover fix works correctly. I did some sanity checks with a few common games and nothing seems to break.

Narann:
@fzurita could you rebase on master so the TravisCI doesn't bug anymore.

Gillou68310:
Done

Narann:
It's passing the build. Is this ready for merge? As it's not that much, I would have loved to have @bsmiles32 view on this.

Gillou68310:
Nope it's not ready yet, I'm still working on the following issues:
https://github.com/mupen64plus/mupen64plus-core/issues/286 https://github.com/mupen64plus/mupen64plus-core/issues/304 https://github.com/mupen64plus/mupen64plus-core/issues/305

Also the fix for Glover might change because it accidentally fix https://github.com/mupen64plus/mupen64plus-core/issues/304 while it's not the root cause of the issue.

https://github.com/mupen64plus/mupen64plus-core/issues/385 will be for another PR

loganmc10:
I would suggest perhaps merging this PR, and then submitting a new PR for the future fixes. The current fixes have been tested and we know they fix these games.

Gillou68310:
The fix for glover is not mergeable yet but the fix for WDC and indi could definitely be merged. Since I'm busy with other things right now I think it's a good idea to merge the 2 first commits until I come back to this!

Narann:
@Gillou68310 You should have merge permission. Haven't you?

Gillou68310:
Ok I pushed the fix for WDC and indi!

fzurita:
Can this be merged now that you fixed Pokemon Puzzle league?

Gillou68310:
I will double check that issue first

gjimenezf:
Ready to close?

PR #386, opened 2017-09-12
Switch from adler32 to xxHash

loganmc10:
You can see some performance benchmarks here (conducted in Java):
https://jpountz.github.io/lz4-java/1.3.0/xxhash-benchmark/

And you and read a bit more about xxHash here:
http://cyan4973.github.io/xxHash/

This should provide a performance improvement in the cached interpreter and dynarecs. xxHash is significantly faster than adler32, while still providing an accurate hash.

### DIFF
--- a/projects/VisualStudio2013/mupen64plus-core.vcxproj
+++ b/projects/VisualStudio2013/mupen64plus-core.vcxproj
@@ -58,6 +58,7 @@
     <ClCompile Include="..\..\src\main\file_storage.c" />
     <ClCompile Include="..\..\src\main\lirc.c" />
     <ClCompile Include="..\..\src\main\main.c" />
+    <ClCompile Include="..\..\src\main\xxHash\xxhash.c" />
     <ClCompile Include="..\..\src\main\md5.c" />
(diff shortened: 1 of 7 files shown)


Narann:
Thanks. It could have been useful to bench real core situation.

As I'm not an expert in hash function I would love @bsmiles32 or @Gillou68310 to review this.

loganmc10:
Yes when I get some time over the next few days I'll probably try to provide some benchmarks. I'm also not an expert, but I assume measuring how much time is spent in the TLBWrite function over a set number of frames would provide a pretty good benchmark for comparison.

loganmc10:
Ok I tested this on Conker's Bad Fur Day. I let TLBWrite execute 10,000 times. I did it 12 times, throwing out the top and bottom values, and then averaged the remaining 10 values:

xxHash: TLBWrite executed for a total of 0.034939159 seconds on average (so those 10,000 executions took 0.034939159 seconds total)

adler32: TLBWrite executed for a total of 0.067649399 seconds on average.

So yes, as expected this change roughly doubles the speed of the TLBWrite function.

Narann:
Here we go!!!

PR #388, opened 2017-09-13
Remove DelaySI hack/option

loganmc10:
Right now we have the SI interrupt delayed in the majority of cases. The only case I know of where DelaySI needs to be off is Mischief Makers (input doesn't work with it enabled). However, I assume if 1 game has issues, others may as well.

The SI DMA has 2 parts: the DMA itself, and the PIF processing. This PR breaks those 2 steps up (right now they both happen at the beginning of the DMA). So for a DMA write, it does the DMA at the beginning of the DMA, and the PIF processing at the end. For a DMA read, it does the PIF processing at the beginning and the DMA at the end. You get the idea.

This removes the need to disable DelaySI for Mischief Makers. Therefore, I assume this is a more accurate way to represent the DMA, and we can remove the DelaySI option altogether.

I've tested a good number of games with this change and I've seen no regressions, however more testing is obviously welcome.

### DIFF
--- a/src/device/si/si_controller.c
+++ b/src/device/si/si_controller.c
@@ -54,22 +54,13 @@ static void dma_si_write(struct si_controller* si)
         *((uint32_t*)(&si->pif.ram[i])) = sl(si->ri->rdram.dram[(si->regs[SI_DRAM_ADDR_REG]+i)/4]);
     }
 
-    update_pif_write(si);
     cp0_update_count(si->r4300);
-
-    if (si->delay_si) {
-        si->regs[SI_STATUS_REG] |= SI_STATUS_DMA_BUSY;
-        add_interrupt_event(&si->r4300->cp0, SI_INT, /*0x100*/0x900);
-    } else {
-        si->regs[SI_STATUS_REG] |= SI_STATUS_INTERRUPT;
-        signal_rcp_interrupt(si->r4300, MI_INTR_SI);
-    }
+    si->regs[SI_STATUS_REG] |= SI_STATUS_DMA_BUSY;
+    add_interrupt_event(&si->r4300->cp0, SI_INT, /*0x100*/0x900);
 }
 
 static void dma_si_read(struct si_controller* si)
 {
-    int i;
-
     if (si->regs[SI_PIF_ADDR_RD64B_REG] != 0x1FC007C0)
     {
         DebugMessage(M64MSG_ERROR, "dma_si_read(): unknown SI use");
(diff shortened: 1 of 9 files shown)


loganmc10:
By the way, it'd probably be better if https://github.com/mupen64plus/mupen64plus-core/pull/360 was merged before this. I don't mind rebasing this, I'm just putting it here now for review

src/device/si/si_controller.c
```diff
@@ -173,12 +151,26 @@ int write_si_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
 void si_end_of_dma_event(void* opaque)
 {
     struct si_controller* si = (struct si_controller*)opaque;
+
+    if (si->regs[SI_STATUS_REG] & SI_STATUS_DMA_BUSY)
+    {
+        update_pif_write(si);
```
bsmiles32: You have some trailing spaces in there

src/device/si/si_controller.c
```diff
@@ -78,20 +69,9 @@ static void dma_si_read(struct si_controller* si)
...
     cp0_update_count(si->r4300);
-
-    if (si->delay_si) {
-        si->regs[SI_STATUS_REG] |= SI_STATUS_DMA_BUSY;
-        add_interrupt_event(&si->r4300->cp0, SI_INT, /*0x100*/0x900);
-    } else {
-        si->regs[SI_STATUS_REG] |= SI_STATUS_INTERRUPT;
-        signal_rcp_interrupt(si->r4300, MI_INTR_SI);
-    }
+    si->regs[SI_STATUS_REG] |= SI_STATUS_RD_BUSY;
```
bsmiles32: Not sure, but shouldn't SI_STATUS_DMA_BUSY be also set ?

src/device/si/si_controller.c
```diff
@@ -173,12 +151,26 @@ int write_si_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
 void si_end_of_dma_event(void* opaque)
 {
     struct si_controller* si = (struct si_controller*)opaque;
+
+    if (si->regs[SI_STATUS_REG] & SI_STATUS_DMA_BUSY)
```
bsmiles32: If we set SI_STATUS_DMA_BUSY for dma_si_read, we should change that with
if (si->regs[SI_STATUS_REG] & (SI_STATUS_DMA_BUSY | SI_STATUS_RD_BUSY) == SI_STATUS_DMA_BUSY)

src/device/si/si_controller.c
loganmc10: No, it doesn't work if you do that. SI_STATUS_RD_BUSY is supposed to be set during the PIF processing, and DMA_BUSY should be set when the DMA is happening. So what we are emulating with this PR, is during a read, we set RD_BUSY until the interrupt, and the emulate an "instant" DMA portion (we never set DMA_BUSY). We do the opposite for the write.

If we were trying to be a little better, we might have it flip from RD_BUSY->DMA_BUSY halfway through the timer, but doing it this way seems to work well.

bsmiles32:
Looks ok to me. I also tested some games with it and saw no regressions (Mischief Maker, Diddy Kong Racing, Blast Corp, Banjo-Tooie).
If you can fix the trailing space and look if also setting the DMA_BUSY for dma_si_read is okay (tested this myself but you can double check) then I'm ok to get this merged.

Good work on removing these hacks !

loganmc10:
> setting the DMA_BUSY for dma_si_read is okay (tested this myself but you can double check)

I thought it broke the input in Mischief Makers, but maybe I'm remembering just setting DMA_BUSY, and not RD_BUSY and DMA_BUSY, in any event, I don't think it should be set (https://github.com/mupen64plus/mupen64plus-core/pull/388#discussion_r138923316)

bsmiles32:
@loganmc10 Thanks for the explanation. Then I'm okay with the PR. Merging this.

----
Also, I'm indeed very busy atm, that's why PR accumulates. Don't desperate, I'll review everything when I get the chance (and usually you do very good work so it's ok). I also saw your mail on the mailing list, I'll try to answer it when I can !
@Narann thanks for merging some other PR recently, as I wasn't available to do it !

PR #390, opened 2017-09-14
Fix set_fpr_pointers

Gillou68310:


loganmc10:
As part of this PR, I would also remove CountPerOp=1 for Rogue Squadron and Battle for Naboo. It was added as a workaround for this problem, but the games perform better with the default CountPerOp, and I', not aware of any other issues when using the default

bsmiles32:
Quickly tested and looked okay to me.

Gillou68310:
> As part of this PR, I would also remove CountPerOp=1 for Rogue Squadron and Battle for Naboo. It was added as a workaround for this problem, but the games perform better with the default CountPerOp, and I'm not aware of any other issues when using the default

Done

loganmc10:
@AmbientMalice I know I've said this so many times, but after this latest group of PR's, I believe the Trading Post in Indiana Jones is really fixed. I played through the first 2 levels twice with a fresh config each time. The game still needs CountPerOp=2 to boot, so I'm curious if that causes any issues, I know you mentioned it has in the past with some physics stuff.

Anyway if you ever get a chance can you try the latest at https://m64p.github.io ? I know you have an interest in this game which is why I bring it up. If you do test I would delete any .eep saves you might have for the game beforehand, although I'm not 100% if it would make a difference.

AmbientMalice:
I'll give it a try this weekend.

AmbientMalice:
@loganmc10 Indiana Jones appears to be working near-perfectly now. (Demanding as heck on multi-threaded Angrylion's.) The raft no longer freezes the game momentarily on CF=2, but I haven't tested the jeep. I'm up to the third mission with no freezes. The third mission could be a fly in the ointment, and I probably should finish it -- but all signs seem positive.

loganmc10:
Awesome that's really encouraging to hear! I'll be curious to hear about the 3rd level/Jeep, I've never gotten that far in my testing

PR #400, opened 2017-09-20
Correct error in PI dma for loop

loganmc10:
Spotted by @richard42 

https://github.com/mupen64plus/mupen64plus-core/pull/375#discussion_r139870019


richard42:
oops, didn't see this. I just pushed the same change.

Issue #402, opened 2017-09-21
Custom Robo V2 freeze


loganmc10:
First reported here:

https://github.com/gonetz/GLideN64/issues/1583

After you start a game, the screen stays blank and the music loops.

It seems to possibly be related to EEPROM saving? The save screen acts differently in Project64. In mupen64plus, it seems to think that you already have 3 save files (with blank names), if you click on one, it acts like there is an existing save game (it doesn't ask you to name it). If you choose "overwrite" (the 2nd option), when you pick a name, it'll copy that name over all 3 saves.

Project64 starts with 3 empty saves, if you click on one, it'll ask you to name it, and that name will only apply to the save you chose.

Mupen64 0.5.1 has a similar issue, although it behaves slightly differently. When you first start, it has no saves (good), and you can name one, but the game will still lock up. If you close and go back into the game, you'll see that it copied the name of your save over all 3 saves.

loganmc10:
Oh easy fix, just needed to correct the save size! https://github.com/mupen64plus/mupen64plus-core/pull/403 @AmbientMalice

AmbientMalice:
Good job.

PR #403, opened 2017-09-21
Fix save size for Custom Robo V2

loganmc10:
Fixes https://github.com/mupen64plus/mupen64plus-core/issues/402


Narann:
Merged, but @loganmc10, could you point your source for save size? :)

loganmc10:
Sure, like I mentioned, I realized that the issue was with eeprom saving. So I looked at the PJ64 RDB to see if they have any special settings for this game, and I saw that they have 16KB eeprom set. I tested that out and it fixed all the issues, and here we are!

Narann:
I was expecting a wonderful list of all N64 games save sizes. ;D

Issue #405, opened 2017-09-22
PIF rewrite broke loading save states at startup


fzurita:
si_end_of_dma_event with a flags of 1 only happens at game startup. So if you are loading a save state at startup, a flags value of 1 never happens.

If you re-start the game and then load a save state, it works ok.

loganmc10:
Thanks for tracking this down, I can reproduce this in Linux. When you say "flags value of 1" what do you mean exactly?

loganmc10:
And yes, I bisected it to this commit: 752a7354b2b8a71029bc538212194162580f2f50

fzurita:
Flag value of 1 is SI_STATUS_DMA_BUSY

process_pif_ram is only called when SI_STATUS_DMA_BUSY is set in si->regs[SI_STATUS_REG]

loganmc10:
I'm not sure that it's related to that. I reverted my DelaySI changed (so the value of SI_STATUS_DMA_BUSY shouldn't affect whether process_pif_ram is called) (https://github.com/mupen64plus/mupen64plus-core/commit/0bcfd3cc780d56c9dab8d4e837e8f2a179df6129) while keeping the PIF changes by @bsmiles32 and it's still broken.

Hopefully @bsmiles32 might have an idea, but from my testing it doesn't seem related to the DelaySI/interrupt changes from the earlier commit.

fzurita:
I was thinking, a possible fix could just be a change of behavior. For example, if the core is commanded to load a save state before ```setup_pif_channel``` has been called, then don't load it quite yet. Wait until ```setup_pif_channel``` is called, then load it.

Now I just have to think of a non-hacky way of doing that.

bsmiles32:
Thanks both of you for bringing this to my attention. From what I understand, the deep issue is that with this refactoring I added new state variables (eg channels[0..4].{tx,rx,tx_buf,rx_buf}) which are not saved inside the savestate, but should. Before the refactoring, these were not needed as the pif ram content was always parsed, thus avoid the need for them. But now these variables are part of the emulated system state and therefore needs to be also saved inside the savestate.

[...]

PR #406, opened 2017-09-25
Bump savestate format to 1.2

bsmiles32:
Recent development omitted to save several emulated device state
variables. We remedy this by appending them at the end of the savestate
file. When loading older savestate files (<1.2 or pj64) we adopt a best-effort
approach to set the missing state to a reasonnable state. But it may
cause some instability.

### DIFF
--- a/src/device/ai/ai_controller.h
+++ b/src/device/ai/ai_controller.h
@@ -55,13 +55,13 @@ struct ai_controller
     uint32_t regs[AI_REGS_COUNT];
     struct ai_dma fifo[AI_DMA_FIFO_SIZE];
     unsigned int samples_format_changed;
+    uint32_t last_read;
+    uint32_t delayed_carry;
 
     struct r4300_core* r4300;
     struct ri_controller* ri;
     struct vi_controller* vi;
     struct audio_out_backend* aout;
-    uint32_t last_read;
-    uint32_t delayed_carry;
 };
 
 static uint32_t ai_reg(uint32_t address)
--- a/src/device/si/game_controller.c
+++ b/src/device/si/game_controller.c
@@ -59,7 +59,7 @@ static int is_pak_present(struct controller_input_backend* cin)
     return 0;
 }
 
-static void standard_controller_reset(struct game_controller* cont)
+void standard_controller_reset(struct game_controller* cont)
 {
     /* reset controller status */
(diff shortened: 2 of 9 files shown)


bsmiles32:
Should fix https://github.com/mupen64plus/mupen64plus-core/issues/405

loganmc10:
I tested loading older save states and it works again. Looks good

I don't think we should store the fb stuff in the save state. fb->infos is supposed to be populated by the graphics plugin, I think if we re-populate it from the save state, the gfx plugin might not expect that.

Also, fb->once is used to disable dp->r4300->recomp.fast_memory. If we load the value of fb->once from the save state, we'll probably miss changing dp->r4300->recomp.fast_memory

richard42:
It looks good to me, except I would recommend using a different data type for serializing the time_t values:

https://stackoverflow.com/questions/13957254/c-c-safest-way-to-send-time-t-over-socket

Cast them to uint32_t or int64_t

fzurita:
It seems to have fixed the save state issues in mupen64plus-ae

bsmiles32:
@loganmc10 You're probably right about the fb state, so I removed that from this PR
@richard42 Now, I (de)serialize time_t variables as int64_t values to avoid the vagueness of specification o the time_t type.

Also I did 2 small fixes :
- After re-initializing the pif channels, I need to call the input-plugin controllerCommand(-1,NULL) to notify that pif processing has ended
- There was some extra for loops for channels re-initialization which were a left-over from other tries at fixing the problem.

loganmc10:
You've still got:

```
/* reset fb state */
    memset(&g_dev.dp.fb, 0, sizeof(g_dev.dp.fb));
    g_dev.dp.fb.once = 1;
```

in savestates.c in 2 spots that I see.

bsmiles32:
I thought that, resetting the fb state would allow the video plugin to start from a state that it knows how to deals with (eg same as poweron). Otherwise, if I don't put anything, fb state will be uninitialized which seems worse to me. Did I miss something here ?

loganmc10:
No you are right, I've never looked at how the save states worked very closely. For some reason I assumed that poweron() would execute again, but it doesn't seem that way. It's good like it is now

bsmiles32:
Good. I guess it is mergeable now !

PR #407, opened 2017-09-25
Add AppVeyor config

loganmc10:
AppVeyor is like Travis for Windows.

This will do 2 things for us: it will automatically test the Windows/MSVC builds, just like Travis does for the Unix builds.

It will also test the x86 new dynarec for us, hopefully this will allow us to catch any compile-time errors to the new dynarec.

Someone is going to need to configure the GitHub repo to integrate with AppVeyor, I assume @Narann might know how to do this.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,19 @@
+version: 1.0.{build}
+
+configuration:
+  - New_Dynarec_Release
+  - Release
+platform:
+  - x86
+  - x64
+
+matrix:
+  exclude:
+    - platform: x64
+      configuration: New_Dynarec_Release
+
+before_build:
+- cmd: git clone --depth 1 https://github.com/mupen64plus/mupen64plus-win32-deps.git ../mupen64plus-win32-deps
+build:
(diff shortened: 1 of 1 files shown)


Narann:
Anyone have opinion about integrating AppVeyor? Looks like it's not a "Github App". @bsmiles32 @Gillou68310 @richard42 ?

loganmc10:
It does integrate with GitHub, if you look at the buildbot-type-thing I do for m64p.github.io, you can see:

https://github.com/m64p/mupen64plus-GLideN64/commits/master

If you hover over the green check marks, you'll see that it does Travis and AppVeyor. It will pass or fail PR's just like Travis does. However, in the settings for that repo it does show up under "Webhooks" instead of "Integrations and Services"

Gillou68310:
It looks good to me!

bsmiles32:
I for one would be very pleased when this gets merged, as I'm one of the worst offender when it comes to break Windows builds. Hope this will helps me avoid these issues in the future.

Narann:
I'm preparing the stuff right now. It looks seriously intrusive but that's why I asked after all! :)

Issue #409, opened 2017-09-27
Mischief Makers - Framerate halved


Krusher97:
Windows 64-bit build download from here : https://m64p.github.io/
In Mischief Makers the framerate drops from 60FPS to 30FPS in level 2-3. This happens in both Angrylion and GlideN64. The GlideN64 OSD even shows that the game is running at 100% speed. This causes a very visible slowdown which is not present on the actual console.
Speedrun of the level : https://youtu.be/RSAqQB5mlH4?t=6m56s
Random lets play of the level on a real N64 : https://youtu.be/aYK4XdfErHs?t=10m1s
Video of the game running on the emulator : https://my.mixtape.moe/xktdrg.webm
Recording of a console : https://my.mixtape.moe/fqlijm.mp4

loganmc10:
Can you post a save state?

This is an old, old issue, there is a bug report from 2008 saying the same thing:

https://github.com/mupen64plus/mupen64plus-user-issues/issues/71#issuecomment-130953267

Does it happen in Project64?

AmbientMalice:
I just tested with Project 64. Setting Counter Factor to 1 appears to eliminate the slowdown, but I'm unsure whether it could potentially have negative side effects elsewhere. (Timers, for example.)

loganmc10:
@Krusher97 do you know how to modify the CountPerOp setting in mupen64plus? It's in the core settings.

In mupen64plus-gui, you go to Settings->Core and Plugin Settings. In the "Core" tab, scroll down near the bottom and change "CountPerOp" from 0 to 1. Can you let me know if that solves the issue?

Krusher97:
That seems to have eliminated the slowdown. Couldn't find any issue with the timer either.

loganmc10:
Ok I created https://github.com/mupen64plus/mupen64plus-core/pull/410 to make CountPerOp=1 the default for that game

PR #410, opened 2017-09-27
Fix slowdowns in Mischief Makers

loganmc10:
Fixes https://github.com/mupen64plus/mupen64plus-core/issues/409

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -8695,6 +8695,7 @@ GoodName=Mischief Makers (E) [!]
 CRC=418BDA98 248A0F58
 Players=1
 SaveType=Eeprom 4KB
+CountPerOp=1
 
 [5690D74157C6623E2928A6F0353EF4AF]
 GoodName=Mischief Makers (E) [b1]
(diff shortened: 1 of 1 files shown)


loganmc10:
Hey by the way it looks like the AppVeyor checks are working, thanks @Narann !

Narann:
I realize we put `CountPerOp=1` for many games recently. Is this really the solution?

loganmc10:
It's definitely the worst thing about the emulator currently (accuracy wise). Every N64 emulator I know of (except CEN64) uses this hack (1964, Project64, mupen64, mupen64plus). By default it increments the COUNT register by 2 for every operation, but that's not how it worked on the N64.

I've never seen any solid proposal on how to fix this problem without going "all in" like CEN64 does.

But to answer your question, yes, it fixes some games, breaks others. BattleTanx needs CountPerOp=3 to even boot. Resident Evil needs CountPerOp=1 for the voices to work. Rogue Squadron needs CountPerOp=1 to prevent sound crackling during the opening Star Wars scroll, but it causes the CPU usage to go up, so we leave it at 2.

The list goes on and on, almost every game is likely affected in some way. If someone could come up with a better way to emulate the COUNT register without killing the performance it would be a massive step forward for N64 emulation

PR #411, opened 2017-09-28
More si refactorings

bsmiles32:
Main items of this work are :
- input backend API now only consists of 1 function - to get the input state. Pak detection and Controller connection test are handled internally by the core.
- Preliminary N64 Mouse support
- Extraction of controllers and paks from si into their own subdirectories
- Extraction of a cart subdirectory (with eeprom and af_rtc for now, other pi devices will also go there)
- Separation of backends interface from their mutable state
- Implement pak switching logic by reusing the "Mempak switch" button
- Use the internal rom db to know which paks are compatible with a game, and only switch theses
- Preliminary implementation of gb cart loader core command (for the Core-side). see https://github.com/mupen64plus/mupen64plus-core/pull/337 for anterior discussion.

Please test and report issues.


Narann:
Thanks @bsmiles32, looks like we have a fail with appveyor. Do you have access to the build log?

bsmiles32:
@Narann Yep, I saw the build logs and I was just me not updating the msvc project files properly.... Should be fixed with the new commits. Thank you (@Narann and @loganmc10 ) very much for having setup this AppVeyor thing, me likes it very much !!!

loganmc10:
I'll test this out in a bit, but it looks like the docs still need to be updated to include M64CMD_SET_GB_CART_LOADER

Also, main.c currently has an implementation of the loader functions. Was that just for testing?  There is a null version but it's ifdef'ed out

bsmiles32:
> but it looks like the docs still need to be updated to include M64CMD_SET_GB_CART_LOADER
Yep
> Also, main.c currently has an implementation of the loader functions. Was that just for testing? There is a null version but it's ifdef'ed out
Yep, it was just to ease testing, but shouldn't be part of the final merge. Still for local quick testing, it's more convenient than the hard-coded gb roms that were previously inside the source :)

Also, I will update the rom db a bit to allow proper pak compatibility.

loganmc10:
Ok I got around to testing it (with Pokemon Stadium). Seems like it's working properly. What games actually support Pak switching? Pokemon Stadium does not it seems.

One thing: it seems like some variables might not be re-initializing properly? If I open Pokemon Stadium to this point:

pokemon_stadium-000

And then Pak switch, I get this:
pokemon_stadium-001

Which I'm sure is normal, the game says not to do that. However, if I then restart the game, I get that same "Transfer Pak is not set properly" message. If I restart it a third time, then it's back to normal. I'm testing this in my GUI, so "CoreStartup" isn't being called between games, it's just executing M64CMD_STOP and then M64CMD_EXECUTE to stop/start the game.

loganmc10:
Sorry, one other question, I haven't looked through the code extremely close. Does this support changing the GB cart mid-game? Like is there a way to simulate removing the Transfer Pak, changing the GB cart, and then re-inserting the Transfer Pak?

As long as init_ram/init_rom are called each time the Transfer Pak is inserted then I assume it would work, but is there a way to simulate removing/reinserting the transfer pak?

bsmiles32:
> What games actually support Pak switching?

Perfect Dark

For Pokemon Stadium, I got the same message when switching paks in-game. I had to reset the console for it work normal. But I'll have to investigate your unitialized variable issue.

> Sorry, one other question, I haven't looked through the code extremely close. Does this support changing the GB cart mid-game? Like is there a way to simulate removing the Transfer Pak, changing the GB cart, and then re-inserting the Transfer Pak?
As long as init_ram/init_rom are called each time the Transfer Pak is inserted then I assume it would work, but is there a way to simulate removing/reinserting the transfer pak?

This is still not implemented. I would have to re-purpose the "rumble switch" button for that I guess. But yeah, that would be cool to have :) Especially for pokemon.

data/mupen64plus.ini
```diff
@@ -16,15 +16,17 @@ CountPerOp=1
 [34AB1DEA3111A233A8B5C5679DE22E83]
 GoodName=007 - The World is Not Enough (E) (M3) [!]
 CRC=3B941695 F90A5EEB
-SaveType=Controller Pack
+SaveType=None
```
loganmc10: Why change all these to "None"? There is a CoreGetRomSettings function that will return this information

loganmc10:
It seems like the "Mempak" settings in the INI file might be redundant, wouldn't it make more sense to check if the SaveType is "Controller Pak" and then just add the Transferpak entry for the games that support it?

bsmiles32:
> It seems like the "Mempak" settings in the INI file might be redundant, wouldn't it make more sense to check if the SaveType is "Controller Pak" and then just add the Transferpak entry for the games that support it?

The Mempak seems redundant at first, but some games support both in cartridge save and Mempak (eg Perfect Dark). See http://s9.zetaboards.com/Nintendo_64_Forever/topic/7041206/1/ for a good list.

data/mupen64plus.ini
bsmiles32: These are for games that only support Mempak, no in cart saves. But to be honest, I did a mass replace with regex, and I'm currently fixing these on a per-game basis. But it's a time consuming process...

loganmc10:
Couldn't we just change SaveType into an array of 5 chars (the number of different save types) and change the INI parser to support commas? like ```SaveType=Eeprom 16KB, Controller Pack```

I know it would make the code a little more complex but it would make the INI file a little easier to understand

bsmiles32:
I'm open to suggestions as for the db format, I just went with what seemed the easiest for me without much thinking. I'll push what I have or now, and when the format is settled I will try to continue the tedious work :)

src/backends/plugins_compat/input_plugin_compat.c
```diff
@@ -78,10 +55,31 @@ uint32_t input_plugin_get_input(void* opaque)
         input.getKeys(control_id, &keys);
     }
 
+    /* disconnect current pak immediately after "pak switch" button is released */
+    if (((keys.Value & PAK_SWITCH_BUTTON) == 0) && ((last_values[control_id] & PAK_SWITCH_BUTTON))) {
+        change_pak(&g_dev.controllers[control_id], NULL, NULL);
+        switch_delay[control_id] = SDL_GetTicks();
```
loganmc10: I don't think it's a good idea to use SDL_GetTicks() inside the emulation code. It can lead to varying experiences. What if the user switches paks while fast forward is enabled for instance? It would be better to use count_per_op, similar to how the interrupts are scheduled. This will lead to the same experience no matter how fast/slow the game might be running on the user's machine.

bsmiles32: I just used the same technique that the SDL input plugin. But you're right, we should avoid that. Maybe a simple counter would be enough.

bsmiles32:
@loganmc10 GB cart switching is implemented now. When press the "Rumble Switch" button, the GB cart loading procedure is triggered. I might have forgotten to disable that with games incompatible with the tpak, but yeah, I can add that later if needed.
If you keep the PR as is, you can edit the mupen64plus.cfg file while the game is running, and then press the "Rumble Switch" button to make the core load the new gb cart.
I tested with Perfect Dark (because it's the game that supports all paks, and allow to change between them en live, no need to reboot). I also tested with Pokemon Stadium, but this one requires a reset after changing the pak/GB cart. I also tried with Pokemon Stadium 2, but it requires a reset when changing paks/GB carts and we have some bugs preventing proper reset.

Enjoy !

About the db format, should we extend the SaveType to accept several types (comma separated ?) or should SaveType be reserved for the CartSaveType (Flash,SRAM,Eeprom 4k, Eeprom 16k) and just add a MemPak=Yes for the games that accepts the mempak. This would match other paks style : rumblepak, transferpak and maybe in the future the bio sensor pak :)

For more context, I may try in a more or less distant future to put more stuff in the rom db, especially cart types such as what do we map at 0x08000000 (none, flash, sram or something else), or what joybus device si connected inside the cart (none, eeprom, rtc, ....). That will allow for a more declarative style, and avoid the auto-detect flash/sram heuristic that we have now or avoid having both eeprom and rtc initialized (when none, or only one is effectively used). Also this would pave the way for more exotic N64 accessories/carts...
I might also add in the db if the game expects|is compatible with a VRU or a train controller, ...

@richard42 Any thoughts ?

loganmc10:
> About the db format, should we extend the SaveType to accept several types (comma separated ?) or should SaveType be reserved for the CartSaveType (Flash,SRAM,Eeprom 4k, Eeprom 16k) and just add a MemPak=Yes for the games that accepts the mempak. This would match other paks style : rumblepak, transferpak and maybe in the future the bio sensor pak :)

I'm fine with whatever, just having "SaveType=None" and "MemPak=Yes" seemed a little confusing, renaming SaveType to CartSaveType makes more sense.

I still have the issue in Pokemon Stadium where the ""Transfer Pak is not set properly" message will persist between games. It displays it if I open the ROM a second time, but not the third time. I'll try and see why that might be.

loganmc10:
@bsmiles32 I notice in the logs it prints out:

```
Core: Removing pak from controller 0
```
The second time I start the game, so it looks like it calls main_switch_pak() or something

bsmiles32:
Went through the full rom dataset to add/adjust mempak/rumble/transferpak entries. USA and Europe sets should be almost good (except for mistakes on my side as it is a tiring process). Japanese set is much harder for me because I don't understand Japanese, so I can't know what's compatible.

For now, only paks marked as compatible are switchable with the pak switch button, but as the db is still a bit WIP, it might be better to allow all kind of paks until we are more confident with the db ? Or keep it that way and have user report when they encounter a problem ?

bsmiles32:
@loganmc10 Can you test if my latest commit fixes your stop/execute initialization issue ? I think it was due to the usage of initialized static variables inside function which are initialized only once and not at emulation start.

bsmiles32:
Ok, now this PR should be complete. Documentation has been updated, API version bumped, some minor refactorings, and gb cart change button disabled when the game is not compatible with the transferpak.
Some more testing is welcome.

After this gets merged, maybe we should change the names of the Mempak Switch and Rumble switch buttons to a more appropriate name such as "Pak switch" and "GB cart change" buttons in the input plugin ?

tony971:
Will this feature have compatibility with libretro .srm-style saves? Or just the traditional .sav files?

bsmiles32:
Don't know much about the libretro srm format sorry. The core just expects a file with the RAM content. This also mean that for now RTC data that some GB emu addes at the end of sav files is not supported.

loganmc10:
@bsmiles32 now in Pokemon Stadium I get a "Saved file not found" message. If I revert the last commit, then it works as normal, here is what is in the log file:

```
Core: GB Loader ROM: C:\Games\N64\pokered\pokered.gb - zu
Core: GB cart type (13) mbc3 RAM BATT
Core: Providing default RAM content
Core: GB Loader RAM: C:\Games\N64\pokered\pokered.sav - zu
Core: Using a 32768 bytes GB RAM
Core: Game controller 0 (Standard controller) has a Transfer pak plugged in
```

Also, shouldn't all the stuff inside the ```#if 1``` go away before this is merged? The issue with stopping/restarting does seem to be fixed in the latest commit BTW

bsmiles32:
@loganmc10 The empty save issue is due to having a save file of different size from what the game expects. Here pokemon should have a 32k save, but I assume you had a <32k save. I reverted to the old way and be more forgiving about size mismatches.

I've also disabled the testing code in this PR, but will leave it in there until the ui-console is updated to provide this functionality.

loganmc10:
Ok looks good to me! Everything is working as expected

tony971:
> Don't know much about the libretro srm format sorry. The core just expects a file with the RAM content.

From what I can tell, the .srm file format is just a raw save ram dump. In the case of  game boy saves, the files *should* contain the same data, but with different file sizes.

https://forums.libretro.com/t/srm-to-sav-incompatibilities/1075

https://forums.libretro.com/t/solved-save-game-compatibility-help/932

bsmiles32:
Now that GB CART LOADER functions have been added to the ui-console, I've removed the testing code from this PR.

bsmiles32:
Updated against master.

richard42:
After this is merged, do we need to remove the "Rumble Pak" switch from all of the InputAutoConfig.ini sections, and the corresponding input plugin code?

bsmiles32:
No. The rumble switch button has been repurposed to change GB cart. So after merging we could rename both mempak and rumblepak switch buttons to "Pak switch" and "GB cart switch". And remove Pak switching code from input plugin.

bsmiles32:
I was thinking we could generalize in the future the GB cart loader interface to more kind of content not necessarily GB related. For instance PIF Boot ROM, 64DD Disks, cart ROM, flash, SRAM or Eeprom, .... If so we should rename it to a more generic name such as : "frontend loader" and rename the current functions to load_gb_rom, load_gb_ram, or something like that. What do you think ?

tony971:
"Accessory loader" sounds more obvious, but I may be misunderstanding the intention.

richard42:
Regarding generalizing the interface and re-using for other purposes; this is kind of a weird API in that the front-end registers a callback which then gets executed when the core needs to load the GB data. For other data loading needs, we have different mechanisms. The ROM is loaded with an explicit core command ("pushed" by the front-end). EEPROM/SRAM save data is (de)serialized from files by the core itself. Do you think there are any other data types for which it makes sense to use a callback architecture like this, as opposed to something more conventional?

bsmiles32:
My first candidate for extending this loading API are the 64dd disks because they can be swapped in-game just like the gb carts or the {mem,rumble,transfer}paks. But as you point out the frontend could also be "pushing" them to the core instead of having the core "pull" them. For the GB cart having it this way allows to centralize the GB cart parsing/loading procedure inside the core and not split it across multiple frontends. But for 64dd disks, this may or may not be the best.

Then after thinking about the 64dd disks, I thought, why not use this mechanism for all other cases. That's why I proposed it for other types of saved content (sram, flash, eeprom, mempaks), PIF Boot ROM [not yet supported], and even cart rom. I know some of them are already handled internally by the core, but this forces "fixed" names for the saved content decided by the core, not the user. It is not a very big deal, but I think it would have been a nice thing to have.

All in all, it was more a "what if" type of question to see what others were thinking, not a definite proposition. I would rather see that PR merged sooner and re-examine when needed how we want to extend the API.

richard42:
I don't really like this callback type mechanism for loading everything, due to the extra complexity. I can see that it's useful for the GB carts, so that's okay, and it may make sense for the DD disks as well. But I would avoid using it for other things if possible. I understand your point about having filenames fixed by the core, but this can also be an advantage as it allows different front-ends to use the same savestates.

bsmiles32:
Renamed GB cart loader to a more generic name : media loader. That way, we can extend this mecanism to 64dd disks in the future without keeping a confusing name.
PR https://github.com/mupen64plus/mupen64plus-ui-console/pull/41 has also been updated accordingly.

@richard42 Any other blocking issue before merging this ?

richard42:
now it has conflicts with your memory refactoring changes. I think you can merge this (and ui-console PR #41) after this is fixed.

Issue #361, opened 2017-08-13
RNG Broken


AmbientMalice:
This is essentially the same issue that Project 64 has, as pointed out here https://github.com/project64/project64/issues/1290 But it only occurs in PJ64 with Fixed Audio Timing.

I've tested mupen64plus, and in SSB64, the Master Hand pulls the same character out of the box every time, and in Episode I - Racer, the intro always shows Sebulba instead of Anakin. Changing CF settings and switching between interpreter and recompiler does seem to have some impact, but only in the choice of character in SSB64. The character will be the same every time, but changing the settings seems to affect the character.

loganmc10:
This is probably easily fixed. Do you know (or is there any documentation) on how the RNG is calculated in SSB64? Like is it using the value of AI_LEN_REG or something?

When we figure out what value it's using for the RNG, we can probably just rand() it a little bit to create the RNG effect. On a real N64, the speed of the DMA would vary a bit every time, just like a file copy in Windows isn't going to take the exact same time every attempt. But in the emulator, it is emulating a "perfect" DMA each time, so the RNG is broken.

loganmc10:
I'll probably play around with this in the next few days. I think it either has to be the value of AI_LEN_REG, or how long it takes for the audio interrupt to fire, I don't know what else it could be using for RNG

loganmc10:
So yeah, changing the length of the interrupt does affect the character that comes out of the box in SSB64.

To really get this right, we might need some help from some ROM hackers or the TAS community. We need to know what it's looking for. If I make it "a little random", then it'll switch (randomly) between Fox and Pikachu. If I make it a little more random, then Samus and Yoshi will appear sometimes. I thought other characters could come out like DK, but I never saw him. We can't just make it wildly random. We'll need to figure out what the "random range" should be.

loganmc10:
@AmbientMalice do you know the best place to get in contact with someone that might be able to figure out (or already knows) what Smash Bros uses to pick the character? I'd really like to know for sure what it is doing before I try to come up with a solution.

@Isotarge perhaps you know? Or know someone? I'd like to know what input Smash Bros is using to determine which character to pull out of the box in the opening sequence. We know it's related to the audio, but I'd like to know how it is using the audio information.

Isotarge:
[...]

PR #415, opened 2017-10-01
Randomize PI/SI interrupt timing

loganmc10:
This is to address https://github.com/mupen64plus/mupen64plus-core/issues/361

This is ready, but I put the WIP tag because it should get some testing. Right now I've only done very limited testing.

2 easy tests for this: The intro to Smash Bros and Star Wars Ep I Racer. The master hand should pull a random character out of the box, right now it always pulls out Mario. The Ep I Racer intro should either be Anakin or Subulba, right now it's always Subulba.

I added a core option to enable/disable randomizing the interrupt timing. Disabling it will be needed for some scenarios like regression testing, netplay, and possibly TAS runs. I think enabling it by default would be good, since it's more similar to how the N64 operated, assuming there aren't any regressions.

### DIFF
--- a/src/device/device.c
+++ b/src/device/device.c
@@ -53,6 +53,7 @@ void init_device(struct device* dev,
     unsigned int count_per_op,
     int no_compiled_jump,
     int special_rom,
+    int randomize_interrupt,
     /* ai */
     void* aout, const struct audio_out_backend_interface* iaout,
     /* pi */
@@ -117,7 +118,7 @@ void init_device(struct device* dev,
 
     init_memory(&dev->mem, mappings, ARRAY_SIZE(mappings), base, &dbg_handler);
     init_r4300(&dev->r4300, &dev->mem, &dev->ri, interrupt_handlers,
-            emumode, count_per_op, no_compiled_jump, special_rom);
+            emumode, count_per_op, no_compiled_jump, special_rom, randomize_interrupt);
     init_rdp(&dev->dp, &dev->r4300, &dev->sp, &dev->ri);
     init_rsp(&dev->sp, (uint32_t*)((uint8_t*)base + MM_RSP_MEM), &dev->r4300, &dev->dp, &dev->ri);
     init_ai(&dev->ai, &dev->r4300, &dev->ri, &dev->vi, aout, iaout);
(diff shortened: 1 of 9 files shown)


AmbientMalice:
Someone who's familiar with Mario Kart 64's correct behavior can hopefully comment on whether this fixes the wrong items being given to the player repeatedly.

theboy181:
You have a PR forPJ64 handy? like to test too.

AmbientMalice:
>You have a PR forPJ64 handy? like to test too.

Project 64 works differently. The problem is how audio timing has been handled, and is being handled. N64 games rely very heavily on audio timing. If you turn off Fixed Audio Timing in PJ64, most RNG stuff should work okay. PJ64 needs a new audio plugin, for a start. Zilmar does seem to be working on something.

loganmc10:
OK I've had a chance to test this a bit more. I don't see any regressions. I also tested the first level of Indiana Jones twice, since that was a game that always locked up at the end in the past, it worked both times.

loganmc10:
Here is a 64-bit Windows build if anyone would like to test:

[mupen64plus-GLideN64-rng.zip](https://github.com/mupen64plus/mupen64plus-core/files/1350123/mupen64plus-GLideN64-rng.zip)

AmbientMalice:
Haven't noticed any regressions. Tested a few games including Mischief Makers.

bsmiles32:
Found a regression : Conker Bad Fur Day will freeze during the intro with these changes. I remember it to be very sensitive to PI timings, so maybe the % 0x1000 is too big for this game.

loganmc10:
> during the intro

What part of the intro? Like when the N64/Rare logo is showing? or after the game starts?

I'd like to try and reproduce it and see if there is some way to make it a little more robust, 0x1000 is probably bigger than necessary, but along with lowering the value I'd like to see if I can figure out what is causing the freeze in the first place.

bsmiles32:
Freezes when Conker cuts the N64 logo.
The freezing is related to PI timings. I remember having a similar issue in an unpublished PI refactoring branch where I changed the PI timing to be constant instead of proportional to the length of the DMA.

loganmc10:
Ok I thought I had a potential fix but it turned out to be a dead end. I've modified the range to 0x40, it does seem to provide just as good randomness as 0x1000

loganmc10:
I haven't noticed any regressions since decreasing the variance to 0x40/64. I think this can probably be merged

PR #419, opened 2017-10-03
Better flags and timing for PI/SI interrupts

loganmc10:
I've actually been working on this and researching this for a while now, but I never found a case of it helping until @bsmiles32 comment in my other PR (https://github.com/mupen64plus/mupen64plus-core/pull/415#issuecomment-333855557), where Conker's would freeze if the PI timing was off (adding 0x1000 to the PI interrupt delay causes the game to freeze at the N64 logo)

The PI and SI have 2 kinds of data transfers: IO and DMA. It had always been a little unclear to me what the difference is but now I understand. A DMA is a DMA, initiated by a write to the read/write register of the PI/SI. IO is when there is a write/read to the memory space of the ROM or PIF.

With this PR, the DMA's only set the DMA_BUSY flag, not the IO_BUSY flag. I never actually set the IO_BUSY flag, I just trigger an interrupt immediately after an IO read/write.

The SI processing still needs to be done at the end of the DMA, not the beginning (or else Mischief Makers input won't work). I had to add a "si_type" field to keep track of what kind of processing to do at the end of the DMA. I added it to the end of the current savestate revision. I'm hoping to "sneak it in" there since it's only been a couple days since it was last bumped...

EDIT: I forgot where I was going with all this! This change makes Conker's more tolerant to PI timing changes (adding 0x1000 doesn't freeze the game anymore), and it's hopefully a generally more accurate way of emulating the DMA/IO interrupts

### DIFF
--- a/src/device/pi/pi_controller.c
+++ b/src/device/pi/pi_controller.c
@@ -154,9 +154,8 @@ static void dma_pi_write(struct pi_controller* pi)
         force_detected_rdram_size_hack(&pi->ri->rdram, pi->cic);
     }
 
-    /* mark both DMA and IO as busy */
-    pi->regs[PI_STATUS_REG] |=
-        PI_STATUS_DMA_BUSY | PI_STATUS_IO_BUSY;
+    /* mark DMA as busy */
+    pi->regs[PI_STATUS_REG] |= PI_STATUS_DMA_BUSY;
 
     /* schedule end of dma interrupt event */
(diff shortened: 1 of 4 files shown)


loganmc10:
Hm Bugs Life doesn't boot with this change, I'll have to do a bit more research

loganmc10:
I'm going to re-open this as a different PR in a bit

PR #420, opened 2017-10-04
Don't use IO_BUSY flags for DMA

loganmc10:
The IO/RD_BUSY flags are not meant for DMA's. Only the DMA_BUSY flag should be set during a DMA.

In order to accomplish this I had to add an si_type field, I added it to the end of the current savestate version.

I tested this with the usual suspects like Mischief Makers and everything still seems to work normally.

There may not be any real difference with this PR, since most games probably just check ```if IO_BUSY | DMA_BUSY```, but the DMA_BUSY flag is the only flags that should be set for a DMA


loganmc10:
Any comment on this? I can share my source for this if needed on Discord or something...

Basically the DMA_BUSY flag should be used for DMA requests processed via the PI/SI controller, the IO_BUSY flag is used for direct CPU writes, like when the CPU read/writes to the PIF/cart_rom via their memory address instead of via DMA, this is why these 2 different flags exist, I misunderstood their purpose in the past.

loganmc10:
This has been rebased against master. I also added a commit to make the PI DMA lengths consistent (https://github.com/mupen64plus/mupen64plus-core/pull/420/commits/1dfecd4af3866e596ceac4d978e6cd116325c578). I tested a number of SRAM and Flash RAM games without issue. The DMA alignment requirements are the same regardless of the address register, so it makes sense that they should all behave the same in this regard

loganmc10:
Thanks for the edits, looks good to me

loganmc10:
By the way, from what I read the word alignment is actually enforced (the last 2 bits are stripped from the address). I don't know if there are any games that run into that though

loganmc10:
@bsmiles32 I made that change to force word alignment (https://github.com/mupen64plus/mupen64plus-core/pull/420/commits/e415e75560c566a53ce7ae9502a7e979de773da0), I believe this is how the N64 actually did it, I'm not sure if any games actually write non-aligned addresses, but I believe this is how they would be handled

PR #423, opened 2017-10-04
Add SP_STATUS_REG and RDRAM_SIZE to gfx_info

loganmc10:
It has become apparent that in order to tackle some of the remaining GFX ucodes, the GFX plugins will need access to SP_STATUS_REG (see https://github.com/mupen64plus/mupen64plus-rsp-hle/pull/51#issuecomment-333310441)

This should allow that, I also added RDRAM_SIZE, since right now most GFX plugins use some kind of hack to try and determine that value.

@richard42


src/api/m64p_plugin.h
```diff
@@ -103,6 +103,15 @@ typedef struct {
...
     void (*CheckInterrupts)(void);
+
+    /* The GFX_INFO.version parameter was added in version 2.5.1 of the core.
+       Plugins should ensure the core is at least this version before
+       attemping to read GFX_INFO.version. */
+    unsigned int version;
+    /* SP_STATUS_REG and RDRAM_SIZE were added in version 2 of GFX_INFO.version.
+       Plugins should only attempt to read these values if GFX_INFO.version is at least 2. */
+    unsigned int * SP_STATUS_REG;
+    unsigned int * RDRAM_SIZE;
```
bsmiles32: RDRAM_SIZE should be const unsigned int* to prevent some abuse, no ?

loganmc10: True, I thought about just passing the actual value, but I just know in 3 years from now somebody is going to develop a speedrunning technique that involves ripping the expansion pack out in the middle of the game...

Issue #164, opened 2016-07-02
Feeding the audio plugin samples and telling the audio plugin to output to the driver should not be the same audio API function


wareya:
Samples are fed to the audio plugin indiscriminately during CPU emulation. When the audio plugin receives samples, it can sleep. **This is very bad for input latency.** Delays must not happen anywhere except **immediately** before input is sampled. The audio plugin should not feed samples to the driver except immediately before input is sampled.

I think new_vi() might also have an input/delay ordering problem but I can't test it since this is in the way.

write_aiX/??? -> X/gen_interupt -> write_ai_regs/ai_end_of_dma_event -> fifo_push/pop -> do_dma -> push_audio_samples -> push_audio_samples_via_audio_plugin -> aiLenChanged -> SDL_Delay

richard42:
Thanks for the suggestion regarding the proper frame limiter delay placement.  I'm gathering notes to eventually refactor the audio/input plugins and integrate into the core.  If/when I get around to this, I will make other related design changes to improve the emulator performance, so I'll make sure to consider your suggestions.

PR #424, opened 2017-10-05
SDL_Delay is only called just before input polling.

bsmiles32:
This should help reduce the input lag as suggested by wareya in issue https://github.com/mupen64plus/mupen64plus-core/issues/164

Be sure to test with `AUDIO_SYNC = False` in audio plugin so that SDL_Delay is only called right before input_polling.

@wareya can you test if it makes a difference for you ?

### DIFF
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -857,15 +857,14 @@ static void pause_loop(void)
  * Allow the core to perform various things */
 void new_vi(void)
 {
-    gs_apply_cheats();
-
-    main_check_inputs();
-
     timed_sections_refresh();
 
-    pause_loop();
+    gs_apply_cheats();
 
     apply_speed_limiter();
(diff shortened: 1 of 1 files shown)


fzurita:
I wonder if it would be beneficial to call it right before the delay and also right after?

loganmc10:
All that main_check_inputs does is run SDL_PumpEvents (polls input devices). I believe that calling it before and after would be the same as just calling it after. input-sdl executes SDL_JoystickUpdate() inside the GetKeys() function anyway, so the joystick is polled for input when GetKeys happens.

It would be more important to make sure that GetKeys is called at the proper time.

wareya:
>@wareya can you test if it makes a difference for you ?

I would test it, but I don't have a mupen64plus environment set up right now. Sorry about that. The code looks right, though!

Issue #418, opened 2017-10-02
Castlevania gets stuck in the intro demo


Jj0YzL5nvJ:
As the title says, at starting the gameplay demo of the intro, the game stays in the darkened screen.
Tested with:
```
Core: Goodname: Castlevania (U) (V1.0) [!]
Core: Name: CASTLEVANIA         
Core: MD5: 1CC5CF3B4D29D8C3ADE957648B529DC1
Core: CRC: F35D5F95 8AFE3D69
Core: Imagetype: .z64 (native)
Core: Rom size: 12582912 bytes (or 12 Mb or 96 Megabits)
Core: Version: 1449
Core: Manufacturer: Nintendo
Core: Country: USA
```

```
Core: Goodname: Castlevania (U) (V1.0) [!]
Core: Name: CASTLEVANIA         
Core: MD5: 1CC5CF3B4D29D8C3ADE957648B529DC1
Core: CRC: F35D5F95 8AFE3D69
Core: Imagetype: .v64 (byteswapped)
Core: Rom size: 12582912 bytes (or 12 Mb or 96 Megabits)
Core: Version: 1449
Core: Manufacturer: Nintendo
Core: Country: USA
```

```
Core: Goodname: Castlevania (U) (V1.2) [!]
Core: Name: CASTLEVANIA         
Core: MD5: 06B58673F7D31C56F8FE8186E86F6BD6
Core: CRC: 4BCDFF47 AAA3AF8F
Core: Imagetype: .v64 (byteswapped)
Core: Rom size: 12582912 bytes (or 12 Mb or 96 Megabits)
Core: Version: 1449
Core: Manufacturer: Nintendo
Core: Country: Unknown (0x245)
```

P.S: The roms work in Project64 1.6.1

fzurita:
I tried playing through a little bit of the game. I chose a character and go to the point where you fight a bunch of skeletons. At what point is it freezing? Is it past that?

loganmc10:
> I tried playing through a little bit of the game. I chose a character and go to the point where you fight a bunch of skeletons. At what point is it freezing? Is it past that?

Same, no issues here

Jj0YzL5nvJ:
The game is perfectly playable, but it gets stuck without user interaction.
In this part: https://youtu.be/kdXdANAATOI?t=1m54s

fzurita:
Oh ok. Yeah, I can confirm. The core seems to get stuck there in done sorry of infinite loop.

[...]

PR #427, opened 2017-10-06
Do a proper open bus read

loganmc10:
Fixes https://github.com/mupen64plus/mupen64plus-core/issues/418

Castlevania tries to read from "Cartridge Domain 1 Address 3" (http://en64.shoutwiki.com/wiki/N64_Memory#Memory_Map_Overview)

I have no idea with that address space is for. I seem to recall Paper Mario tried to access it as well via PI DMA when I was working on fixing duplicate saving in that game. Nothing I read gives me any hints as to what it might be for.

Anyway, when the N64 tries to read from unmapped memory, it's an "open bus" and it shouldn't just return 0. It returns some strange mash-up of the address.

Interestingly, the intro in the E version is different from the U version in Castlevania (the guy takes a different path and fights different enemies). I guess they were programmed differently, which might explain why this didn't happen in the E version

### DIFF
--- a/src/device/memory/memory.c
+++ b/src/device/memory/memory.c
@@ -86,12 +86,14 @@ static int writed(writefn write_word, void* opaque, uint32_t address, uint64_t v
 
 static void read_nothing(void)
 {
-    *g_dev.r4300.rdword = 0;
+    *g_dev.r4300.rdword = *r4300_address(&g_dev.r4300) & 0xFFFF;
+    *g_dev.r4300.rdword = (*g_dev.r4300.rdword << 16) | *g_dev.r4300.rdword;
 }
 
 static void read_nothingd(void)
 {
-    *g_dev.r4300.rdword = 0;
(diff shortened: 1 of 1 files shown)


loganmc10:
Just because I find this interesting, from Wikipedia:

> The PAL version of the game features voice acting for Dracula's servant in the Castle Wall and Castle Keep levels; the North American version did not include the voice work for the latter level.

The PAL version was released 4 months after the North American version. It seems like it probably received a bit more polish and perhaps this bogus read was a bug that was the result of a rushed release...

loganmc10:
I added a commit to do the same for the 64DD registers. F-Zero still works with this method (testing the U ROM). This may help other games like the English translations of Densha de Go! @bsmiles32 (https://krikzz.com/forum/index.php?topic=6362.msg49577#msg49577)

bsmiles32:
Cannot test at the moment, but looks good. Nice reads also. Thank you very much !

PR #428, opened 2017-10-07
Raspberry Pi: use new vendor library names

psyke83:
Needed for new firmwares on Raspbian stretch/ SDL 2.0.6. Also working with recent jessie firmwares.


Narann:
Thanks,
Please provide more context. I guess the open source OpenGL driver is now provided with Debian right?
So, with this PR, you want to keep old driver.
My question  is: Why not use the open source drivers ?

loganmc10:
The closed source driver is what comes enabled by default, the user needs to manually enable the open source driver. To use the open source driver they could just compile without setting VC=1

PR #431, opened 2017-10-11
Update to Visual Studio 2015/SDL2

loganmc10:
I have PR's for the other repo's incoming as well. This updates the build to VS2015 and SDL2.0.6 for Windows


fzurita:
How about having multiple projects for different msvc versions?

loganmc10:
I think it becomes a bit of a pain maintaining all those files. They are already a bit of an afterthought for most the devs lol

I don't see any real barrier to just supporting VS2015, it's free to download and it's not even the most recent version. Actually getting your hands on VS2013 is a bit harder in my experience. I have VS2017 and it seems to support VS2015 projects without issue, but I don't know how to get it to compile VS2013 projects

fzurita:
Yeah, you bring up good points.

bsmiles32:
> They are already a bit of an afterthought for most the devs lol

Sorry for that... The thing is, maintaining both Makefile and MSVC solution is kind of a pain for me as I have no access to a windows dev machine, so updating the MSVC file is always in "blind" mode when I don't completely forget about it. Having a single file to maintain would be far easier. Maybe looking at CMake or meson could help in this regard ? Yet, what we have now kind of works and migrating to another solution would be more work than editing once in a while the msvc file... so meh

loganmc10:
Oh I didn't mean just you. I also pretty much exclusively work in Linux, I just test things in Windows from time to time, but still usually with MinGW/Makefile. I'm still not 100% sure how to change stuff in these files via Visual Studio without it changing 5 other things I didn't want

fzurita:
Do you mean that we have to add more to ".gitignore"?

loganmc10:
No like it changes all kinds of version numbers and things in the vcxproj file when I try to add a .c file to it, I end up just adding it in a text editor. I'm probably just hopelessly stubborn.. I'm sure there must be a way

richard42:
Thanks for this; switching to VS2015 is good, it's probably about time. But I'll wait until you have PRs for all the modules including rice and glide64mk2 and then merge them all at the same time. I don't want to have them out of sync with each other.

loganmc10:
Good point, neither video-rice or video-glide64mk2 will compile using SDL2 right now though (https://github.com/mupen64plus/mupen64plus-video-glide64mk2/issues/90#issuecomment-336161406)

richard42:
Apparently the SDL2 headers for opengl have been modified in a way which is inconvenient for us. I hate having to use GLEW; our previous solution was about as good as it could be.

rlabrecque:
Is there any reason to support SDL1 at all anymore? Why ever `#ifdef SDL2`?

richard42:
the last release of SDL1 was 4 years ago, so it's probably safe to remove support for it in the near future.

loganmc10:
Yeah as far as I can tell Glide64mk2, rice, and audio-sdl are all incompatible with SDL2 on Windows, once that is all corrected then SDL1 could go

Papermanzero:
Is it possible to add the rumble feature on windows with SDL2?

PR #436, opened 2017-10-14
PI refactorings

bsmiles32:
Main focus of this PR is to move some dma logic inside each dma'ble component (flashram, sram, cart_rom) to make pi_controller more agnostic to the type of component it accesses. I also moved these components inside the cart subsytem.


loganmc10:
A couple notes just from reading:

You can probably use ```length = (length & UINT32_C(0x00fffffe)) + 2;``` in all cases. I just did it only for cart_rom since there was a case where it was required (AI Shogi 3), but it probably could/should be used in all cases.

From a git perspective, this conflicts in a pretty major way with both https://github.com/mupen64plus/mupen64plus-core/pull/415 and https://github.com/mupen64plus/mupen64plus-core/pull/420. It would be nice if they could be reviewed and either merged first, or incorporated into this PR

src/device/device.c
```diff
@@ -44,6 +44,46 @@ static void write_open_bus(void* opaque, uint32_t address, uint32_t value, uint3
...
+    *handler = &h; \
+    } while(0)
+
+    if (address >= MM_CART_ROM) {
+        if (address >= MM_CART_DOM3) {
+            /* 0x1fd00000 - 0x7fffffff : dom3 addr2, cart ??? */
+            RW(NULL, cart_dom3);
+        }
+        else {
+            /* 0x10000000 - 0x1fbfffff : dom1 addr2, cart rom */
```
loganmc10: Should this comment say 0x1fcffff.....? It says dom3 starts at 0x1fd...

bsmiles32:
I tried not to change any behaviour in this PR even if I could have done some changes like you suggested. This is on purpose. Changing behaviour can be tricky and I thought it would be more appropriate in a later PR. But yeah there is a lot of potential to gather common behaviour at the pi_controller level.

I'm fine with any order for the merges, I can rebase this PR if needed. Just, I will be away for the whole week, so I won't be able to do anything in the coming days.

src/device/device.c
bsmiles32: 0x1fc00000 is allocated to pif rom, pif RAM. Cartridge domain 3 does indeed start at 0x1fd00000, but I don't know what is mapped there and how it is accessed.

loganmc10: Ah ok makes sense

bsmiles32:
Rebased against master.

richard42:
bsmiles32, you can merge this when you get back next week.

loganmc10:
I finally got around to testing this, Paper Mario (U) crashes, here is the backtrace:

```
0x00007fffbfbf279d in flashram_command (flashram=0x7fffc0f8f7d8 <g_dev+17834712>, command=3523215360) at ../../src/device/cart/flashram.c:40
40	    uint8_t* mem = flashram->istorage->data(flashram->storage);
(gdb) backtrace
#0  0x00007fffbfbf279d in flashram_command (flashram=0x7fffc0f8f7d8 <g_dev+17834712>, command=3523215360) at ../../src/device/cart/flashram.c:40
#1  0x00007fffbfbf2adc in write_flashram_command (opaque=0x7fffc0f8f7d8 <g_dev+17834712>, address=134283264, value=3523215360, mask=4294967295) at ../../src/device/cart/flashram.c:136
#2  0x00007fffbfbf1b93 in write_cart_dom2 (opaque=0x7fffc0f8f760 <g_dev+17834592>, address=134283264, value=3523215360, mask=4294967295) at ../../src/device/cart/cart.c:165
#3  0x00007fffbfc1b455 in mem_write32 (handler=0x7fffc0f9b9b0 <g_dev+17884336>, address=134283264, value=3523215360, mask=4294967295) at ../../src/device/memory/memory.h:106
#4  0x00007fffbfc1bd3c in r4300_write_aligned_word (r4300=0x7fffbfe8d500 <g_dev>, address=134283264, value=3523215360, mask=4294967295) at ../../src/device/r4300/r4300_core.c:398
#5  0x00007fffbfc0d87b in SW (r4300=0x7fffbfe8d500 <g_dev>, op=2890268672) at ../../src/device/r4300/mips_instructions.def:333
#6  0x00007fffbfc1b14d in InterpretOpcode (r4300=0x7fffbfe8d500 <g_dev>) at ../../src/device/r4300/pure_interp.c:687
#7  0x00007fffbfc1b363 in run_pure_interpreter (r4300=0x7fffbfe8d500 <g_dev>) at ../../src/device/r4300/pure_interp.c:736
#8  0x00007fffbfc1b8ae in run_r4300 (r4300=0x7fffbfe8d500 <g_dev>) at ../../src/device/r4300/r4300_core.c:157
#9  0x00007fffbfbf4c4c in run_device (dev=0x7fffbfe8d500 <g_dev>) at ../../src/device/device.c:223
#10 0x00007fffbfc2a7aa in main_run () at ../../src/main/main.c:1323
```

I believe Paper Mario writes some goofy value to CART_ADDR_REG. It might be using the wrong handler or something. This is why Paper Mario had those duplicate saving issues for so long

bsmiles32:
Rebased against master and fixed the crash in Paper Mario.
@loganmc10 : The crash was due to wrong "opaque" object passed to cart_dom2 mmio handlers. Thanks for reporting this.

loganmc10:
The crashing is fixed, but the duplicate saving in Paper Mario is back. You can test this using the U rom by:

Making a save game, then close the game and reopen it, you'll see that your save game now occupies all 4 slots.

Just for reference here is the PR that originally fixed it: https://github.com/mupen64plus/mupen64plus-core/pull/375

loganmc10:
I don't think using cart_rom_dma_read/cart_rom_dma_write is the correct approach. Since the whole thing lies outside the ROM's address space, it just writes all 0's in this case.

It's more like a "DMA open bus" situation. Writing all 0's seems to work instead of the address shifting thing that is used for CPU reads from invalid memory addresses. I can't remember if I actually tried handling it like the other open bus situation, or if I just tried all 0's. All 0's is what we do now anyway

loganmc10:
But now that I remember, you are just trying to mimic what we currently do, so then yes, cart_rom_dma_read/cart_rom_dma_write would be accurate to what we are currently doing

bsmiles32:
Exactly, for this PR I want same behavior. but after this is merged, we can try more accurate behavior. Things that come to mind is indeed the open bus behavior, and ideally doing the effective dma copy inside pi controller instead of in each devices. Also, maybe we could derive the dma timings from the pi_bsd_domX_zzz_regs. Also, having the cart decide what's mapped in addr2 space (rom, SRAM, flash). Same for addr1 with DD.

loganmc10:
Looks good to me, I believe I've tested a game with every type of save, no issues anymore

bsmiles32:
Great ! If you're okay, I'll merge this right away :)

loganmc10:
yeah fine with me

PR #437, opened 2017-10-14
ROM database update

Enker:
This should make the database on par with PJ64 :smile:. The main changes are GoodN64 additions, and I checked/added all Mempak/Rumble/Transferpak entries for Japan.
I also tested/confirmed some cart save types, and added CountPerOp=1 to the games below. I tested them with the latest M64P GUI build.

Goemon games: Fixes slowdowns during some cutscenes.

Pokemon Snap: project64/project64#1393, same with M64P

Wonder Project J2: project64/project64#1238, same with M64P


loganmc10:
Wow this seems like a massive effort, did you use some kind of tool? 

I noticed Transferpak was removed from the Japanese version of Perfect Dark and Mickey's Speedway, do they not support it?

Also I noticed at least 1 MD5 changed (E version of Indiana Jones), can someone confirm that? I don't think I have a copy

Enker:
Yes, it took me a few days :sweat:. I didn't use a tool, I just compared ROMs against PJ64's rdb, and added the ones that matched with GoodN64 V327.

Those two (J) games do not support the Transferpak, due to the GB games not being released in Japan.

The previous Indiana Jones (E) hash was little-endian. I noticed it when I was renaming it.

loganmc10:
Thanks, like I said, massive effort and very tedious, good work

Enker:
No problem :). There are still tons of missing GoodN64 roms, but I only added the non-bad/hack ones that I noticed missing. We'd need a tool to add those, if they are wanted for the database. Though copying the game settings would be tricky.

PR #439, opened 2017-10-15
Fix crash when loading savestate.

bsmiles32:
savestate loading code was trying to poweron rumble pak on a
non-initialized rumble pak.
Now we check that the game is rumble pak compatible (therefore it has
been initialized) before calling the poweron function.

A similar issue was also present for transferpak.


fzurita:
I just tested this. It does seem to fix the crashing when loading save states.

PR #443, opened 2017-10-18
Remove CountPerOp=1 for Ep I Racer

loganmc10:
Some PR over the summer has fixed the audio in this game with CountPerOp=2, it used to not work (https://github.com/mupen64plus/mupen64plus-core/pull/272)

The CPU usage in this game is a lot higher with CountPerOp=1, so since there are no more issues with CountPerOp=2, it should be the default again. Hopefully this means I can finally play it on my tablet at full speed!

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -14298,7 +14298,6 @@ CRC=53ED2DC4 06258002
 Players=2
 SaveType=Eeprom 16KB
 Rumble=Yes
-CountPerOp=1
 
 [8CC7E31925FBFA13A584F529E8912207]
 GoodName=Star Wars Episode I - Racer (E) (M3) [f1] (Save)
@@ -14311,7 +14310,6 @@ CRC=61F5B152 046122AB
 Players=2
 SaveType=Eeprom 16KB
 Rumble=Yes
-CountPerOp=1
 
 [CFA21E43DAC50DFF3122ABF3AF3511F8]
 GoodName=Star Wars Episode I - Racer (J) [b1]
@@ -14324,7 +14322,6 @@ CRC=72F70398 6556A98B
 Players=2
 SaveType=Eeprom 16KB
 Rumble=Yes
-CountPerOp=1
 
 [E4353B43CB4302B7A308A42D6BB04435]
 GoodName=Star Wars Episode I - Racer (U) [f1] (Save)


AmbientMalice:
From memory, Episode 1 Racer has a pretty atrocious framerate with 2. It's one of the reasons the game has CF=1 on PJ64. The game had framerate issues on real N64 hardware, and that's kind of a problem for a fast paced racing game.

loganmc10:
I played through a race on Tatooine and it seemed quite smooth, normal/playable to me at least. This is easy to test though, the CountPerOp settings can be overridden for any game.

The difference in CPU usage is pretty dramatic. On my Windows laptop it shows the CPU usage at 24-25% during a race (4 logical processors so that is almost maxed out). With CountPerOp=2 the CPU usage is around 15%

AmbientMalice:
>The difference in CPU usage is pretty dramatic. On my Windows laptop it shows the CPU usage at 24-25% during a race (4 logical processors so that is almost maxed out). With CountPerOp=2 the CPU usage is around 15%

The game runs at 30fps with frequent dips with CPO 2. It runs at 60-70fps at CPO 1. It doesn't appear to have to proper frame limiter. There's no right answer to this, but 60fps is inherently better than 30fps with constant dipping. The problem is, of course, that running the game at 60fps is inherently more demanding. Same as running Rogue Squadron at 60fps with a hack.

loganmc10:
Well the right answer is what it ran at on a real N64, what was the framerate there?

AmbientMalice:
>Well the right answer is what it ran at on a real N64, what was the framerate there?

Under 30 FPS pretty much constantly. It's one of the reasons the PC version is preferred, because the N64 version is hard to control -- almost brokenly so in high resolution mode -- and the Dreamcast version is locked at 30fps. (Yet ironically the N64 version does run at 60fps if overclocked.)

I think the fundamental problem is there's no simple user facing overclocking function. The game running atrociously is kinda accurate to hardware, but I wouldn't wish that experience on anyone. However, accuracy should ideally be the default. This problem will only get worse as N64 emulators improve in accuracy. Running games at a higher framerate has been a luxury afforded by a sloppy approach to emulator timings.

I'm still not sure what the right approach with games like Perfect Dark is. The game runs at 17fps on a regular basis on mupen64plus. That's absolutely horrible. But raising CountPerOp to 1 causes some minor bugs in addition to a massive framerate boost. The former is more accurate, but the latter is a much more enjoyable way to play the game.

One possible solution would be an option users tick to automatically override the CPO setting for certain games. These games could have canoverlock=true in the RDB, and if the player has gamespecificoverclock=true enabled, those games are automatically given inaccurate, but faster timing.

loganmc10:
Yeah I definitely think the default should be how an N64 would run it. It probably wouldn't hurt to have an overclocking option. I haven't really looked at what PJ64 did to accomplish that. Setting CountPerOp=1 sort of accomplishes that, but I'm sure there is a better way. Then people could just enable or disable overclocking via the config as they see fit

PR #444, opened 2017-10-18
Re-work the way VI interrupts are timed

loganmc10:
A VI interrupt should happen when VI_CURRENT_REG = VI_V_INTR_REG. VI_CURRENT_REG basically has no meaning unless VI_V_SYNC_REG is set. Therefore no interrupt should be triggered until VI_V_SYNC_REG is set.

So, this removes the more hackish method we currently use of just kicking off the VI's when the game starts, and replaces it with this method:

The first VI is triggered when VI_V_SYNC_REG has a nonzero value written to it.

Benefits:

- Indiana Jones used to not boot with CountPerOp=1 (if no save file existed), now it does
- Both BattleTanx (U) games needed CountPerOp=3 to boot, now they don't

I think this shows that this is probably a more accurate way to kick off the VI's. I tested a few games, including the usual suspects like Pokemon Puzzle League and didn't see any regressions.


Gillou68310:
@loganmc10 this seems to have introduced a large delay at startup.

loganmc10:
I noticed that Super Mario seemed to take a bit longer for the image to appear, I didn't notice any other games, do you have an example?

I'm open to suggestions, a long delay isn't necessarily an indication of an issue, the console might have had the same delay. I believe this is a more accurate way of kicking off the VI's but I could be wrong

Gillou68310:
Well, almost every game I tested have this delay, turok 2 is an another example.

loganmc10:
Fair enough, still, I think someone would need to measure a real console, from the time they hit the power switch to when the image appears. Time without an image on the screen is still potentially legitimate time.

The way we were doing it before was definitely hacky, like I mentioned, I believe this is the right way to trigger the first VI, but I could be wrong. I don't have a real N64 to test unfortunately

Gillou68310:
How is this done in other emu? PJ64 for example?

loganmc10:
The same way as we did it before, they just kick off the VI's when the game starts with some delay like 5000 (I can't remember the exact value).

They suffer from the same issues with Battle Tanx and Indiana Jones, where you need a specific Counter Factor for those games to boot

Gillou68310:
Also it would be interesting to understand why the core takes so much time before writing to VI_V_SYNC_REG

Gillou68310:
@loganmc10 @richard42 @bsmiles32 we really need to find a workaround for the startup delay. This is probably not a big deal for users but during my debugging sessions I'm spending most of my time waiting for the game to start.
Maybe we can add a config parameter to allow using the old hacky (but fast) way.
What do you think?

loganmc10:
Is it really that long for you? I don't really notice it.

Anyway, I thought this was a pretty good change, it allows battletanx to boot without needing CountPerOp=3 for instance.

A VI interrupt is supposed to happen when VI_CURRENT_REG = VI_INTR_REG. so an interrupt shouldn't happen until the registers are set in such a way that that is possible. That's what this PR was trying to address.

I think the goal should be to emulate the N64 as accurately as possible. I don't have a physical n64 anymore, but I suspect it has similar startup delays. The problem with the old hacky way is that we've now gotten rid of the CountPerOp settings for battletanx (and I think at least 1 other game), so those games will no longer boot using the old method

loganmc10:
By the way, are you sure it's all related to this commit? If you use GLideN64, it recently started compiling all the shaders at boot, instead of on-the-fly, it also generates textures used for noise emulation at boot. So it probably has much longer startup times compared to other GFX plugins, and that would have increased lately

Gillou68310:
> Is it really that long for you? I don't really notice it.

When reverting this commit I went from 10 sec waiting to instant boot in WDC. I just tested again so I'm pretty sure it's not related to any other plugin update ;-)

Also I'm not talking about reverting this commit because I agree with what it addresses. I would rather try to find a non intrusive workaround in order to bypass the startup delay.

loganmc10:
I guess it's probably different in a release build, on my PC when using GLideN64, WDC launches in just under 5 seconds after I click "go" (using mupen64plus-gui). If I use Angrylion Plus, the game launches in just under 2 seconds.

Just re-adding ```add_interrupt_event_count(&dev->r4300.cp0, 5000);``` to the end of poweron_device() would probably speed up the boot (mimic the old behavior).

I don't think it's really needed for end users. you you could probably wrap it like:

```
#ifdef FAST_BOOT
add_interrupt_event_count(&dev->r4300.cp0, 5000);
#endif
```

at the end of poweron_device(), something like that, and then just define FAST_BOOT when you are debugging/working? I think that would be a bit cleaner than adding a config parameter.

Gillou68310:
Ok I made some debugging and the core is looping forever into the same procedure (doing the same thing over and over) until ```cp0_regs[CP0_COUNT_REG] > UINT32_C(0x10000000)``` (see special_int_handler)
Reducing the value will reduce the startup delay but I have no idea what this is used for.
@richard42 @bsmiles32 do you have any information about the special interrupt?

fzurita:
This is an interesting one. On one device, I have a 20-30 second delay when starting CBFD, on another device, the delay doesn't exist. If I didn't FB emulation on GLideN64, the delay goes away (maybe 5 seconds). I'm not sure why one device is affected and the other is not.

bsmiles32:
@Gillou68310 : Interrupt/event scheduling is quite complex and I don't fully understand it. With that said, it works more or less like that :
- we maintain some kind of priority queue ordered (more or less) by their timestamp ("count" in the code). This priority queue is implemented with a single linked-list (whose node are allocated via a pool of reusable nodes). An extra field "cp0_next_interrupt" contains the first upcoming interrupt/event timestamp, so we know if we need to execute some interrupt handler (eg if CP0_COUNT_REG >= cp0_next_interrupt).
- event/interrupt timestamps (eg "count") is 32-bit unsigned and can overflow, so a scheme was devised (when 64-bit registers were not common) to handle gracefully this wrap-around. This is done thanks to the "SPECIAL_INT" event.
- This SPECIAL_INT event is always triggered when CP0_COUNT_REG wraps around, eg reaches 0 (more or less, because we don't check at every cycles if CP0_COUNT_REG >= cp0_next_interrupt). When it is triggered, it sets "special_done" to 1. There is a test to not do anything if CP0_COUNT is above a certain threshold, but I don't understand why it exists.
- This "special_done" variable changes how new events (which have an overflowed counter) are prioritized when they are inserted into the priority queue (see before_event function).
- The before_event function is what keeps the priority queue ordered (however it seems that not all the ordering details are encapsulated inside this very function, some ordering details such as the special handling of special events are also inside the add_interrupt_event_count).
This function is a bit convoluted because it compares events relative to the current CP0_COUNT_REG to see if event timestamps have overflowed or not.

All in all, the interrupt/event scheduling code is quite difficult to reason about because :
- it has been implemented with uint32_t timestamps that can overflow
- such overflow handling logic has diffused in several places of the code (not just the ordering function)
so reasoning about this code is very hard to say the least.

I can see 2 ways to proceed from here :
1. do away with uint32_t timestamps, and use uint64_t so that overflows are not a problem anymore
2. or keep uint32_t timestamps and try to refactor this code so that we have a clear priorty queue code, with a well defined ordering function. But even then I can't guarantee that it would be bug-free.

loganmc10:
Maybe I misunderstand what you're talking about, but isn't the COUNT register wraparound a feature of the R4300? My understanding is that it is a 32 bit register and the r4300 would reset it to 0 when it overflowed

bsmiles32:
@loganmc10 COUNT register is indeed a 32-bit overflowing register. The problem is how to compare events taking into account this overflowing. If we go the 64-bit route, we make COUNT being the lowest 32-bit of the global 64-bit cycle counts, and make events also 64-bit so that ordering them becomes trivial and there would be no need for the wrap-around logic + special_int.

Gillou68310:
@bsmiles32 thanks for the detailed explanation! "SPECIAL_INT" is probably not a good name BTW, "OVERFLOW_INT" would be more appropriate IMO.

> There is a test to not do anything if CP0_COUNT is above a certain threshold, but I don't understand why it exists.

Yep this is a mystery.

> do away with uint32_t timestamps, and use uint64_t so that overflows are not a problem anymore

I like the idea of using a 64bit value for the count, it will definitely simplify the code. Also at 93.75MHz it will take approximately 6239 years to overflow the 64bit value. I think we are good here :-)

Issue #129, opened 2015-08-15
64DD emulation


LuigiBlood:
Been working on 64DD emulation for a while, implemented code on 64DD ports like I did with Project64, and then with the actual emulation, using MAME 64DD disk format as it is closer to physical (makes it easier).

Except it's quite hard to implement, and every reads makes mupen64plus freeze (the debugger still says it works, even though the window clearly froze).
And I don't know why.

clobber:
cc: @richard42 @bsmiles32 @Ari64 @Gillou68310 @littleguy77 ???

LuigiBlood has put a lot of work into reverse engineering the 64DD and provided implementations for other emulators. Does it make sense for 64DD support to slip through the cracks in M64P?

Narann:
64DD support in m64p make sense, I suppose we simply lack of developers tackling this problem. :(

LuigiBlood:
I still want to tackle this but I need help on this one.

vgturtle127:
@LuigiBlood So have you made any progress on this, LuigiBlood?

I would be willing to help, but I don't think the debugger and other stuff works in the GUI's does it? I use Mupen64Plus-Qt...

LuigiBlood:
Absolutely none, and the apparent lack of interest from m64p devs doesn't help at all. D:

vgturtle127:
@LuigiBlood Huh. Not even sure what to say about that. :disappointed: 

Is your fork up to date with the latest changes?

LuigiBlood:
Honestly you may see 64DD support for mupen64 code for RetroArch (I promised it). And we'll see if I'll port that at some point.

vgturtle127:
@LuigiBlood What about BizHawk? I know that sounds weird, but it is pretty much the same kind of thing, right?

LuigiBlood:
BizHawk is more special. But not my priority yet.

vgturtle127:
@LuigiBlood More special? :laughing: 

I understand, I didn't know it was much different from the RetroArch core other than some Banjo-Kazooie/Banjo-Tooie additions.

Mastergatto:
@LuigiBlood How's going with the 64dd support in m64+? I'd love to try some Mario Talent, unfortunately I can't because I'm on Linux and thus Project64 is not on the table.

LuigiBlood:
I did implement it in parallel-n64 back then... but I honestly don't remember how I really did it.
[...]

PR #446, opened 2017-10-19
[WIP] Add support for 64 Disk Drive.

bsmiles32:
Not ready for merging nor public consumption yet, and will certainly need several iterations to be done (read: I will rebase and force push often !).

I'm trying to finally add 64DD support. I've based my work on the parallel code, cen64 and looked a bit at pj64 source also. I've also updated the code to match the latest core paradigms.
But in the process I might have mixed up some lines between all these. So it doesn't really work : DD IPL boots and try to load a disk, but get stuck with a black screen.
The last DD MMIO/DMA logs show the following.
```
Core: DD DMA write dram=0033f450  cart=05000000 length=000003a0
Core: DD REG: 05000540 -> 00030000
Core: DD ROM: 0609FF00 -> c3dbfe61
Core: DD ROM: 060004BC -> 3c0ba600
```
@LuigiBlood : Any idea where I can look to debug this ?

NDD file and DD IPL ROM paths are hardcoded inside the source (./NUD-EFZJ-JPN.ndd and ./64ddipl.bin). Proper support for disk and ipl loading will come later.

Issue : https://github.com/mupen64plus/mupen64plus-core/issues/129


bsmiles32:
Rebased against master, updated msvc files, did some more editing and now I can get a bit further...

image

LuigiBlood:
That sounds good to me.

bsmiles32:
@LuigiBlood Any idea what I should look into now ? Or games I can test ?

LuigiBlood:
To be honest just about any game in the 64DD library are pushing the GFX plugins and I think are worth taking a look at. Sim City 64 is the one that pushes them harder.

Mario Artist suite is worth a try to add disk swapping support and such. You could also check saving support that way.

Also on parallel-n64, Japan Pro Golf Tour 64 is not playable. I assume it's also the case of mupen64plus.

bsmiles32:
Rebased against master.

bsmiles32:
Rebased against master

bsmiles32:
Rebased against master

LuigiBlood:
`/* XXX: add convert back function */`

Can be found here:
https://github.com/LuigiBlood/ddconvert_back
https://github.com/project64/project64/blob/master/Source/Project64-core/N64System/N64DiskClass.cpp#L449

bsmiles32:
Thanks ! I will indeed need to look into that for good save support. Also I need to do proper savestates, disk swapping, ...
I move slow in this PR, but I think after the next release, I'll be able to have it in shape and ready :)

LuigiBlood:
About M64CMD_SET_MEDIA_LOADER, Mario Artist might be a good way to test since you can load the game + GameBoy Camera ROM and RAM to take photos from.

I implemented GB Camera mapper a while back (I remember it was a mix of previous mappers I think?)
https://github.com/project64/project64/blob/master/Source/nragev20/GBCart.cpp#L1059

bsmiles32:
Didn't know there was already an implementation. Will put it on my todo list.
Or if you want to work on it, be my guest :)

LuigiBlood:
I guess it could easily be done by myself. When I'll have some time to spend I could try to see.
I definitely remember just copy pasting code and change a little bit of it and got away with it.

bsmiles32:
Rebased against master.
Added convert back function.

bsmiles32:
Rebased against master

bsmiles32:
Rebased against master

LuigiBlood:
```
if (w == 0xe848d316 || w == 0x2263ee56) {
    DebugMessage(M64MSG_WARNING, "Loading a saved disk ");
}
```

What does this mean, actually?

bsmiles32:
I guess it is 64 disk magic constant check like
https://github.com/project64/project64/blob/def0550dede289f7702c5613b69f796e7ff68ac8/Source/Project64-core/N64System/N64DiskClass.cpp#L116

LuigiBlood:
Actually that magic number is the disk region. e848d316 is Japanese, 2263ee56 is US (unreleased but we have the US IPL so why not)

Narann:
> Actually that magic number is the disk region. e848d316 is Japanese, 2263ee56 is US (unreleased but we have the US IPL so why not)

A perfect candidate for some constants here. :D

bsmiles32:
Updated against master.

bsmiles32:
Rebased against master. Added savestate support, SDK and MAME dump format support, preliminary save support (but disabled because of severe slowdown).

SM64DD: works OK
DKDJ: seems to boot (hear music, brief "loading" message), but can't get in game
DKIJ: dito
DKKJ: dito
DMBJ: seems OK - emit warnings in menu screen with JCMD warning 23 01 14 (not DD related)
DMGJ: seems OK
DMPJ: goes to menu, but can't do much
DMTJ: seems OK
DPGJ: goes up to press start screen, then freezes
DRDJ: goes to press start screen, but doesn't react to any presses
DSCJ: boots, goes in-game, but many missing graphics
EFZJ: boots, but complains about missing F-ZEROX cart, even if it is plugged :(

@richard42 Even if support is really basic, maybe we could get this (and the ui-console part) merged so that our users can enjoy it. (And I wouldn't have to keep rebasing every now and then...). What do you think ?

LuigiBlood:
DMPJ: Cursor is invisible? Plugin's fault. Must enable "Render frame buffer as texture" or something like that. Otherwise it should work okay.
DPGJ: That's what I got on parallel-n64.
DRDJ: That's normal it wants the modem pak.
DSCJ: Is a pain to emulate
EFZJ: You should boot F-Zero X cart. Cart takes priority over IPL.

bsmiles32:
@LuigiBlood Very helpful thanks ! Indeed when I force FZ to boot on the cartridge things start to work. With the dynarec, it works fine. With the pure interpreter and cached interpreter I get some strange issue where it doesn't get past the 64DD loading part. Very strange (and not reassuring...).

For this test, I forced booting on the cart for FZ (in device.c and plugin.c), but proper implementation would be "if DD disk only, boot on DD, otherwise use cart ROM". The problem is that this DD disk only case is not possible currently as ui/core loading sequence requires a ROM.

LuigiBlood:
That's pretty much how it should work, yeah. That's where it gets complicated.

bsmiles32:
Implemented in a hackish way the boot device selection: we boot on cart unless 64DD is plugged and the plugged cartridge is not a combo game (eg cart+disk). Be sure to plug a Japanese cart otherwise 64DD region checks will fail. With that F-Zero now works on dynarec (but some bugs prevent it to work on pure,cached interpreter).

I guess, we're now at almost parity with libretro/cen64/MAME 64DD wise. There are still some core bugs that needs to be ironed out. Then I'll look at disk swapping and saving, which will bring us closer to PJ64 level of support.

fzurita:
@bsmiles32 how does this deal with game specific settings when it's always the IPL ROM being loaded?

src/device/device.c
```diff
@@ -152,11 +145,25 @@ void init_device(struct device* dev,
         { A(MM_PI_REGS, 0xffff), M64P_MEM_PI, { &dev->pi, RW(pi_regs) } },
         { A(MM_RI_REGS, 0xffff), M64P_MEM_RI, { &dev->ri, RW(ri_regs) } },
         { A(MM_SI_REGS, 0xffff), M64P_MEM_SI, { &dev->si, RW(si_regs) } },
+        { A(MM_DOM2_ADDR1, 0xffffff), M64P_MEM_NOTHING, { NULL, RW(open_bus) } },
```
fzurita: Should we really be expanding usage of these macros? It feels like they really obfuscate the code.

bsmiles32: These macros allows for less repetition while still being somewhat readable.
Also they are scoped (undef'd after table declaration) to avoid unwanted usage.

Surely it is much easier to read/modify than what we had before the big memory refactoring :)

fzurita: Yes, true, easier to modify. But I almost have to manually expand the macro to wrap my head around. You do get used to it, but it's a pain at first.

bsmiles32:
@fzurita The IPL ROM is inside the disk drive itself, not the cartridge, nor the disk. It is used as a boot menu and store some hard-coded assets. In the core we use it's presence as a test for enabling DD emulation. Beside the disk drive itself, the user can insert a disk (in the console-ui you use the --dd-disk argument), which is the effective program that is run (when no cartridge is concurrently inserted).
Note: there is currently a limitation in the console-ui that requires a cartridge ROM even if strictly speaking, none is mandated in this case.
Some disk games are also meant to be played with a matching cartridge (eg F-Zero-X) to have the full experience, in this case, the console boots on the cartridge, which can then use use the disk at will.

The heuristic we use to notify the GFX plugin of what game is played is here:
https://github.com/bsmiles32/mupen64plus-core/blob/740406ff156ce2b8867f7c416996404e2f0981bb/src/plugin/plugin.c#L233

Clearly, it is not perfect [disks cannot be discriminated], but given that there is no "header" in the 64DD disk, we don't have much choice.
Maybe we could, change the API to pass a truly unique game identifier (like SHA-256) so that plugins know what game is loaded, and there what settings to apply. Or embed in the core some "fake headers" for the disk games (based on their cartridge port) to make the GFX plugin happy. Given that there are only a few disk-only games, hard-coding them may still be acceptable.

fzurita:
My question was going to lead to do so you answered some of the quetions I had. What about loading appropriate settings for a disk from mupen64plus.ini?

bsmiles32:
I think the MD5 computation for identification in mupen64plus.ini is still based on the cartridge ROM [1], so for now, it is not possible to load game-specific parameters for disks.

[1] https://github.com/bsmiles32/mupen64plus-core/blob/740406ff156ce2b8867f7c416996404e2f0981bb/src/main/rom.c#L154

bsmiles32:
But you can cheat and insert a "compatible" cartridge ROM (japanese, otherwise some regional check can fail)...

fzurita:
What about this? For 64DD cartridges, we could combine the IPL+cartrige or ROM+cartridge together to come up with one big md5?

LuigiBlood:
If you do the big MD5 just mind that there are portions you just cannot do any checksums on because it's different on each disk.

fzurita:
One big MD5 still doesn't fix the ROM header issue. I guess a ROM header could be faked out by the core based on the combined MD5.

fzurita:
@LuigiBlood is there such thing as a "pristine" disk image where nothing has been written yet? It would be nice if an emulator could direct any writes to a separate file that indexes into the pristine disk image.

bsmiles32:
Since there is not really any choice when combining a cart and a disk, I don't think a big MD5 brings much. However, generating a "fake" header based on a hash of the "read-only part" of the disk, could work.

LuigiBlood:
@fzurita I did exactly that on Project64. But I didn't say it's not possible to do a MD5. Depending on Disk Type you can actually make it work. Just ignore System Area (different dates on each disk) and RAM area (save data).

loganmc10:
I think this should be merged soon, it can always be improved upon later, it's all pretty self contained so it's not going to cause issues with other games, lets not make @bsmiles32 keep rebasing this until the end of time.

fzurita:
@richard42 can this be merged? I would love to add support to Mupen64plus AE.

LuigiBlood:
I'm literally just waiting for this to be merged so I can work on the promised Game Boy Camera Transfer Pak support.

Narann:
I think @bsmiles32 has the permissions to merge. Do you wait for anyone's approval first ?

loganmc10:
This PR caused a new issue in Coverity:

```
*** CID 278587:  Null pointer dereferences  (FORWARD_NULL)
/src/device/dd/dd_controller.c: 425 in read_dd_regs()
419         switch(reg)
420         {
421         case DD_ASIC_CMD_STATUS: {
422                 /* clear BM interrupt when reading gap */
423                 if ((dd->regs[DD_ASIC_CMD_STATUS] & DD_STATUS_BM_INT) && (dd->regs[DD_ASIC_CUR_SECTOR] > SECTORS_PER_BLOCK)) {
424                     clear_dd_interrupt(dd, DD_STATUS_BM_INT);
>>>     CID 278587:  Null pointer dereferences  (FORWARD_NULL)
>>>     Passing "dd" to "dd_update_bm", which dereferences null "dd->idisk".
425                     dd_update_bm(dd);
426                 }
427             } break;
428         }
429     }
430     
```

richard42:
I saw this also, but disregarded it. If dd is NULL it will crash at line 423 anyway, so who cares about crashing inside the dd_update_bm() function on line 425?

bsmiles32:
@LuigiBlood I've been working on GB cam support with promising results while waiting for this to be merged :

gbcam_wip

It still needs some cleaning, but I should PR soonish :)

LuigiBlood:
I wanted to do it once you were done with DD but I guess you did it yourself instead :P

hissingshark:
Hi, it's the first time this year I've rebuilt from master and I'm getting a segfault, which I've bisected back to commit `e302bc2` "Add support for 64 Disk Drive".  Not opened a separate issue as this still seems to be a work in progress.

Mario Kart shows the spinning logo at the start but segfaults just before the start screen.
Super Smash Bros segfaults during the intro animation.

I'm on armv8, Amlogic S905x (Mali 450 GPU).  GLES2.  And I was using the mupen64plus-video-n64 plugin.

Anything I can provide to assist debugging?

Issue #445, opened 2017-10-19
Can't change pak type after game has started using CONTROL.plugin API


fzurita:
The old API to change the pak type doesn't seem to work. It seems to be getting stuck in rumble pak mode. One example is Diddy kong racing.

fzurita:
Ok, I think I see what it's happening. It's only allowing pak types that are in the mupen64plus.ini?

bsmiles32:
Yep ! Pak switching is only between compatible paks. If there is a deeper issue let me know because I only tested this pak switching with mupen64plus-input-sdl and raphnet. The core only uses the Controls[i].Plugin at startup, then switch between compatible paks when the PAK_SWITCHING_BUTTON (0x4000) is pressed.

fzurita:
Yeah, I'm seeing that now. The old API allowed switching paks by using ```CONTROL``` plugin field. We only seem to read that at startup.

bsmiles32:
> If a non compatible Pak is selected, should the core use "none" instead?
Currently, I don't know what the result of that (because I hadn't considered this case when implementing this).
I think we have 2 options there :
- like you said, as the plugin is not compatible, the core ignore that and switch to none (or the first compatible pak) with a warning.
- or "trust" the user, and add, for this run, the selected pak to the list of compatible paks and proceed as normal.

fzurita:
Any reason to not always load all pak types for all games? In a real system, there was nothing preventing someone from plugging in a rumble pak into any game.

bsmiles32:
Only switching between compatible paks reduces the number of presses the user has to do for selecting the right pak. Especially convenient for games that are compatible with only 1 pak. And you don't gain anything by inserting a non compatible pak.

At least that was the idea.

fzurita:
Yeah, I get it. Unfortunately, that is forcing frontends to use a key to switch Pak type. Not all devices have keys though, for example, touchscreen Android phones.

In Android, we are depending on a user to enter a menu while in game and select a Pak type using touch. We then update the ```CONTROL.plugin``` value to the user selected one.

fzurita:
I created this pull request: https://github.com/mupen64plus/mupen64plus-core/pull/449

fzurita:
@bsmiles32 your update to my fix almost works. In smash brothers, if the controller Pak selected at startup is the memory Pak, the core courses the rumble Pak instead of none.

PR #449, opened 2017-10-21
Allow switching controller paks using CONTROLS.plugin API

fzurita:
Allow switching controller paks using CONTROLS.Plugin API. Instead of reading the Plugin parameter once at startup, it will constantly be read, like old behavior.

Also, if an invalid pak is selected for a game, it will use "None" instead of the first available. This is more similar to older behavior.

This addresses https://github.com/mupen64plus/mupen64plus-core/issues/445

### DIFF
--- a/src/api/m64p_plugin.h
+++ b/src/api/m64p_plugin.h
@@ -138,7 +138,7 @@ typedef struct {
 typedef struct {
     int Present;
     int RawData;
-    int  Plugin;
+    int Plugin;
 } CONTROL;
 
 typedef union {
--- a/src/backends/plugins_compat/input_plugin_compat.c
+++ b/src/backends/plugins_compat/input_plugin_compat.c
@@ -30,6 +30,8 @@
 
 #include <stdint.h>
 #include <string.h>
+#include <api/m64p_plugin.h>
+#include <api/callbacks.h>
 
 /* XXX: this is an abuse of the Zilmar Spec normally this value is reserved */
 enum {
@@ -44,6 +46,8 @@ enum {
 enum { PAK_SWITCH_DELAY = 20 };
 enum { GB_CART_SWITCH_DELAY = 20 };
 
+int PrevControlPlugin[4];
+
 static int is_button_released(uint32_t input, uint32_t last_input, uint32_t mask)
 {
     return ((input & mask) == 0)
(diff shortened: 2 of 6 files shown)


fzurita:
If possible, I would appreciate some testing of the "Next Pak" button. I did not test that functionality since my setup doesn't make that easy to test.

fzurita:
@bsmiles32 what do you think about this?

loganmc10:
I haven't gotten a chance to test this yet, but this:
```
static int switchToNextPak[4] = {0};
static int switchToSpecificPak[4] = {0};
```

Is probably not a good way to initialize the variables if you're counting on that setting them to 0. Remember that the core can technically stop and start multiple ROMs without ever being unloaded

fzurita:
I'm thinking that should be fine. The individual members of those arrays get set back to zero after paks are done switching.

If you start changing paks and then before paks are done changing, the game is switched, then things may get weird. You would start the second game with the second available Pak instead of the first. But I think that would be ok.

fzurita:
@bsmiles32 any chance of merging this? Without it, the Android input plugin doesn't work correctly.

bsmiles32:
@fzurita Sorry it took me so long to review your PR. See https://github.com/fzurita/mupen64plus-core/pull/1 for comments and fixes. Can you test if selecting the pak type from menu still works as intended on your end ?

Also, I'm a bit hesitant here (not your PR, just overall design I ended up with) because, now we have 2 ways to handle pak switching and I don't like that. I originally put that "pak cycling" logic inside the core for convenience (core knows about which paks are compatible for a given game) but doing so I resorted to the hacky, input-sdl specific, PAK_SWITCH_BUTTON to trigger the pak switch. I feel like fixing these hacks would be easier if/when we get rid of input plugins and integrate them inside the core.

fzurita:
Thanks for rewriting my fix :). I'll check if it works correctly still with the Android input plugin.

fzurita:
@bsmiles32 you changes work correctly on my device. I updated the pull request with your changes.

One less hacky way of doing this maybe be to have  the special button press change ```CONTROLS.plugin``` instead of changing the selected Pak directly. That way, there is only one method of doing things.

fzurita:
@bsmiles32 are we ok with merging this after your changes? Or would you rather come up with a less hacky solution?

bsmiles32:
@fzurita Merging this now, because it fixes a bug. Properly addressing this can come later as it is not a user-facing issue.
Also, changing the Controls[i].Plugin from the core might not be a so good idea. If you change that from the core, the input plugin can get "confused" that one of its variable changed "magically"...

PR #454, opened 2017-10-24
CountPerOp=1 for Gauntlet Legends

loganmc10:
Allows for better performance when using GLideN64. Still needs a fairly quick CPU, but the audio is really bad when using GLideN64 HLE with CountPerOp=2

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -5581,6 +5581,7 @@ Players=4
 SaveType=None
 Mempak=Yes
 Rumble=Yes
+CountPerOp=1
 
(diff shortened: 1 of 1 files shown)


AmbientMalice:
Does this cause negative side effects with Angrylion's beside being more demanding? I haven't had time to test.

loganmc10:
I get nowhere near full speed using angrylion, and mupen64plus has no fps counter using angrylion so it's hard to say what is working faster. It would be good for someone to test, I'm not exactly sure why it seems to perform so much better using 1, using the flickery version in glide64mk2 it seems ok at 2

PR #455, opened 2017-10-25
Complete (U) save types + Add four games

Enker:
The retail (U) save types should be accurate now :). I checked the board pics found here: http://s9.zetaboards.com/Nintendo_64_Forever/topic/7464768/1/
Note: I couldn't find a pic for Monster Truck Madness 64

I tested the ones I changed to confirm the save type.


AmbientMalice:
mupen64plus uses the entry RefMD5= to reduce duplication. If different variations of the game use the same settings, refer them to the MD5 hash of the "master" version.

Enker:
I thought it was only for alternate dumps of the same version? I can change it if it's better that way.
Most rom V1.1/etc have the full settings currently.

AmbientMalice:
For example, Rogue Squadron has this:

[47CAC4E2A6309458342F21A9018FFBF0]
GoodName=Star Wars - Rogue Squadron (U) (M3) (V1.0) [!]
CRC=66A24BEC 2EADD94F
Players=1
SaveType=Eeprom 4KB
Rumble=Yes

GoodName=Star Wars - Rogue Squadron (U) (M3) (V1.1) [!]
CRC=C7F30CFA ECB0FA36
RefMD5=47CAC4E2A6309458342F21A9018FFBF0

Enker:
I added another save fix and made the four V1.1 games use the RefMD5 settings format.

bsmiles32:
@Enker Thanks for your work on the mupen64plus ini file. This is tedious work and we greatly appreciate your contributions.

RefMD5 indeed allow to inherit keys/values from a parent entry. This is typically used for several dumps of the "same" rom. However there is no clear rule (that I know of) which say what should inherit from what: should a rom revision inherit from the original version ? should a rom released in a different region inherit from another and which one (U/E/J/???)... I don't know. Maybe @richard42 knows more about that ?

Enker:
@bsmiles32 No problem :). I'm glad I could help improve the emulator's compatibility.

Yeah, it would be nice to have a description of how RefMD5 should be used. It's currently almost exclusively used for roms with the same region+version.

Issue #456, opened 2017-10-25
Waialae Country Club won't save


Enker:
This game needs Sram forced in the settings, but M64P seems to ignore it and use auto-detection instead.
It will show a Controller Pak screen if it has the wrong save type, but this game doesn't actually support the Controller Pak. PJ64 needs Sram forced for this game as well.

loganmc10:
Yeah I see what we're doing wrong here, I'll submit a PR in the next day or so

Enker:
I found another odd thing with this game. If the .sra file is zero-filled, then the game options will all be OFF by default.

Filling the file with any non-zero data makes it behave normally. You can also select Setup Data/Initialize Data to fix the options.

loganmc10:
Is it the same in Project64? I've wondered about this kind of thing actually. I mean it's feasible that the manufacturer could have shipped the game with the SRAM pre-populated with some data, the ROM file wouldn't contain that information

Enker:
I tested with PJ64. I assume it's the same with M64P.

loganmc10:
Yeah it's the same in M64P. I don't really know much about sram/flashram. I know for the mempaks we do initialize it with non-zero data:

https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/controllers/paks/mempak.c#L35-L54

But I have no idea where that information came from, I assume somebody figured it out by reading the data off a fresh mempak.

Like I said, as far I understand, the manufacturer could have shipped this game with the SRAM pre-populated with some data, there is really no way for us to know. Even if we found somebody with a cartridge, it's not going to be a "fresh" cartridge, so I don't know how we'd ever figure this one out

loganmc10:
Actually now that I look, both the eeprom and flashram are actually initialized with all 1's, but the sram is given all 0's.

Initializing the sram to all 1's like is done with eeprom/flashram fixed the issue. I'll need to test the other sram games to make sure there are no issues there

Enker:
Cool. I'm not sure what data the real carts have. In order to check it, we'd need a backup from a game that doesn't use the entire 32KB of Sram.

PR #457, opened 2017-10-25
Add support for SRAM cpu read/write

loganmc10:
This will fix https://github.com/mupen64plus/mupen64plus-core/issues/456 once https://github.com/mupen64plus/mupen64plus-core/pull/455 is merged.

The auto-detection scheme we use to detect SRAM/flashram doesn't seem to work with Waialae Country Club.

This change will take the value in ROM_SETTINGS if it can, but still try to fallback to the auto-detection scheme if ROM_SETTINGS doesn't specify the type.

This PR adds support for reading/writing to sram via read_cart_dom2/write_cart_dom2, Waialae Country Club does this.

Finally, it gets rid of read_cart_dom2_dummy/write_cart_dom2_dummy, and maps the whole DOM2 address space to read_cart_dom2/write_cart_dom2. I'm not sure why the dummy functions are there, perhaps someone knows the back story?

So far I've tested Paper Mario and Waialae Country Club, I'll do some more testing. I'm hoping @bsmiles32 can comment on this since you seem to be pretty familiar with the memory mapping and just refactored this code

### DIFF
--- a/src/device/cart/sram.c
+++ b/src/device/cart/sram.c
@@ -74,3 +74,24 @@ unsigned int sram_dma_write(void* opaque, uint8_t* dram, uint32_t dram_addr, uin
     return /* length / 8 */0x1000;
 }
 
+void read_sram(void* opaque, uint32_t address, uint32_t* value)
+{
+    struct sram* sram = (struct sram*)opaque;
+    const uint8_t* mem = sram->istorage->data(sram->storage);
+
+    address &= SRAM_ADDR_MASK;
+
+    *value = *(uint32_t*)(mem + address);
+}
+
+void write_sram(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
+{
+    struct sram* sram = (struct sram*)opaque;
+    uint8_t* mem = sram->istorage->data(sram->storage);
+
+    address &= SRAM_ADDR_MASK;
+
+    *(uint32_t*)(mem + address) = value & mask;
+
+    sram->istorage->save(sram->storage);
+}
--- a/src/device/cart/sram.c
+++ b/src/device/cart/sram.c
@@ -32,7 +32,7 @@
 
 void format_sram(uint8_t* mem)
 {
-    memset(mem, 0, SRAM_SIZE);
+    memset(mem, 0xff, SRAM_SIZE);
 }
 
 void init_sram(struct sram* sram,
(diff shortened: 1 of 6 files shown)


loganmc10:
Ok I've tested quite a few games now, the only thing I notice is now in quite a few Flash RAM games (noticed it in Jet Force Gemini, Pokemon Snap, Pokemon Stadium) I get a bunch of these messages:

```
Core Warning: unknown flashram command: 0
```

I added this commit: https://github.com/mupen64plus/mupen64plus-core/pull/457/commits/1264b548822a29277e7ca1d88d610bf69ff4aa3e to just ignore those writes, I guess this was taken care of by the "dummy write" before.

Project64 handles this situation by treating it the same as 0xf0000000 (FLASHRAM_MODE_READ) https://github.com/project64/project64/pull/1039, but I didn't see any reason why they chose that, since our current method isn't causing any issues that I know of, I figured we might as well keep the same behavior as before

loganmc10:
I added another commit to initialize the sram to 0xff (the same that is done for eeprom and flashram), see https://github.com/mupen64plus/mupen64plus-core/issues/456#issuecomment-339459518

I tested a number of sram games like Ocarina of Time, Mario Golf, Resident Evil 2. I didn't go too far into these games though. More testing on this is always welcome.

AmbientMalice:
I'd recommend testing Fushigi no Dungeon, since it writes to flash ram literally every time you take a step.

loganmc10:
Yeah it seems to work fine, I started a game and reloaded it without issue.

src/device/cart/sram.c
```diff
@@ -74,3 +74,24 @@ unsigned int sram_dma_write(void* opaque, uint8_t* dram, uint32_t dram_addr, uin
...
+}
+
+void write_sram(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
+{
+    struct sram* sram = (struct sram*)opaque;
+    uint8_t* mem = sram->istorage->data(sram->storage);
+
+    address &= SRAM_ADDR_MASK;
+
+    *(uint32_t*)(mem + address) = value & mask;
```
bsmiles32: use instead:

```
masked_write((uint32_t*)(mem + address), value, mask);
```

bsmiles32:
Looks good to me except for the masked_write in write_sram. Good work !

loganmc10:
Updated to use masked_write

PR #459, opened 2017-10-27
Fix reset

bsmiles32:
@Gillou68310 noticed that reset has been broken since remove_memd PR and bisected it at PIF Boot ROM rewrite. But a deeper analysis revealed several point of regressions (at least 3) and general weakness of the reset implementation. I'm not sure this PR fixes all regressions related to reset as I can still trigger some cascade of warnings at times. But I haven't been able to exactly track down why it is like that and what broke it.

### DIFF
--- a/src/device/r4300/interrupt.c
+++ b/src/device/r4300/interrupt.c
@@ -449,6 +449,9 @@ void nmi_int_handler(void* opaque)
     g_gs_vi_counter = 0;
     init_interrupt(&r4300->cp0);
 
+    dev->vi.next_vi = cp0_regs[CP0_COUNT_REG] + dev->vi.delay;
+    add_interrupt_event_count(&r4300->cp0, VI_INT, dev->vi.next_vi);
+
     // clear the audio status register so that subsequent write_ai() calls will work properly
     dev->ai.regs[AI_STATUS_REG] = 0;
     // set ErrorEPC with the last instruction address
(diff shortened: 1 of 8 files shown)


src/device/r4300/interrupt.c
```diff
@@ -460,6 +449,9 @@ void nmi_int_handler(void* opaque)
     g_gs_vi_counter = 0;
     init_interrupt(&r4300->cp0);
 
+    dev->vi.next_vi = cp0_regs[CP0_COUNT_REG] + dev->vi.delay;
```
loganmc10: Sorry I'm not at my computer for the next few days, does setting next_vi to 0 work here (and don't schedule any interrupt). I would think that just setting next_vi to 0 should still allow the VI interrupts to restart, the method you have here might cause them to restart too soon

bsmiles32: Just setting next_vi = 0 (and removing the subsequent add_interrupt_event_cont) still breaks the reset unfortunately.

loganmc10:
Ok I tested this out a bit, hard reset seems to work reliably, although if I do a hard reset, the core will freeze when shutting down and print a bunch of these messages: ```Core Warning: Instruction pointer is 0 at dyna_stop()```

I still can't really get soft reset to work reliably (tested Super Mario 64), I think I got it to work once, but usually I'll just get a bunch of ```Core Warning: Unaligned dword read``` messages

Issue #458, opened 2017-10-27
FlashRam saving fails in Derby Stallion 64


Enker:
Just keep pressing A to test the saving. PJ64 fixed this issue by changing the "flash status" bytes. They also have a RDB setting to make the (Beta) rom boot.

According to the cartridge reader tool made by sanni, there are three different flash chips used by N64 games. https://github.com/sanni/cartreader

Can these be detected by M64P? They each have a slightly different flash status, so it might help other games as well.

bsmiles32:
Fixed with PR #460

PR #460, opened 2017-10-27
Add support for different flashram types.

bsmiles32:
Default to MX29L1100 as it is the most common one and is what pj64 uses.

I don't understand Japanese, but it looked like Derby Stallion could save stuff so it is a potential fix for https://github.com/mupen64plus/mupen64plus-core/issues/458
Will need some more testing with other flashram games though before being merged.

### DIFF
--- a/src/device/cart/flashram.c
+++ b/src/device/cart/flashram.c
@@ -46,11 +46,11 @@ static void flashram_command(struct flashram* flashram, uint32_t command)
         break;
     case 0x78000000:
         flashram->mode = FLASHRAM_MODE_ERASE;
-        flashram->status = 0x1111800800c20000LL;
+        flashram->status[0] = 0x11118008;
         break;
     case 0xa5000000:
         flashram->erase_offset = (command & 0xffff) * 128;
-        flashram->status = 0x1111800400c20000LL;
+        flashram->status[0] = 0x11118004;
         break;
     case 0xb4000000:
         flashram->mode = FLASHRAM_MODE_WRITE;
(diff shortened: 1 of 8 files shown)


src/device/cart/flashram.c
```diff
@@ -87,11 +87,11 @@ static void flashram_command(struct flashram* flashram, uint32_t command)
         break;
     case 0xe1000000:
         flashram->mode = FLASHRAM_MODE_STATUS;
-        flashram->status = 0x1111800100c20000LL;
+        flashram->status[0] = 0x11118001;
         break;
     case 0xf0000000:
         flashram->mode = FLASHRAM_MODE_READ;
-        flashram->status = 0x11118004f0000000LL;
+        flashram->status[0] = 0x11118004;
```
rlabrecque: Any chance you could name these or something?

bsmiles32: Unfortunately there is no datasheet for these flashram chips so I don't really know the meaning of these bits. However, I came across a "close cousin" (namely MX29L1611) which had some similarities with those used by big N. In its datasheet "0xc2" was the manufacturer code, plus @sanni (https://github.com/sanni/cartreader/blob/master/Cart_Reader/N64.ino#L1819) indicates that different products gets different answers. That's how I ended up with this PR with lower status bits for manufacturer code and product id.

Enker:
I found two games that don't use the MX29L1100 flashram chip.
(U) board pics: http://s9.zetaboards.com/Nintendo_64_Forever/topic/7464768/1/
A few (E) board pics are here: http://n64.icequake.net/mirror/tcsr2001.tripod.com/

Jet Force Gemini (E) = 29L1101KC-15B0
WWF No Mercy (U) (V1.0) = MN63F81MPN, V1.1 is unknown

bsmiles32:
Did some more test with many flashram based games and no issue so far. Also it effectively fixes Derby Stallion saves. So I think it is ready for merge.

loganmc10:
Looks good to me, this is basically how it works in PJ64 so we knows it's received a lot of testing

PR #461, opened 2017-10-28
Get rid of USE_SDL flag

loganmc10:
I added this flag a while ago as the beginning of an effort to make SDL usage optional in the core. It never got very far, other parts of the core still rely on SDL, so there is no use for this flag.

Also, I never actually added the USE_SDL flag to the MSVC project files, so the MSVC version probably hasn't been dealing with the SDL mutex at all.


loganmc10:
Here is the original PR for reference: https://github.com/mupen64plus/mupen64plus-core/pull/202

Issue #467, opened 2017-11-03
Some devices are failing CoreStartup with M64ERR_NO_MEMORY


fzurita:
@bsmiles32 It seems like recent change is breaking many Android devices. Specifically this commit:

https://github.com/mupen64plus/mupen64plus-core/commit/143cfa7139faf757c7d0e508291e34c24a1fecbc#diff-fa9734f1550e974c80b15a8a6fed3ad7

A lot of devices are failing with ```M64ERR_NO_MEMORY``` on CoreStartup.

Is there any way to fallback to the old behavior if we are unable to allocate that much RAM?

So far this error has been limited to Samsung devices.

bsmiles32:
I think the problem here is that either for their device overcommitting is disabled or that their malloc implementation doesn't use an anonymous mmap internally. I think replacing the malloc/free with mmap/munmap (VirtualAlloc/VirtualFree for Win32) will fix this. Will do a PR when I get back home.

fzurita:
The above didn't work for them, it instead crashed when accessing the memory. I'm having them try your pull request next.

bsmiles32:
@fzurita Can you try this : https://github.com/mupen64plus/mupen64plus-core/pull/474

bsmiles32:
@fzurita Can you try this : https://github.com/mupen64plus/mupen64plus-core/pull/474

bsmiles32:
@fzurita Can you try this : https://github.com/mupen64plus/mupen64plus-core/pull/474

fzurita:
Ok, I'll let you know when I can.

Issue #277, opened 2017-05-15
Implement "fast memory" accesses


bsmiles32:
See http://www.redream.io/docs/memory-access for a good explanation of what I mean by "fast memory".

The basic idea is to mmap the CPU address space in such a way that normal memory accesses are direct and accesses to MMIO trigger a pagefault handler which in turn will "call" (not really that in practice) the right r/w handler. As outlined in the article, the difficulty lies in the details, especially what can and cannot be done inside a pagefault handler, but as other other implementation exists, this shouldn't be too difficult.

I expect a noticeable speedup from this.

ricrpi:
That is certainly an interesting way of doing MMIO. I was thinking of having a second memory mapped region for a callback function to register into. Then in the dynarec, when a read or write instruction is emitted, additional instructions are added to check for a call back function (and call if not NULL)

PR #468, opened 2017-11-03
Use virtual memory to allocate g_mem_base (if possible).

bsmiles32:
This should fix issue https://github.com/mupen64plus/mupen64plus-core/issues/467

Basically, it replaces (if possible) the big malloc with virtual memory allocation in such a way that no unneeded physical pages gets allocated. When available (-DHAVE_MMAP) we use the mmap/munmap/mprotect functions, otherwise if it is a Windows plateform we use the VirtualAlloc/VirtualFree functions and otherwise only rely on the C standard library malloc/free and hope that internally, given the size of the alloc, it will use virtual memory.

I have tested myself on Linux with and without mmap, but couldn't test other platforms.

### DIFF
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -160,6 +160,7 @@ endif
 ifeq ($(OS), FREEBSD)
   TARGET = libmupen64plus$(POSTFIX).so.2.0.0
   SONAME = libmupen64plus$(POSTFIX).so.2
+  CFLAGS += -DHAVE_MMAP
   LDFLAGS += -Wl,-Bsymbolic -shared -Wl,-export-dynamic -Wl,-soname,$(SONAME)
   LDLIBS += -L${LOCALBASE}/lib -lc
   ifeq ($(ARCH_DETECTED), 64BITS)
@@ -171,6 +172,7 @@ endif
 ifeq ($(OS), LINUX)
   TARGET = libmupen64plus$(POSTFIX).so.2.0.0
   SONAME = libmupen64plus$(POSTFIX).so.2
+  CFLAGS += -DHAVE_MMAP
   LDFLAGS += -Wl,-Bsymbolic -shared -Wl,-export-dynamic -Wl,-soname,$(SONAME)
   LDLIBS += -ldl
   # only export api symbols
@@ -187,6 +189,7 @@ ifeq ($(OS), OSX)
   OSX_SDK_PATH = $(OSX_SDK_ROOT)/$(shell ls $(OSX_SDK_ROOT) | tail -1)
 
   TARGET = libmupen64plus$(POSTFIX).dylib
+  CFLAGS += -DHAVE_MMAP
   LDFLAGS += -framework CoreFoundation -dynamiclib
   LDLIBS += -ldl
   ifeq ($(ARCH_DETECTED), 64BITS)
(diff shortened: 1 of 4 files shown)


fzurita:
This didn't work. Allocation succeeds, but the core crashes when trying to use the memory allocated. I'll try to get a back trace.

fzurita:
Ok, I must had been looking at the wrong log file. It didn't crash, but the allocation failed. We got the ```M64ERR_NO_MEMORY``` error code. I made sure that the mmap portion of the code was being used by removing the malloc and Windows versions.

I think it's the mmap called that failed because I never saw "Failed to change memory protection."

bsmiles32:
Rebased against master, fixed the C99-ism (variable declaration inside for loop).

@fzurita Does the workaround suggested in [1] works ? (eg reducing dalvik.vm.heapsize to increase chances of having a continuous 512M chunk available ?).

[1] https://stackoverflow.com/questions/33897711/android-mmap-fails-with-out-of-memory

fzurita:
This is what I found suggesting that:

https://github.com/android-ndk/ndk/issues/449

It may be that it only affects mmap64.

I did not find the page you found. That's a nice find. I'll look into that.

bsmiles32:
> android-ndk/ndk#449

In this link they use mmap for file-backed memory, not anonymous allocation. So I think the issue was more that in their device your users didn't have a continuous 512M block available in their virtual address space. Trying the suggested workaround might help (and maybe even without this PR).

fzurita:
@bsmiles32 The proposed solution requires a rooted device. It's not something I can change for the app. It's something users have to do on their devices.

fzurita:
Currently, about 25% of the users of Mupen64plus-AE on Android are using 5.0 to 5.1.

They are probably all affected. Since there is nothing I can do from the app itself, it's probably better to have some kind of workaround.

One option would be for Mupen64plus-AE to roll back to the older version of the core.

Another option would be to keep two versions of the core, an old one and a newer up to date one. In the app, if I detect Android 5.0 or 5.1, I can make the app use the older version of the core.

loganmc10:
I thought most Linux systems used mmap under the hood anyway. If malloc is returning NULL, I don't think you're going to have much luck tricking it into succeeding. Android has a mind of its own when it comes to memory management.

From looking at the code, the only real benefit g_mem_base currently offers is getting rid of the if statement in fast_mem_access, but I believe the plan was to implement this: https://github.com/mupen64plus/mupen64plus-core/issues/277, which would make having some kind of fallback system probably quite complex.

Anyway if https://github.com/mupen64plus/mupen64plus-core/issues/277 isn't going to happen, then I'd say get rid of g_mem_base, but if that is going to happen, this is quite a tricky problem...

fzurita:
What kind of speed up would #277 give?

bsmiles32:
I will push another attempt when I get back home tonight, which tries the big 256M alloc first and if it fails does a smaller alloc (~73M) and uses some if/else if/else for selecting the right address.

In my wip proof of concept of fast mem I wasn't able to get much speed up for the pure/cached interpreter because of the overhead of page faults (< 2% speedup if I remember correctly). However, for the dynarec we can optimistically emit the page-fault way and if it triggers too much regen the block with the non page-fault way, I expect maybe a 10% speedup, but as you say this may not be the bottleneck, so meh. Anyway, this work is still very early WIP and may not be successfull after all.

bsmiles32:
Replaced with another approach in https://github.com/mupen64plus/mupen64plus-core/pull/474

Issue #469, opened 2017-11-04
Hey You Pikachu corrupt graphics


loganmc10:
![hey_you _pikachu-000](https://user-images.githubusercontent.com/848146/32401202-028a2948-c0d0-11e7-9ed4-68bc377ebc62.png)

Happens during the "Press start" screen, that yellow blob should be Pikachu

I bisected the issue to 4ee13fefbffb0da74cf7515412275a7d4d5c3176 @bsmiles32

loganmc10:
To confirm this issue, here are the results from PeterLemon's (https://github.com/peterlemon/n64) SW test with the latest master:

![cpu_test_cpu_sw-000](https://user-images.githubusercontent.com/848146/32402432-aad00604-c0e9-11e7-85fe-8966e84e497f.png)

PR #471, opened 2017-11-05
Fix LWL, LWR, LDL, LDR, SWL, SWR, SDL, SDR implementations

bsmiles32:
Fixes https://github.com/mupen64plus/mupen64plus-core/issues/469

Did similar fixes for other unaligned memory accesses to avoid undefined behavior. But I suspect there are more bugs in there because all these instructions don't look symmetrical...

### DIFF
--- a/src/device/r4300/mips_instructions.def
+++ b/src/device/r4300/mips_instructions.def
@@ -70,10 +70,10 @@
  */
 
 #define BITS_BELOW_MASK32(x) ((UINT32_C(1) << (x)) - 1)
-#define BITS_ABOVE_MASK32(x) (~((UINT32_C(1) << (x)) - 1))
+#define BITS_ABOVE_MASK32(x) (~(BITS_BELOW_MASK32((x))))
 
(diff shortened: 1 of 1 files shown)


loganmc10:
Hey You Pikachu does appear to be fixed, but now the emulator fails the LWL test (it passed before). It now passes the SWL test.

cpu_test_cpu_lw-000

bsmiles32:
@loganmc10 When I try to run this test rom, I don't get anything printed on screen (GLideN64) what is your configuration to get that running ?

loganmc10:
Yeah I use Angrylion (https://github.com/ata4/angrylion-rdp-plus) + rsp-cxd4, the text is hard to read unless you set ViMode=1 in the config for angrylion plus

bsmiles32:
@loganmc10 Thanks for the info.
Updated PR with fixed LWL, LDL instructions.

loganmc10:
Yep, Hey You Pikachu works, and it passes all the PeterLemon SW/LW tests, looks good to me

PR #472, opened 2017-11-06
If an invalid controller pak is selected, use NONE instead

fzurita:
If an invalid controller pak is selected for a game, use no controller pak instead.

See #445 

This way if a user selects an invalid controller pak, they don't get an unexpected one instead.


fzurita:
@bsmiles32 is this ok?

bsmiles32:
👍

PR #474, opened 2017-11-08
Big alloc fallback

bsmiles32:
Another try at the memory allocation failure encountered in some Android devices. As this is likely due to a shortage of virtual memory, in this PR we provide a fallback mechanism which requires much less memory.


loganmc10:
Tested both modes on Linux, no issues

fzurita:
I have confirmation from a user with the issue that this now works correctly! I can only confirm that it works correctly on Android devices that don't have the issue. This should be merged soon.

PR #475, opened 2017-11-08
Fix handlers[i].{read,write}32 comparison in dynarec.

bsmiles32:
Edit: There was another regression introduced also during the memory refactoring; handlers array is now indexed with physical addresses instead of physical, so we need to mask higher bits.

So everything should be good now. I used Derby Stallion for testing because it crashed with the dynarec almost instantaneously.

---- Previous message

I'm trying to fix computation of handlers[i].{read,write]32 inside dynarec but not done yet.
*It crashes dynarec for now*

Offsets computation are wrong now because each element of handlers array is sizeof(struct mem_handler), not just sizeof(void*). Given that struct mem_handler is basically 3 pointers we just need to multiply by 3 the array index to fix the computation. At least that is what I think, but doing so seems to crash... so anyone has a suggestion ?

Commit message below :

This regression has been introduced in
commit 82b20c8a1b48cf09a081b31a7a963fc1985bf86a.

Implementation assumes that sizeof(mem_handler) is 3 * sizeof(void*)
therefore we just need to multiply (address >> 16) by 3 to get the right
offset. Multiplication by 3 can be achieved with lea dst, [reg + 2*reg]
instruction.

### DIFF
--- a/src/device/r4300/x86/assemble.h
+++ b/src/device/r4300/x86/assemble.h
@@ -552,6 +552,13 @@ static osal_inline void mov_reg32_preg32pimm32(int reg1, int reg2, unsigned int 
    put32(imm32);
 }
 
+static osal_inline void lea_reg32_preg32x2preg32(int reg1, int reg2, int reg3)
+{
+   put8(0x8D);
+   put8(0x04 | (reg1 << 3));
+   put8(0x40 | (reg2 << 3) | reg3);
+}
+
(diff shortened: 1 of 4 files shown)


bsmiles32:
@loganmc10 Can you test this please :)

loganmc10:
Yes the crashing seems to be fixed now! I've got FBInfo working with the dynarec again.

I only tested x64, I've never had much luck cross compiling the x86 build on Linux

PR #476, opened 2017-11-11
Preliminary refactorings 

bsmiles32:
This PR is just a collection of preliminary refactoring commits extracted from PR #446. They are not directly related to 64DD so we can incorporate them hopefully sooner.

### DIFF
--- a/src/device/cart/cart.c
+++ b/src/device/cart/cart.c
@@ -100,7 +100,6 @@ void init_cart(struct cart* cart,
                /* cart ROM */
                uint8_t* rom, size_t rom_size,
                struct r4300_core* r4300,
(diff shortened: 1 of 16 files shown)


bsmiles32:
Added a last minute fix which should help regarding soft reset "locking".

loganmc10:
Looks good to me, the only difference besides the soft reset fix that I noticed is that force_detected_rdram_size_hack() is now called after the Pi DMA write, it's done before the write currently (if I'm reading correctly). I assume it makes no difference

bsmiles32:
Doing the force rdram detection hack before or after the DMA is not important, both are fine indeed.

PR #477, opened 2017-11-11
Fix FBInfo

loganmc10:
This should be ready to merge.

This doesn't fix FBInfo with the new dynarec, it writes directly to RDRAM, so someone who is more familiar with the new dynarec will need to find a way to have it call write_rdram_fb/read_rdram_fb for the addresses in fb->infos[].addr

You can test this out in Mario Kart (Luigi's Raceway with the monitor), or Dr Mario (the pills), or Pokemon Snap (the red dot) using a recent version of GLideN64.

If someone knows a better/more eloquent way to determine the actual address & length in write_rdram_fb() I'm all ears, right now I just have a switch statement going through the various mask possibilities.

FBInfo in general still has some issues, Super Mario and Donkey Kong both freeze, but they also freeze in 1964, which is supposed to have a good (but slow) implementation of FBInfo. The puzzle transition in Banjo-Kazooie also has issues, but it seems to be timing related, changing CountPerOp will affect how much of the puzzle transition appears. But I figured if we are going to support this feature, we should have a good implementation, even if the GFX plugin could probably still use some work

### DIFF
--- a/src/device/rdp/fb.c
+++ b/src/device/rdp/fb.c
@@ -21,6 +21,7 @@
 
 #include "fb.h"
 
+#include "api/callbacks.h"
 #include "api/m64p_types.h"
 #include "device/memory/memory.h"
 #include "device/r4300/r4300_core.h"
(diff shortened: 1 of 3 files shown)


bsmiles32:
Can confirm that Mario 64 freezes, Dr Mario pills are working, and that I can see something in the Luigi RaceWay screen, but it seems upside down for me. Still it is an improvement compared to a black screen, so you are on a good track. Will try to look a bit closer :)

mariokart64-000

bsmiles32:
I did a similar attempt at fixing FB info in https://github.com/bsmiles32/mupen64plus-core/tree/fbinfo
Basically :
- there was a bug in how we track "dirty pages"
- taking into account the write mask fixes Dr Mario pills
- looking at the spec, it seems that we need to notify the gfx plugin after having written a value to rdram not before. But it doesn't seem to matter that much as I saw no real change during my limited testing.

However contrary to your PR I didn't change the dirty page tracking part. I think here the spec is ambiguous. In our implementation we divide the RDRAM into 0x800 * 4k pages and mark these as dirty if gfx plugin says that they contain some fb data. Then before notifying the GFX plugin about CPU reads, we check if the corresponding page has already been read and if not notify the GFX plugin and mark it as non dirty. This kind of page tracking allows for efficient tracking of page dirty-ness.

However, it seems that GLideN64 understand the spec differently (and it is understandable) and doesn't track pages the same way, the 4k range starts at the address of access making the tracking of the 4k ranges more difficult.

Given that, at the time of writing the spec computers were more modest, I think that the intended page tracking is the one we have in mupen64plus aka 0x800 * 4k pages.

Also what concerns me is that in the spec we mandate a 4k page read, but do we have guarantees that it won't overwrite any non FB data ?

loganmc10:
Awesome I'll try it out in a bit. I agree about the 4k block, but I noticed that GLideN64 did "address + 0x1000". Not even 1964 implements the dirty page tracking, so I think I can talk to gonetz and get him to switch to using 4k blocks.

> Also what concerns me is that in the spec we mandate a 4k page read, but do we have guarantees that it won't overwrite any non FB data ?

The GFX plugin should know where the Framebuffer ends, so I think it's up to the GFX plugin to not write outside the actual FB. GLideN64 does have a related problem however, in that sometimes FBRead will be called for an address that is no longer being used as a Framebuffer, but the GFX plugin doesn't realize that, so it copies data that it shouldn't into rdram. GLideN64 hasn't really figured this out yet, and it's probably why some games are freezing

loganmc10:
Also I think:
```
for (j = fb_mapping.begin >> 12; j < (fb_mapping.end >> 12); ++j)
```
 Should be ```<= end``` right?

bsmiles32:
Good catch ! fb_mapping.end is inclusive.
Will fix

loganmc10:
One other thing, are you sure you've got the masks right? I noticed you did it opposite of me, if you look at the red dot in Pokemon Snap, it looks off in your version:

pokemon_snap-001

Here is the diff to fix it:
```
             switch(mask)
             {
             case 0x000000ff:
-                addr += 3;
+                addr += 0;
                 size = 1;
                 break;
             case 0x0000ff00:
-                addr += 2;
+                addr += 1;
                 size = 1;
                 break;
             case 0x00ff0000:
-                addr += 1;
+                addr += 2;
                 size = 1;
                 break;
             case 0xff000000:
-                addr += 0;
+                addr += 3;
                 size = 1;
                 break;
 
             case 0x0000ffff:
-                addr += 2;
+                addr += 0;
                 size = 2;
                 break;
 
             case 0xffff0000:
-                addr += 0;
+                addr += 2;
                 size = 2;
                 break;
```

then it looks like this:

pokemon_snap-000

bsmiles32:
I always get confused regarding byte/word endianness issue. Seeing that your code gets better results, I think you're probably right. However then I would correct it with :
```
addr += {0,1,2,3}^S8 // for byte accesses
addr += {0,2}^S16 // for hword accesses
```

loganmc10:
Ok Pokemon Snap looks good with those changes. I think your branch is good to be merged. There are still issues on the GFX plugin side, and there might be other reads/writes we need to track (technically the RSP DMA or PI DMA could read/write to the FB address space as well), but for now this is good I think

PR #478, opened 2017-11-12
Fix FbInfo 2

bsmiles32:
This PR does almost the same as PR #477 but doesn't change the dirty page tracking method.


loganmc10:
Did you see that this has an AppVeyor failure? It didn't like the "inline" I think

loganmc10:
I think you can use ```osal_inline``` instead of ```inline```

bsmiles32:
Done: 067f45cfc4d8c737af7aaf4d7d468a29cfc50b9a

PR #481, opened 2017-11-14
Add more FB writes and reads

loganmc10:
There are a few other instances where the framebuffer might be modified. This tracks reads/writes in the PI and SP DMA's.

If you test the latest master version of GLideN64 with this PR Super Mario 64 no longer freezes.

It exposes another bug with have with the dynarec. When using the Cached Interpreter with FBInfo, you can see Mario's head in the Super Mario 64 intro. If you use the 64-bit dynarec, the head is missing


loganmc10:
This should be ready to merge and relatively safe as well, it only affects FBInfo

PR #483, opened 2017-11-18
Extract PIF, RCP and RDRAM modules

bsmiles32:
Just moving some files around to isolate PIF and RCP modules.


bsmiles32:
Also extracting RDRAM module.

PR #485, opened 2017-11-20
Modify DPC_STATUS_FREEZE hack

loganmc10:
Right now the hack works like this:

If DPC_STATUS_FREEZE bit is set when RSP task should run, run the RSP task when DPC_STATUS_FREEZE is unset.

This works mostly ok, but we did have to add ```dp->sp->regs[SP_STATUS_REG] &= ~SP_STATUS_YIELD;``` to get Perfect Dark working in LLE, as well as the ```!get_event(&dp->r4300->cp0.q, SP_INT)``` check to get it working in HLE. Also, the "puzzle effect" in Banjo-Kazooie is half-black when using FBInfo in GLideN64.

What this change does:

If DPC_STATUS_FREEZE is set when RSP task should run, still run the RSP task, but delay the DP Interrupt until the DPC_STATUS_FREEZE bit is unset. The RSP should still run even in the RDP is frozen.

If a VI Interrupt happens when DPC_STATUS_FREEZE is set, delay updating the screen until its unset. This should simulate the RDP not finishing until the DPC_STATUS_FREEZE bit is unset.

I tested every Rare game in HLE and LLE and they all work still, no hacky checks like we had before. Additionally, the "puzzle effect" in Banjo-Kazooie now works with FBinfo in GLIdeN64.

I updated the savestate to include this, I noticed we aren't actually using the v1.2 save states currently, I updated the file to start using v1.2, @bsmiles32 I don't know if you held that back on purpose or not.

Testing is welcome, Rare games are the only ones I know of that use this bit. Additionally, it seems like many LLE GFX plugins are unsetting the DPC_STATUS_FREEZE bit (GLideN64 and angrylion-plus do this). I submitted a PR to angrylion-plus to stop doing this (https://github.com/ata4/angrylion-rdp-plus/pull/41) and I'll submit one to GLideN64 as well, unsetting that bit breaks this system.

### DIFF
--- a/src/device/device.c
+++ b/src/device/device.c
@@ -177,7 +177,7 @@ void init_device(struct device* dev,
             &dev->mi, &dev->ri, &dev->pif.cic);
     init_ri(&dev->ri, &dev->rdram);
     init_si(&dev->si, &dev->mi, &dev->pif, &dev->ri);
-    init_vi(&dev->vi, vi_clock, expected_refresh_rate, &dev->mi);
+    init_vi(&dev->vi, vi_clock, expected_refresh_rate, &dev->mi, &dev->dp);
 
     init_pif(&dev->pif,
         (uint8_t*)mem_base_u32(base, MM_PIF_MEM),
--- a/src/device/rcp/rdp/rdp_core.c
+++ b/src/device/rcp/rdp/rdp_core.c
@@ -29,11 +29,8 @@
 #include "device/rcp/rsp/rsp_core.h"
 #include "plugin/plugin.h"
 
-static int update_dpc_status(struct rdp_core* dp, uint32_t w)
+static void update_dpc_status(struct rdp_core* dp, uint32_t w)
 {
-    /* see do_SP_Task for more info */
-    int do_sp_task_on_unfreeze = 0;
-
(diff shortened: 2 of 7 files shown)


bsmiles32:
Also tested this on my end (Mario Kart, OoT, MajorasMask, IndianaJones, Mischief Maker, ...). Didn't find any obvious regression. I also tested Knife Edge Nose Gunner, but it is still running too fast.

As for the savestate version, it is an oversight on my part.

Jj0YzL5nvJ:
This pull cause shaky graphics on Banjo-Kazooie by using `rsp-cxd4` +` GLideN64` with HLE mode.

loganmc10:
> This pull cause shaky graphics on Banjo-Kazooie by using rsp-cxd4 +GLideN64 with HLE mode.

This does seem to be the case, however the shaky graphics don't happen with rsp-hle, and they don't happen in LLE mode with any plugin, so it would probably be safe to assume that a change could be made to rsp-cxd4 to fix that issue

purpasmart96:
Yeah this makes rsp-cxd4 freak out on the banjo games, works with mupens hle rsp though

Issue #447, opened 2017-10-19
Removing the GoldenEye TLB hack (Pure Interpreter)


loganmc10:
I'm referring to this bit of code:

https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/tlb.c#L125-L150

GoldenEye actually works with no issues without that code using the Cacher Interpreter or the dynarec. It exits using the Pure Interpreter with this:

```
Core Error: reserved opcode: 8005DD90:1
```
Which it gets to from here:
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/pure_interp.c#L337

@bsmiles32 I'm wondering if you know why this is? It seems strange to me that the Pure Interpreter would arrive at an opcode that the cached interpreter wouldn't

bsmiles32:
This is new to me. I did try some time ago to get rid of this hack, but without it the game refused to boot (but I don't remember which emumode I used). I was more of the impression that the root cause was that PAGEMASK was not honored as it returns (tlb_lookup & ~0xfff) | (address & 0xfff) in the general case and returns (some_base_address + (address & 0xffffff). But I didn't get that far with the investigation, and got distracted by other parts of the core.
But, having the pure_interpreter behaving differently than other emumode is worrying... it's a bug ! Maybe bisecting when this different behavior emerged could help.

loganmc10:
I remember coming across this a long time ago when I was trying it out, I think it's worked without the hack in the cached interpreter and dynarec for a long time. I'm not sure about the new dynarec, which employs the same hack inside its code

loganmc10:
I also tested the x86 new dynarec with the ```tlb_hacks()``` function removed, as well as the bits from ```virtual_to_physical_address()```, and it still works as normal, so it seems like it's just the Pure Interpreter that currently needs that hack to function.

loganmc10:
@bsmiles32 I think the issue relates to how fast_mem_access() works. If an address can't be found via the TLB, virtual_to_physical_address() returns 0. fast_mem_access() then returns mem_base + address (0), and then the Pure Interpreter uses that address to interpret the Opcode.

That doesn't seem right to me, this means that the interpreter will try and execute the opcode found at mem_base + 0 if the TLB can't find the address. I assume it's executing the wrong code, which is causing the game to stop.

[...]

PR #487, opened 2017-11-23
Remove GoldenEye TLB hack

loganmc10:
Fixes https://github.com/mupen64plus/mupen64plus-core/issues/447

This will need to be tested on every dynarec/interpreter. I've tested the Pure/Cached Interpreter as well as the 64-bit dynarec. I tested the x86 new dynarec a while ago, but it probably should be tested again. ARM should also be tested ( @fzurita )

The only game that should be affected is GoldenEye, however I did make a change to how the Pure Interpreter works generally.

### DIFF
--- a/src/device/device.c
+++ b/src/device/device.c
@@ -95,7 +95,6 @@ void init_device(struct device* dev,
     unsigned int emumode,
     unsigned int count_per_op,
     int no_compiled_jump,
-    int special_rom,
     int randomize_interrupt,
     /* ai */
     void* aout, const struct audio_out_backend_interface* iaout,
@@ -167,7 +166,7 @@ void init_device(struct device* dev,
     init_rdram(&dev->rdram, mem_base_u32(base, MM_RDRAM_DRAM), dram_size);
 
     init_r4300(&dev->r4300, &dev->mem, &dev->mi, &dev->rdram, interrupt_handlers,
-            emumode, count_per_op, no_compiled_jump, special_rom, randomize_interrupt);
+            emumode, count_per_op, no_compiled_jump, randomize_interrupt);
     init_rdp(&dev->dp, &dev->sp, &dev->mi, &dev->mem, &dev->rdram, &dev->r4300);
     init_rsp(&dev->sp, mem_base_u32(base, MM_RSP_MEM), &dev->mi, &dev->dp, &dev->ri);
     init_ai(&dev->ai, &dev->mi, &dev->ri, &dev->vi, aout, iaout);
--- a/src/device/device.h
+++ b/src/device/device.h
@@ -109,7 +109,6 @@ void init_device(struct device* dev,
     unsigned int emumode,
     unsigned int count_per_op,
     int no_compiled_jump,
-    int special_rom,
     int randomize_interrupt,
     /* ai */
     void* aout, const struct audio_out_backend_interface* iaout,
--- a/src/device/r4300/new_dynarec/new_dynarec.c
+++ b/src/device/r4300/new_dynarec/new_dynarec.c
@@ -1678,7 +1678,7 @@ static void remove_hash(int vaddr)
 #error Unsupported dynarec architecture
 #endif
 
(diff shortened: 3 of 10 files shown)


loganmc10:
Just re-tested the x86 new dynarec and it still works without the hack

fzurita:
The game seems to start and run. Performance seems a lot worse though, at least in Android. Was this a performance hack?

loganmc10:
Well it's possible when it was first implemented it was necessary for some reason, but now it is basically a performance hack.

GoldenEye is a strange game in that is will purposely trigger TLB exceptions (it will try to execute code that isn't TLB mapped, which triggers a TLB exception, which will then map the TLB entry, and then try to execute the code again).

The hack bypasses the TLB exceptions, and points to the address needed directly. So yes I would except that CPU usage would go up since all these TLB exceptions need to be processed.

It still runs fine on my laptop, the new dynarec implements its own TLB code so there is the chance that this affects the new dynarec more negativity than the other methods? I'm not really sure.

loganmc10:
So potentially we could leave the hack in place for the new dynarec, since that is mostly targeted at mobile devices? thoughts @bsmiles32 @Gillou68310 ?

sgorman:
It was implemented back in the day during the initial versions of the
project for performance enhancements as Logan mentioned the ROM does some
kind of mapping that wasn't being handled correctly, there was a big demand
by some communities for some of the tweaks to appear in the emulator
system. I think the solution is to implement the correct exception handlers
for the TLB, otherwise you will probably notice a big performance hit as
mentioned in the comments. Good luck.

- ODS

fzurita:
That does bring up the question. Could this hack be applied to other TLB games with bad performance? For example, Conker's bad Fur day?

loganmc10:
Ok I modified the PR to leave the new dynarec as it is (I just renamed tlb_hacks() to tlb_speed_hacks() to make it clear that it's not needed for accuracy).

loganmc10:
> Could this hack be applied to other TLB games with bad performance? For example, Conker's bad Fur day?

I assume so, it looks like Conker's is doing a very similar thing (same company so it makes sense):

It does this (it's slightly different since GoldenEye actually maps the TLB into the ROM address space, Conkers looks like it's using the RDRAM address space):
```
tlb exception !!! @ 151e8620, 2, add:15019f08
map 151e8000 : 8035afff
tlb exception !!! @ 151ebb50, 2, add:151e865c
map 151eb000 : 8035bfff
tlb exception !!! @ 150432bc, 2, add:151ebcb8
map 15042000 : 802d0fff
map 15043000 : 8033efff
tlb exception !!! @ 151ec164, 2, add:151ebd04
map 151ec000 : 8035cfff
map 151ed000 : 8035ffff
tlb exception !!! @ 150417ac, 2, add:150436b0
map 15041000 : 8035dfff
tlb exception !!! @ 1517f4d8, 2, add:15019c3c
map 1517e000 : 80328fff
map 1517f000 : 8034bfff
tlb exception !!! @ 150347e8, 2, add:15018c6c
```

You can see a TLB exception happens, and then it maps the area into the TLB. If we pre-mapped those entries, the TLB exceptions would never happen.

If you look at the very top example, what you want to have happen is that when the game asks for address 0x151e8620, you want the function to return 0x8035a620 as the "TLB match"

For GoldenEye they figured out a simple pattern that would match all cases. I don't immediately see a simple pattern, but I'm not great at that sort of thing

fzurita:
It may explain why N64oid had so much better performance in some TLB games.

In wonder if a hack like this could be automated.

fzurita:
Just a thought. Is this the performance improvement that bsmiles32 fastmem is supposed to give?

fzurita:
I did some more research on fastmem, it seems like it would be great for TLB games. So the answer to my question would be a yes.

psyke83:
Thanks - PR sent. https://github.com/mupen64plus/mupen64plus-core/pull/491

PR #490, opened 2017-11-28
Warnings

bsmiles32:
With this PR all warnings except those about unused parameters should be gone. This is tested on GCC with -Wall -Wextra -pedantic.


bsmiles32:
@richard42 rebased against master

PR #494, opened 2017-12-02
Update Makefile

StenApp:
First try to get it fixed for macOS


StenApp:
Needed for files_macos.c to compile and set paths right

PR #495, opened 2017-12-02
Create files_macos.c

StenApp:
First try to get a proper Mupen64Plus app on macOS

### DIFF
--- a/files_macos.c
+++ b/files_macos.c
@@ -189,7 +189,7 @@ const char * osal_get_user_configpath(void)
     /* append the given sub-directory to the path given by the environment variable */
     if (path[strlen(path)-1] != '/')
         strcat(path, "/");
-    strcat(path, "Application Support/Mupen64Plus/");
+    strcat(path, "Library/Application Support/Mupen64Plus/");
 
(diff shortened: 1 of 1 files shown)


StenApp:
Doesn't work completely,  `CFCopyHomeDirectoryURL()` is iOS only...

richard42:
looks pretty good. I can't merge this until the ios-only function is fixed though

StenApp:
Looking on how to get the path anyhow.

StenApp:
Grrrrrrrr!

save folder goes to ApplicTsave in user Library instead of created folder /Users/currentuser/Library/ApplicationSupport/Mupen64Plus/save

need help!

richard42:
this change looks okay. You might also want to check that getenv("HOME") returns a non-NULL pointer, initialize the path[] array to an empty string, and keep the fallback to the getpwuid() function, with something like this:
`    if (strlen(path) == 0)
    {
        struct passwd* pwd = getpwuid(getuid());
        if (pwd)
            strcpy(path, pwd->pw_dir);
    }`

PR #496, opened 2017-12-03
Various GB cart enhancements.

bsmiles32:
With this work we get :
- partial support for gb camera (no image capture yet, but may come in the future)
- mbc2 support
- mbc5 rumble support

### DIFF
--- a/projects/VisualStudio2013/mupen64plus-core.vcxproj
+++ b/projects/VisualStudio2013/mupen64plus-core.vcxproj
@@ -61,6 +61,7 @@
     <ClCompile Include="..\..\src\device\controllers\paks\rumblepak.c" />
     <ClCompile Include="..\..\src\device\controllers\paks\transferpak.c" />
     <ClCompile Include="..\..\src\device\gb\gb_cart.c" />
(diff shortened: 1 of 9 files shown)


loganmc10:
Won't modifying the save state like that break compatibility with current save states? Won't it throw off the offsets of the data if you don't add it to the end?

That being said, we only actually started using v1.2 a couple days ago so if this was merged soon it probably won't affect many people

bsmiles32:
Yeah it will break recent savestates, but anyway, people using a non released version should expect breakages unfortunately.

fzurita:
Shouldn't we try to maintain compatibility with older save states?

loganmc10:
Since we don't have a consistent release schedule, and I don't even know if another release is ever going to happen, I tend to agree.

In this case, it'll only break save states made since Nov 28 (when 7abc81852f6e5879407c2fe20349058481113260 was merged). It should also only break save states where a gb cart is present, save states without a gb cart shouldn't be affected based on my understanding

fzurita:
Ok, that's good, not a big deal then.

PR #501, opened 2017-12-14
Implement enough of RDRAM subsystem to remove rdram detection hack.

bsmiles32:
Previously, the RDRAM detection/initialization was wrong and a post
RDRAM initialization hack was triggered (just before initial PI DMA)
to force specified amount of RDRAM to be recognized.

With this PR, several aspects of the RDRAM have been implemented which
results in proper detection of RDRAM by the IPL3:
- support for up to 8 (IPL3 hardcoded limitation) RDRAM modules
- partial support for individual rdram address mapping,
ADDR_SELECT has not been implemented though
- support broadcast register writes
- simulating read failure when current calibration is not done yet


bsmiles32:
cc : @tj90241 I guess you could be interested in this ?

loganmc10:
I tested a good number of games with the expansion pak on and off. Didn't see any issues, everything seemed to work as expected.

PR #503, opened 2017-12-17
Device: RDRAM: Use the R4300 helpers to get the R4300 registers

Miouyouyou:
Since the "regs" member is not available on the r4300_core
structure when enabling ARM DYNAREC, accessing this member
directly will cause compilation errors with this setup.

Signed-off-by: Myy Miouyouyou <myy@miouyouyou.fr>


bsmiles32:
Thanks for your PR. It looks good and was indeed an oversight from my part.
Unfortunately our AppVeyor instance is stuck at the moment so CI checks will take a while before completing. But after, I'll merge that.

Good Work !

PR #510, opened 2018-01-09
Cleanings

bsmiles32:
This is just some more internal source cleanings such as reindentation, forward declaration, header inclusion or just moving some code around.


fzurita:
@bsmiles32 This broke the new dynarec build in Android ARM. You are just missing this include in new_dynarec.c:
```
#include "device/r4300/x86/assemble.h"
```

bsmiles32:
Could you post the compiler error please so I can check what declaration is missing.
Also, it seems counter intuitive that the ARM dynarec would require x86 specific declarations...
I can see for instance, EAX being needed in syscall_assemble, do_ccstub without proper ifdef x86/endif, so is that a bug or using this in the ARM dynarec is ok ?

cc: @Gillou68310

fzurita:
Yes, it was complaining about a missing ```EAX``` declaration. Even in the arm build, that was pointing to the x86 assemble.h in the old code.

fzurita:
I can't give you the exact error right now since I'm not close to a computer I can use to build.

PR #511, opened 2018-01-09
Remove write_cart_rom

bsmiles32:
This doesn't seem to be needed anymore.
All-Star Baseball 2001, International Track & Field 2000, Midway's
Greatest Arcade Hits Vol. 1 are still playable.

Something (maybe in the TLB handling code) must have made these bootable again, without the write_cart_rom hack(?).

See https://github.com/mupen64plus/mupen64plus-core/issues/311, https://github.com/mupen64plus/mupen64plus-core/issues/221

### DIFF
--- a/src/device/cart/cart_rom.c
+++ b/src/device/cart/cart_rom.c
@@ -46,8 +46,6 @@ void init_cart_rom(struct cart_rom* cart_rom,
 
 void poweron_cart_rom(struct cart_rom* cart_rom)
 {
-    cart_rom->last_write = 0;
-    cart_rom->rom_written = 0;
 }
 
 
@@ -56,22 +54,7 @@ void read_cart_rom(void* opaque, uint32_t address, uint32_t* value)
     struct cart_rom* cart_rom = (struct cart_rom*)opaque;
     uint32_t addr = rom_address(address);
 
-    if (cart_rom->rom_written)
-    {
-        *value = cart_rom->last_write;
-        cart_rom->rom_written = 0;
-    }
-    else
-    {
-        *value = *(uint32_t*)(cart_rom->rom + addr);
-    }
-}
-
-void write_cart_rom(void* opaque, uint32_t address, uint32_t value, uint32_t mask)
-{
-    struct cart_rom* cart_rom = (struct cart_rom*)opaque;
-    cart_rom->last_write = value & mask;
-    cart_rom->rom_written = 1;
+    *value = *(uint32_t*)(cart_rom->rom + addr);
 }
 
 unsigned int cart_rom_dma_read(void* opaque, const uint8_t* dram, uint32_t dram_addr, uint32_t cart_addr, uint32_t length)
--- a/src/device/cart/cart_rom.h
+++ b/src/device/cart/cart_rom.h
@@ -34,9 +34,6 @@ struct cart_rom
     uint8_t* rom;
     size_t rom_size;
 
-    uint32_t last_write;
-    uint32_t rom_written;
-
     struct r4300_core* r4300;
 };
 
@@ -52,7 +49,6 @@ void init_cart_rom(struct cart_rom* cart_rom,
 void poweron_cart_rom(struct cart_rom* cart_rom);
 
 void read_cart_rom(void* opaque, uint32_t address, uint32_t* value);
-void write_cart_rom(void* opaque, uint32_t address, uint32_t value, uint32_t mask);
 
 unsigned int cart_rom_dma_read(void* opaque, const uint8_t* dram, uint32_t dram_addr, uint32_t cart_addr, uint32_t length);
 unsigned int cart_rom_dma_write(void* opaque, uint8_t* dram, uint32_t dram_addr, uint32_t cart_addr, uint32_t length);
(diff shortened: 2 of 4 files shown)


loganmc10:
I went back to before https://github.com/mupen64plus/mupen64plus-core/issues/311 was merged and removed write_cart_rom and it actually makes these games boot even then. The issue they had was because they were trying to write "0" to the ROM and that wasn't registering, but it seems that just disabling all writes also fixes the issue.

I assume there are other games that are doing this, it's just that International Track & Field 2000 and Midway's Greatest Arcade Hits Vol. 1 were trying to write "0" which didn't work before. PJ64 has a similar hack in place. It would be interesting to know the history behind it.

But anyway, I actually tested about 20-25 games and I didn't see any issues.

loganmc10:
I looked into this, mupen64 and 1964 also had/have this hack in place.

The 1964 source code has:

```
NeedToApplyRomWriteHack = FALSE;
	if( strnicmp(currentromoptions.Game_Name, "A Bug's Life", 12) == 0 ||
		strnicmp(currentromoptions.Game_Name, "Toy Story 2", 11) == 0 )
	{
		NeedToApplyRomWriteHack = TRUE;
		TRACE0("Using Rom Write Hack");
	}
```

Sure enough, this PR breaks A Bug's Life and Toy Story 2

They print a bunch of these kinds of messages:
```
Core Error: Invalid PIF address: 1FC0FF88
Core Error: Invalid PIF address: 1FC0FF88
Core Error: Invalid PIF address: 1FC0FF90
Core Error: Invalid PIF address: 1FC0FF90
Core Error: Invalid PIF address: 1FC0FF98
Core Error: Invalid PIF address: 1FC0FF98
Core Error: Invalid PIF address: 1FC0FFA0
Core Error: Invalid PIF address: 1FC0FFA0
Core Error: Invalid PIF address: 1FC0FFA8
Core Error: Invalid PIF address: 1FC0FFA8
Core Error: Invalid PIF address: 1FC0FFB0
```

Perhaps this hack represents some kind of actual hardware bug that someone figured out long ago? Or maybe there is a clue in the PIF errors, I'm not exactly sure.

loganmc10:
Just a bit more info, I asked zilmar about this on Discord and he said:

> I tried it on a real n64, that is the behaviour my n64 gave

> if i remember correct, it is like it stored any write to rom in like memory, then  the next read, read that back

> if you read it again it is back to normal

bsmiles32:
@loganmc10 Awesome investigation work ! Thanks !

Closing this as this cart write behavior is indeed needed.

jago85:
I've just read this conversation after I stumbled across the write_cart_rom-function. I was curious what this function is good for and investigated what happens on real hardware. I think I've found some details that could be interesting for you.

First of all, I think that the implemented behaviour is wrong. I don't know exactly why you need this, but the behaviour is impossible on a real cart. The write signal is not even connected to the ROM chip. I think the guy who claimed he had tested it on a real N64 just forgot to wait for the flag PI_STATUS_IO_BUSY to reset.

I've tried 3 test cases:
1. Write a value to 0xB0000000 (first cart rom addr) and read it back immediately
2. Write a value to 0xB0000000, wait while PI_STATUS_IO_BUSY is set and than read it back
3. Write a value to 0xB0000000 when a button is pressed, read it back, when another button is pressed (don't check the flag)

The results were the following:
1. The written value is read back (tried 0x00000000 and 0x5555AAAA), the flag PI_STATUS_ERROR is set atfer this. I had a look at the data bus with a logic analyser. The read request isn't even transmitted to the cart. Seems that the PI is then returning the last written value.
2. The read value is 0x80371240 (the error flag is not set)
3. like 2.) because the buttons can't be pressed that fast and there is at least one frame between the accesses (which is much longer than the PI is busy)

After having a deeper look with the logic analyzer at the write coming from Midway's Greatest Arcade Hits Volume 1 (U) I saw that it writes 0 to 0x1000010C (which is 0xB000010C) and than 0x49533634 to 0x10000100 (which is 0xB0000100). 0x49533634 is ASCII for "IS64" so this likely has something to do with the IS64 dev kit. After this it read multiple times from 0x1000010C and always gets 0x3C0FA3F0 back from the cart which is the ROM data at this address.

So I think you really could get rid of this function unless there is a serious reason to keep it. Or you could emulate the correct behaviour of the PI and restore the normal read functionality after the PI is not busy anymore. However, this is only necessary if a software uses the PI incorrectly. In the current state, the game definitely gets different data than in the real hardware.

Here are some of my logic analyzer images. For those who know about the PI hardware signals: The read and write signals are inverted to be displayed active high. ad_ff2 is the multiplex bidirectional address/data bus either carrying addresses or write data from the console to the cart or read data from the cart to the console. The value of cart_addr is calculated by the analyzer.

This is an image of the whole process from Midway's Greatest Arcade Hits Volume 1 (U):
(more reads from the same address follow after the end of the sample buffer)


Zoomed into the write of "IS64" to 0x10000100 (marked the data 0x49533634 coming from the console):


And the read from 0x1000010C:


This is my test case 2:


And finally test case 1:
You can see that the read access is missing, because it was launched while the PI was busy:

loganmc10:
@jago85 thanks so much for looking into this. Sorry it took this long to implement. As far as I can tell there are actually 2 games that attempt to read the written value while IO_BUSY is still set (A Bug's Life and Toy Story 2). In those games, case 1 applies, and they expect to get the last written value back.

Another game we came across (WCW Backstage Assault), writes to the ROM and then attempts to read back the value after IO_BUSY is clear, so case 2 applies

Issue #512, opened 2018-01-09
Tetris 64 (J) freezes on boot.


AmbientMalice:
Works on PJ64.
![tetris64-000](https://user-images.githubusercontent.com/10704701/34748594-0e1f04ba-f5e9-11e7-96fa-bafb2b7005f9.png)

loganmc10:
Interesting. I know the unique thing about this game is the use of the bio sensor.

I see a screenshot here: http://www.raphnet-tech.com/products/raphnetraw/index.php

Showing it working with mupen64plus. So either it used to work, or it only currently works with the raphnet plugin. I'll look into it more tomorrow

loganmc10:
So this is a regression, which appeared when we introduced the use of the SI_DMA_BUSY flag. It seems to be affected by the timing of the SI interrupt. Shortening the SI interrupt delay fixes this game. If I make the interrupt delay too short, Body Harvest won't boot, if I make it too long, this game won't boot.

I never found an "overlapping" value where both games worked. I'll need to investigate further.

bsmiles32:
Can confirm that commit d9523052c8400fb284bd40128c37636c8cb2fe1a "Add SI_STATUS_DMA_BUSY" broke Tetris 64.

bsmiles32:
Tetris can boot with a SiDmaDuration <=170 (RandomizeInterrupt=False). Given that the interrupt timing jitter is <64 when RandomInterrupt=True, I took a safe value of 100 to make sure it can boot even with this option enabled. And with RandomInterrupt=True we get to enjoy the random selection of screen transitions :)

bsmiles32:
PR #513 has been merged, closing this.
Please re-open if the issue persists.

PR #513, opened 2018-01-10
Add SiDmaDuration option.

bsmiles32:
This is a hack on SI DMA timings to allow games with different timing
requirements to work.

We keep the previously hardcoded value of 0x900 as the default value
but fine-tune it for Tetris 64 (J) which can be made to boot with a
lower value of it.

Therefore we can get both Body Harvest and Tertis 64 (J) to work.

Temporary fix for #512


bsmiles32:
@richard42 This PR fixes a regression and should have no border effect, so I'd like to have it in the coming release if that's fine for you. Thanks

loganmc10:
Looks good to me, tested and played a few short games, no issues.

Given that the timing of these interrupts is mostly guesswork, there may not be a better option.

bsmiles32:
Fixed self-referencing RefMD5 for ROM derived from Tetris 64 (J)

Issue #489, opened 2017-11-26
Reported crashes


fzurita:
Unfortunately, I can't get save states or game names from this, but maybe it's enough to fix a few issues.

As of this revision: bbe9007ac83ac14bdc63db509b6879e4ae6667f8
```
********** Crash dump: **********
pid: 0, tid: 0 >>> org.mupen64plusae.v3.fzurita <<<
#00  pc 00000000000b3a34  libmupen64plus-core.so: Routine dma_sp_write at ./mupen64plus-core/src/device/rsp/rsp_core.c:52
#01  pc 0000000000046d94  libmupen64plus-core.so: Routine mem_write32 at ./mupen64plus-core/src/device/memory/memory.h:110
#02  pc 00000000000a9dcc  libmupen64plus-core.so: Routine write_word_new at ./mupen64plus-core/src/device/r4300/new_dynarec/new_dynarec.c:11218
#03  pc 0000000000000eac  <anonymous>

```

```
********** Crash dump: **********
pid: 0, tid: 0 >>> org.mupen64plusae.v3.fzurita <<<
#00  pc 000000000003c618  libmupen64plus-core.so: Routine r4300_cp0_regs at ./mupen64plus-core/src/device/r4300/cp0.c:88
#01  pc 000000000003d224  libmupen64plus-core.so: Routine add_interrupt_event_count at ./mupen64plus-core/src/device/r4300/interrupt.c:159
#02  pc 00000000000b4fdc  libmupen64plus-core.so: Routine vi_vertical_interrupt_event at ./mupen64plus-core/src/device/vi/vi_controller.c:165
#03  pc 000000000003e604  libmupen64plus-core.so: Routine call_interrupt_handler at ./mupen64plus-core/src/device/r4300/interrupt.c:497
#04  pc 00000000000b5190  libmupen64plus-core.so: Routine .E1 at linkage_arm.S:?

```

fzurita:
Here is a new one, and it's incredibly common against this revision: f5ea0d1e9f899c10f56e3f13f99c2a936fd2d4c8

********** Crash dump: **********
pid: 0, tid: 0 >>> org.mupen64plusae.v3.fzurita <<<
Stack frame #00  pc 0000000000028f94  ./libmupen64plus-core.so: Routine map_region at ./mupen64plus-core/src/device/memory/memory.c:202
[...]

PR #514, opened 2018-01-11
Fix potential crash with FBInfo and save states

loganmc10:
I believe this should fix https://github.com/mupen64plus/mupen64plus-core/issues/489

Doing ```memset(&g_dev.dp.fb, 0, sizeof(g_dev.dp.fb));``` will wipe out ```fb->mem, fb->rdram, fb->r4300```, which is probably what was causing the crash.

Using poweron_fb will only wipe the necessary data

This regression was introduced in https://github.com/mupen64plus/mupen64plus-core/commit/99f00d78a2feec139a60ca7d59fb163daf8029b6

@fzurita perhaps you could test and @bsmiles32 you could review as well?

### DIFF
--- a/src/device/rcp/rdp/fb.c
+++ b/src/device/rcp/rdp/fb.c
@@ -110,7 +110,6 @@ void init_fb(struct fb* fb,
 
 void poweron_fb(struct fb* fb)
 {
-
     memset(fb->dirty_page, 0, FB_DIRTY_PAGES_COUNT*sizeof(fb->dirty_page[0]));
     memset(fb->infos, 0, FB_INFOS_COUNT*sizeof(fb->infos[0]));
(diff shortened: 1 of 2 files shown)


fzurita:
Sure, I'll give it a shot.

bsmiles32:
Good catch ! Savestate was indeed overwriting components pointers. Lgtm

fzurita:
I can confirm, this fixes the crashing reported in #489.

fzurita:
After this change, the crash rate reported in the play store is massively reduced. It's even less than before https://github.com/mupen64plus/mupen64plus-core/commit/99f00d78a2feec139a60ca7d59fb163daf8029b6

There are still a few lingering crashes. I'll create issue for them soon.

PR #523, opened 2018-01-26
Add synchronization event between RSP and CPU

Gillou68310:
The idea of this PR is to use the unused upper 16bits of SP_STATUS_REG in order to generate sync event between RSP and CPU.

Here's an example how this could be used:
https://github.com/Gillou68310/GLideN64/commit/a9cc1492f631af1f1d721d483a43b4b3bb1f01fd

The main advantage of this method is to avoid resuming the locked rsp task while the event has not occurred.


loganmc10:
Looks good to me, tested on the games that use this (WDC, Stunt Racer, Gauntlet Legends) and saw no issue (although visually inspecting the code I wouldn't expect there to be any difference anyway).

Out of curiosity, why rewrite the do_SP_Task() function so much? Why not just add:

```
// Fallback for undefined rsp<->cpu sync event       
if ((sp->regs[SP_STATUS_REG] >> 16) == 0)
    sp->mi->regs[MI_INTR_REG] |= MI_INTR_SP;
```

To the existing ```if ((sp->regs[SP_STATUS_REG] & (SP_STATUS_HALT | SP_STATUS_BROKE)) == 0)``` statement? It seems like it wouldn't make any difference

Gillou68310:
I moved the ```if ((sp->regs[SP_STATUS_REG] & (SP_STATUS_HALT | SP_STATUS_BROKE)) == 0)``` condition inside the ```if (sp->mem[0xfc0/4] == 1)``` block because it should only happen in video ucodes, also ```new_frame``` should only be called when the rsp task is not locked.
But I agree with you that it was not necessary to refactor the ```if (sp->mi->regs[MI_INTR_REG] & MI_INTR_SP)``` block. It's just a personal preference for code readability. If people don't like it I can revert this part.

loganmc10:
> I moved the if ((sp->regs[SP_STATUS_REG] & (SP_STATUS_HALT | SP_STATUS_BROKE)) == 0) condition inside the if (sp->mem[0xfc0/4] == 1) block because it should only happen in video ucodes

We only know of games that do that in the video ucodes (and it might be the case), but wouldn't it be possible for a game to do that in an audio/other ucode?

> It's just a personal preference for code readability

I guess it doesn't really matter, it just seems to duplicate code to me, you do ```cp0_update_count(sp->mi->r4300);``` and ```sp->mi->regs[MI_INTR_REG] &= ~MI_INTR_SP;``` in every if block, you might as well just put ```cp0_update_count(sp->mi->r4300);``` before the if block and ```sp->mi->regs[MI_INTR_REG] &= ~MI_INTR_SP;``` after

loganmc10:
It seems like you were able to implement the ucode without this change, right? What does it do exactly, does it improve performance or something?

Gillou68310:
It's just an accuracy improvement to avoid calling do_SP_Task() at an improper time, at least for WDC and stunt racer. It might improve performance a bit I honestly didn't checked.

fzurita:
This should really be merged now that GLideN64 has merged the referenced commit.

loganmc10:
I believe this is an "extra" piece that wasn't part of the stuff that was just merged into GLideN64

fzurita:
Yeah, good point. Although, according to @Gillou68310 it was supposed to fix the lockups when resuming from a save state in the Boss games. That's the main reason.

loganmc10:
I think that this commit: https://github.com/Gillou68310/mupen64plus-core/commit/d9129e1 was supposed to do that, I'm going to work on that next (getting it mergable/removing the g_dev usage)

fzurita:
Ok, I must have misunderstood what he said.

fzurita:
I tested guantlet legends to see if it would impact performance there. I didn't really see any difference though.

fzurita:
Any reason why this wasn't merged?

richard42:
it has a conflict with the master now

loganmc10:
The other thing is I don't think it actually modified any behavior in any of the games (no known improvements from this change)

PR #529, opened 2018-02-05
Fix compile on MinGW

loganmc10:
For some reason, recent builds on MinGW are failing, the asm_defines_nasm.h file looks like this:

```
)define offsetof_struct_cached_interp_invalid_code (0x00000000
)define offsetof_struct_r4300_core_cached_interp (0x00000af8
)define offsetof_struct_tlb_LUT_w (0x00400680
)define offsetof_struct_tlb_LUT_r (0x00000680
)define offsetof_struct_tlb_entries (0x00000000
)define offsetof_struct_cp0_tlb (0x000002ec
)define offsetof_struct_cp0_count_per_op (0x000002e8
)define offsetof_struct_cp0_last_addr (0x000002e4
)define offsetof_struct_cp0_next_interrupt (0x00000218
)define offsetof_struct_cp0_regs (0x00000000
)define offsetof_struct_r4300_core_cp0 (0x00900c60
)define offsetof_struct_r4300_core_return_address (0x00000a20
)define offsetof_struct_r4300_core_save_rip (0x00000a18
)define offsetof_struct_r4300_core_save_rsp (0x00000a10
)define offsetof_struct_r4300_core_stop (0x00000134
)define offsetof_struct_r4300_core_lo (0x00000108
)define offsetof_struct_r4300_core_hi (0x00000100
)define offsetof_struct_r4300_core_regs (0x00000000
)define offsetof_struct_device_r4300 (0x00000000
```

The output from ```strings``` seems to have Windows newlines, and then gawk can't properly parse the file. Piping it to dos2unix on the way fixes the issue.

dos2unix should be available on all Linux distributions, so I think this would be a pretty safe change. If someone know of a way to fix the gawk script to deal with Windows newlines that would probably work as well.

### DIFF
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -353,6 +353,7 @@ CC        = $(CROSS_COMPILE)gcc
 CXX       = $(CROSS_COMPILE)g++
 STRINGS   = $(CROSS_COMPILE)strings
 AS        = nasm
+TR       ?= tr
 RM       ?= rm -f
 INSTALL  ?= install
 MKDIR    ?= mkdir -p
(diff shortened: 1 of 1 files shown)


loganmc10:
@bsmiles32 @richard42

richard42:
Yeah I saw this but I kind of don't like it because I think dos2unix is not a standard system tool on many Linux distros; it's something that you have to install. I always want to keep dependencies to a minimum. Another possible solution is to use 'tr', which is part of coreutils and installed everywhere. The syntax would look like this:

`tr -d '\r' < input.file  > new.file`

loganmc10:
Yes using tr works as well, I've updated the PR

richard42:
ok thanks.

PR #533, opened 2018-02-12
Remove CountPerOp=1 for Vigilante games

loganmc10:
The Vigilante games are very demanding on the CPU. Having CountPerOp=1 makes this even more pronounced.

@AmbientMalice you seem to know a lot of the history behind these settings. Do you know why the Vigilante games have CountPerOp=1? I assume we just copied it over from PJ64

I tested the games briefly and they seem to work the same using CountPerOp=2, but much less CPU usage.


AmbientMalice:
Nothing comes to mind, sorry.

PR #534, opened 2018-02-12
Don't deal with save states when RSP task is locked

loganmc10:
This is based off of @Gillou68310's work here: https://github.com/Gillou68310/mupen64plus-core/commit/d9129e1 to make sure that save states don't lock up in WDC/Stunt Racer.

@fzurita perhaps you could test this on the device that had the issue?

@bsmiles32 are you happy with how the pointer is passed around?

### DIFF
--- a/src/device/device.c
+++ b/src/device/device.c
@@ -167,7 +167,7 @@ void init_device(struct device* dev,
 
     init_rdram(&dev->rdram, mem_base_u32(base, MM_RDRAM_DRAM), dram_size, &dev->r4300);
 
-    init_r4300(&dev->r4300, &dev->mem, &dev->mi, &dev->rdram, interrupt_handlers,
+    init_r4300(&dev->r4300, &dev->mem, &dev->mi, &dev->sp, &dev->rdram, interrupt_handlers,
             emumode, count_per_op, no_compiled_jump, randomize_interrupt);
     init_rdp(&dev->dp, &dev->sp, &dev->mi, &dev->mem, &dev->rdram, &dev->r4300);
     init_rsp(&dev->sp, mem_base_u32(base, MM_RSP_MEM), &dev->mi, &dev->dp, &dev->ri);
--- a/src/device/r4300/interrupt.c
+++ b/src/device/r4300/interrupt.c
@@ -510,7 +510,7 @@ void gen_interrupt(struct r4300_core* r4300)
         dyna_stop(r4300);
     }
 
-    if (!r4300->cp0.interrupt_unsafe_state)
+    if (!r4300->cp0.interrupt_unsafe_state && !r4300->sp->rsp_task_locked)
     {
         if (savestates_get_job() == savestates_job_load)
         {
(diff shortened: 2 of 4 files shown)


fzurita:
It doesn't seem to work. It still locks up when resuming a save state in the ARM new dynarec.

fzurita:
Here is something interesting though. It only locks up if the save state is loaded at startup. If the save state is loaded after the game starts up, it doesn't freeze. Here is a save state that freezes when loaded at startup for world driver championship (U):
[2018-02-12-23-51-05.v2.sav.zip](https://github.com/mupen64plus/mupen64plus-core/files/1718690/2018-02-12-23-51-05.v2.sav.zip)

loganmc10:
Does this PR make any difference? I don't think I ever experienced freezing before, I'll have to test again tomorrow. I never tested loading at startup

fzurita:
It didn't seem to make any difference. It only took a couple of tries to make it freeze.

loganmc10:
Does it only happen with the dynarec?

fzurita:
I was able to easily reproduce using interpreter as well.

Gillou68310:
> It didn't seem to make any difference. It only took a couple of tries to make it freeze.

I can't reproduce the issue on windows with x86 new dynarec.

> Here is something interesting though. It only locks up if the save state is loaded at startup

This should fix the issue when loading a savestate at startup https://github.com/Gillou68310/GLideN64/commit/7f20b53972ed0e64b00e71723b484e59a9577596

@gonetz The problem is that all CRC in the specialMicrocodes need to be updated!

loganmc10:
@Gillou68310 in your opinion is this PR still needed? It seems to help. In the current master (x64) I can get it to lock up when saving/loading save states (maybe once every 5-8th time). With this PR, it seems to be stable, doesn't lock up.

Or would fixing the CRC's in GLideN64 fix the issue as well (both at startup and in-game)?

Gillou68310:
The CRC fix will only fix the load at startup issue. So yes this PR is still needed to fix the in-game lock up.

fzurita:
@Gillou68310 Can you make a pull request for https://github.com/Gillou68310/GLideN64/commit/7f20b53972ed0e64b00e71723b484e59a9577596?

gonetz:
> The problem is that all CRC in the specialMicrocodes need to be updated!

Could you explain, why?

Gillou68310:
The CRC will only have the correct value if ```uc_size = 4096``` which is unlikely, except if ```uc_size = 0```.

Gillou68310:
> Will it work with zilmar-spec emulators?

Yes it will

> Is uc_size constant for given ucode?

Yes we are already using it to compute ucode's checksum in rsp-hle (see ```normal_task_dispatching``` in hle.c)

bsmiles32:
> @bsmiles32 are you happy with how the pointer is passed around?

@loganmc10 : I'm okay with how you passed the pointers around.
An alternative, would have been to make  interrupt_unsafe_state a bitfield such that any component (eg r4300 and now rsp) can signal collectively wether it is safe or not to savestate.
The test to allow savestating, is still interrupt_unsafe_state == 0, eg all components signals that they are in a state where it is safe to do a savestate.
This way there is no need to pass any pointer around because the rsp can directly prevent savestates by setting the proper bit of rsp->r4300->interrupt_unsafe_state.

@Gillou68310 Congratulations for your work on the Boss ZSort ucode ! This is really a great achievement.

loganmc10:
> An alternative, would have been to make interrupt_unsafe_state a bitfield

Good point, I'll modify it to work that way.

loganmc10:
Thanks

Issue #528, opened 2018-02-03
Log spam when loading save states: Unexpected command format 01 c4 01  


fzurita:
Upon loading a save state with super mario 64, I'm getting massive core log spam with this text:
```
Unexpected command format 01 c4 01 
```

Here is the save state producing the issue. I have no idea what caused it.

[mario64_log_spam.zip](https://github.com/mupen64plus/mupen64plus-core/files/1692313/mario64_log_spam.zip)

loganmc10:
I tested my own save state and the one you provided. I couldn't reproduce the issue, I didn't get any of those messages. Does it only happen with the new dynarec?

fzurita:
Yep, you are right, it's only happening with the new dynarec.

fzurita:
I think I have isolated this to having 4 controllers present.

fzurita:
I think we need need a new label specifically for the new dynarec, there is a large collection of new dynarec issues.

bsmiles32:
By just reading the issue description, this indeed sounds like the savestate being made with a different number of plugged controllers than when loaded. Maybe combined with a savestate made before the PIF refactoring (in which case there we can only do a "best effort" to reload PIF state). This is unfortunate, but as a workaround, just hex editing the saved pif ram could do the trick.

fzurita:
Gotcha, I can tell you for sure that the save state was created after the PIF refactoring.

I can also tell you that it's true, the save state was created using one controller present, but when loaded, the number of controllers was set to 4.

bsmiles32:
Quick question, can you test https://github.com/mupen64plus/mupen64plus-core/pull/535 against this issue ?

fzurita:
Yep, will do

fzurita:
Yep, your changes in #535 fixed this issue.

bsmiles32:
Good !

Issue #531, opened 2018-02-10
Unplugging controllers using CONTROL_INFO doesn't work


fzurita:
Changing pak type using the CONTROL_INFO API seems to work fine, but changing the "Present" of a controller doesn't seem to have any effect at run time. It stays to whatever value was set at startup.

One good way to test this is by using Super Smash Brothers in the VS mode during the character select screen. Unplugging a controller there would make the hand for that controller go away. If one makes a save state, then exits, then changes the number of controller plugged in and loads a save state, then it works correctly.

fzurita:
This would be useful for example if the input plugin detects that a USB controller is unplugged, then the virtual controller would be unplugged. If it's plugged back in, then the virtual controller would be plugged in again.

fzurita:
Fixed after #535

PR #535, opened 2018-02-14
Allow game controllers to be unplugged and plugged in on the fly

fzurita:
This can be done using the controller plugin interface using the "Present" flag for each controller. I have to do more testing on it to make sure everything works correctly.

@bsmiles32 Can you review this and suggest improvements if necessary?

The main changes I did were to always initialize all controllers and use the controller API in pif.c to determine if a controller is present instead of checking for a null pointer for the controller interface.

I did a quick test using the smash brother player select screen and the hands were disappearing and appearing when the controller Present flag was toggled.

This is my attempt at fixing: https://github.com/mupen64plus/mupen64plus-core/issues/531

### DIFF
--- a/src/backends/api/controller_input_backend.h
+++ b/src/backends/api/controller_input_backend.h
@@ -54,6 +54,10 @@ enum mouse_controller_input {
 
 struct controller_input_backend_interface
 {
+    /* Returns emulated controller plugged in status (32-bit)
+     */
+    int (*get_pluggedin)(void* cin);
+
     /* Returns emulated controller input status (32-bit)
      * Encoding of the input status depends on the emulated controller flavor.
      */
--- a/src/backends/api/controller_input_backend.h
+++ b/src/backends/api/controller_input_backend.h
@@ -22,6 +22,8 @@
 #ifndef M64P_BACKENDS_API_CONTROLLER_INPUT_BACKEND_H
 #define M64P_BACKENDS_API_CONTROLLER_INPUT_BACKEND_H
 
+#include "api/m64p_types.h"
+
 #include <stdint.h>
 
 enum standard_controller_input {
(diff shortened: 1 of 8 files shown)


bsmiles32:
Just quick notes for now (I'll try to submit some code later tonight) :
- I'd prefer if we regroup the controller presence and button state query to a single method eg

```
enum controller_input_error_code
{ NO_ERROR, NOT_PLUGGED }
enum controller_input_error_code controller_input_backend_interface.get_input(uint32_t* input)
```

This is simpler to implement for plugins and avoid incoherent state (controller present at presence check, but not at button polling).
- The joybus api shouldn't be changed. The controller presence check should be (re)done in the process_controller_command method.

fzurita:
Yes, they was my initial approach, but I quickly discovered that process_controller_command did not seem to be intended to be called when a controller is not present. I guess, I could change that so it's always called.

bsmiles32:
@fzurita Ok, pushed some modifications to make it work with the approach I proposed. Combined with https://github.com/mupen64plus/mupen64plus-input-sdl/pull/66, I am able to disconnect / reconnect a controller and see the select hand disappear/reappear. However a few gotcha's when reconnecting the controller (which looks related to the input plugin, not the core):
- there is some garbage data coming from the input driver just after reconnection.
- when testing with multiple controllers (eg. 3 controllers plugged, then unplug 1, then another, then replug the former, then the latter....) the input plugin gets confused about which controller maps which N64 controller

fzurita:
Is that hot plug-in limitation of the sdl input plugin itself? The Android input plugin doesn't have such problem. I was able to hot plug without having the controllers connected at startup.

bsmiles32:
I think it is a limitation of current sdl input plugin. Not sure if it is SDL itself or the input plugin with it's SDL1 heritage.

fzurita:
@bsmiles32 I tested your changes against Android and everything works as well as my implementation. I'm able to start with controllers unplugged and then plugged them in on the fly, and the game will see them. Same goes with unplugging. Not to mention, it also fixes https://github.com/mupen64plus/mupen64plus-core/issues/528

fzurita:
One thing I noticed though, in my implementation I made it so that ```channel->ijbd``` could never be null and I don't think you reverted that, but you put back the checks for ```channel->ijbd == NULL```

bsmiles32:
You're right, ```channel->ijbd``` can't be NULL anymore because of how it is currently set up in main.c. However, I'll keep these tests for now in pif.c until I decide how I'll evolve the "device setup" part in main.c. Indeed, in some future work I'd like to add more flexibility regarding backends selection, and this work might make use of ijbd being NULL.

fzurita:
Were there any N64 games that didn't support "hot-pluggin"?

fzurita:
@richard42 can this be merged?

PR #537, opened 2018-02-15
new_dynarec: Fix DMULT/DMULTU recompiled code

Gillou68310:
Interpreted version could be used instead by defining INTERPRETED_MULT64


Gillou68310:
@fzurita, @loganmc10 could you guys test this for regressions?

loganmc10:
I tested x86 (Windows). Looks ok to me, no issues in WDC

fzurita:
I can verify that this fixed the arm new dynarec with boss games. I was hoping this would fix MRC, but it didn't. It sounds like a similar issue to this one. I could try your approach of replacing dynarec instructions with interpreter ones to see which one fixes it.

PR #538, opened 2018-02-15
Alternative implementation of "Don't deal with save states when RSP t…

bsmiles32:
…ask is locked"

See https://github.com/mupen64plus/mupen64plus-core/pull/534

### DIFF
--- a/src/device/r4300/cp0.h
+++ b/src/device/r4300/cp0.h
@@ -168,6 +168,11 @@ struct interrupt_handler
 
 enum { CP0_INTERRUPT_HANDLERS_COUNT = 12 };
 
+enum {
+    INTR_UNSAFE_R4300 = 0x01,
+    INTR_UNSAFE_RSP = 0x02,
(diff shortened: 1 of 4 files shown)


loganmc10:
The only real difference I see is that rsp_task_locked is stored in the save state right? So when we load a save state you should probably set the flag there as well

bsmiles32:
rsp_task_locked is indeed stored inside the savestate, but it may not be needed anymore now that we prevent savestating when rsp_task_locked is set... so updating interrupt_unsafe_state inside the save state loading doesn't much sense as it must be 0 to begin with.
Maybe I should also remove the rsp_task_locked variable from the savestate ?

loganmc10:
Oh right it must always be 0 with this change. I would just leave it, who knows if it may be used in the future again

bsmiles32:
Squashed a small copy paste error which completely broke old savestates loading. Should be fixed now.

PR #541, opened 2018-02-23
Force alignment of PI DMA

loganmc10:
This needs more testing, I'll report back once I've tested it more.

Fixes: https://github.com/project64/project64/issues/633

Taz Express tries to do an unaligned PI DMA transfer. Cart addresses must be 2-byte aligned and DRAM address must be 8-byte aligned.

This may/probably will break some ROM hacks (particularly some popular SM64 hacks). They don't work on a real console, and this is one of the main reasons.

### DIFF
--- a/src/device/rcp/pi/pi_controller.c
+++ b/src/device/rcp/pi/pi_controller.c
@@ -52,8 +52,8 @@ enum
 
 static void dma_pi_read(struct pi_controller* pi)
 {
-    uint32_t cart_addr = pi->regs[PI_CART_ADDR_REG];
-    uint32_t dram_addr = pi->regs[PI_DRAM_ADDR_REG];
+    uint32_t cart_addr = pi->regs[PI_CART_ADDR_REG] & ~UINT32_C(1);
+    uint32_t dram_addr = pi->regs[PI_DRAM_ADDR_REG] & ~UINT32_C(7);
     uint32_t length = (pi->regs[PI_RD_LEN_REG] & UINT32_C(0x00fffffe)) + 2;
     const uint8_t* dram = (uint8_t*)pi->ri->rdram->dram;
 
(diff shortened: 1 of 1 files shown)


fzurita:
Can we not break ROM hacks? Maybe at the same time as introducing this, add a core configuration parameters to allow overriding this?

loganmc10:
Well I guess that's sort of a philosophical disagreement. Those games don't run on a N64, they don't work with GLideN64 or Angrylion. We already ran into another issue with the VSYNC_REG

In my view they aren't really N64 games. The "system" they were written to run on is Project64 1.X

I'm not talking about all ROM hacks of course, just the ones that don't work on the N64.

But this is something that should be decided by @richard42 , it's sort of a project direction decision

fzurita:
Well, I completely agree with you. ROM hacks that don't work with the real hardware are a thing that should had never existed. Unfortunately, there are a lot of people that still play them.

I agree, from a preservation stand point, if the ROM hack doesn't work with the real hardware, it shouldn't work in the emulator. At the same time, if we want this emulator to be more "popular" we should maintain some sort of compatibility.

I think, clearly documented configuration parameters stating that they are not accurate to real hardware is a good compromise on my opinion.

loganmc10:
@Narann I'm not sure why the AppVeyor build failed, can you try restarting it?

loganmc10:
> I think, clearly documented configuration parameters stating that they are not accurate to real hardware is a good compromise in my opinion

I guess I disagree, we'd already need something in the VI and PI controllers, and potentially other places in the future. I feel like it'd just be mucking up the code to support these ROMs. It's the ROMs that are broken, one of the many ROM ninjas out there should go to work in fixing the ROMs.

These broken ROM hacks are still being made to this day, I feel like purposely adding support in the emulator is just encouraging it

loganmc10:
I haven't come across any issues in the many retail games I've tested. I tested "Super Mario Star Road" (ROM Hack), and this change does indeed break it.

So we need some kind of direction from somebody with merge rights @richard42 @bsmiles32 . Do we add a config option like "Allow unaligned DMAs", or do we add INI entries for the ROM hacks to allow unaligned DMAs, or do we break compatibility with the ROM hacks?

richard42:
This is a worthwhile patch as it increases accuracy and fixes Taz Express. I think it's not worth the extra complexity to add configuration options or INI entries to support "homebrew" ROMs which don't work on the real hardware.

PR #545, opened 2018-03-09
Fix: try to keep savestates offsets aligned.

bsmiles32:
Lame attempt at keeping savestate offsets aligned... not really proud of what I've done because it is still very fragile, but should fix current crashes.

Should also probably bump savestate version ? But is it considered a major version bump or minor ?

@Gillou68310 @fzurita can you test if this indeed fixes your crashes ? Thanks

### DIFF
--- a/src/main/savestates.c
+++ b/src/main/savestates.c
@@ -61,7 +61,7 @@ enum { GB_CART_FINGERPRINT_SIZE = 0x1c };
 enum { GB_CART_FINGERPRINT_OFFSET = 0x134 };
 
 static const char* savestate_magic = "M64+SAVE";
-static const int savestate_latest_version = 0x00010200;  /* 1.2 */
+static const int savestate_latest_version = 0x00010300;  /* 1.3 */
 static const unsigned char pj64_magic[4] = { 0xC8, 0xA6, 0xD8, 0x23 };
 
 static savestates_job job = savestates_job_nothing;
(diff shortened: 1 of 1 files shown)


fzurita:
It does fix the crashing when compiling with LTO enabled and loading save states.

Loading save states produced before this change is broken though, probably because you changed the structure of the save state file.

fzurita:
Since you changed the structure of the save state, you do need to change the version I guess so that we don't load broken save states.

Although, I think we should still make a way to load the old save states. Maybe have a converter that is able to convert the old save state format to the new one in memory?

loganmc10:
> It looks like we just need to shift everything by two bytes where you inserted those two extra bytes.

I believe the shifting of the other variables is important as well to keep everything 8-byte aligned, it's not the extra padding, that's what will make upping the save state version or loading old versions a little trickier

fzurita:
Yep, you are right. Stuff got moved around. I didn't pay enough attention. That does make it tricky to load the old save states.

fzurita:
To me it seems simple to keep compatibility. Increment save state version, and don't load the "additional data" if we have the old save state version. Treat the previous save state version like they didn't have the additional data at all.

bsmiles32:
Updated PR with better implementation :
- v1.2 is finally kept unchanged, we just pre-load data to an aligned buffer before doing the real access
- v1.3 changes data offsets to have them aligned like I did in previous commit.

@fzurita @loganmc10 What do you think ? :)

fzurita:
I think this is better. Old save states still work and no crashing.

fzurita:
Can this be merged?

PR #549, opened 2018-03-21
Clear CLOCK_REG when requested

loganmc10:
I have seen some interest lately in emulating the DPC_CLOCK_REG register. See for example https://github.com/cxd4/rsp/issues/14

This should be a harmless change, we currently don't clear the CLOCK_REG when the game writes DPC_CLR_CLOCK_CTR to DPC_STATUS_REG. Given that some RSP/GFX plugins might start incrementing this register, we should implement clearing it. I know of a few games that write DPC_CLR_CLOCK_CTR (Gauntlet Legends and Taz Express come to mind).

Also, DPC_STATUS_START_GCLK should be set on boot, I'm not aware of any game where this makes a difference though.

Renamed some of the status flags to match what is in rcp.h

### DIFF
--- a/src/device/rcp/rdp/rdp_core.c
+++ b/src/device/rcp/rdp/rdp_core.c
@@ -31,11 +31,11 @@
 static void update_dpc_status(struct rdp_core* dp, uint32_t w)
 {
     /* clear / set xbus_dmem_dma */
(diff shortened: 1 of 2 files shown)


loganmc10:
Oh also I added DPC_STATUS_START_VALID and DPC_STATUS_END_VALID. Again, I'm not aware of any game that uses these bits, but they are supposed to be set to indicate that a valid entry was written to START_REG/END_REG

bsmiles32:
Tested a few games and didn't noticed any regression. Looks good to me.

PR #550, opened 2018-03-23
Don't call SDL_PumpEvents() in core

loganmc10:
SDL_PumpEvents is described as:

```
SDL_PumpEvents() gathers all the pending input information from devices and places it in the event queue
```

It is related to the input functions. In the past, mupen64plus-ae has used a slightly modified version of SDL2, but now I am trying to update it to use a vanilla version of 2.0.8, however it crashes when calling SDL_PumpEvents(). Presumably because there is no keyboard and SDL_INIT_JOYSTICK is never called. The input subsystem is not setup.

I moved SDL_PumpEvents to the beginning of the ```GetKeys()``` function in input-sdl. I tested using a keyboard and a joystick, both in-game inputs and events (like pause/quit game) still work with keyboard or joystick. See https://github.com/mupen64plus/mupen64plus-input-sdl/pull/68

### DIFF
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -213,7 +213,6 @@ static void main_check_inputs(void)
 #ifdef WITH_LIRC
     lircCheckInput();
 #endif
-    SDL_PumpEvents();
 }
 
 /*********************************************************************************************************


bsmiles32:
Not an expert on this matter so don't hesitate to point if I'm wrong, but I'm not sure it is the best move to completely remove this from the core as the core itself relies on it for checking various button-presses for triggering pause,savestates,mute,.... There are also non-SDL input plugins (raphnet comes to mind) that would probably break the core with this scheme.

Maybe we could start thinking about moving input plugins inside the core so we can have a tighter integration/control over what's done.

loganmc10:
> I'm not sure it is the best move to completely remove this from the core as the core itself relies on it for checking various button-presses for triggering pause,savestates,mute,.... There are also non-SDL input plugins (raphnet comes to mind) that would probably break the core with this scheme.

I tested this using Input-SDL and all the hotkeys like fast-forward, pause, save state, etc... still work. You make a good point about non-SDL input plugins like raphnet though. I just tested this with the raphnet plugin and all the hotkeys still work. I'll will admit I'm not 100% sure why they still work... but they do, so it doesn't seem like this change will break any of the existing input plugins

fzurita:
Could an alternative be to move this call into the input plugin and then it can be called with an API call? It would mean expanding the input plugin API.

loganmc10:
It's already called in the input plugin, this call is redundant as far as I could tell

loganmc10:
@fzurita if I recall correctly this change should allow you to update the version of sdl2 used in AE to the latest upstream version without any real modifications

fzurita:
Yep, I remember, I'll get to that.

Zapeth:
This change causes the emulator to become unresponsive for certain roms. More specifically those who don't process any controller input, like most of [these test roms](https://github.com/PeterLemon/N64) (`process_controller_command()` is never called in those cases)

PR #553, opened 2018-03-29
Fix write_memory_8

rogual:
This was overflowing a u8, causing zeroes to be written most of
the time instead of the intended value.


bsmiles32:
Thanks for your contribution !

PR #556, opened 2018-04-02
Add VidExt_GL_GetDefaultFramebuffer

loganmc10:
On some platforms (for instance, iOS) the default framebuffer object depends on the surface being rendered to, and might be different from 0.

This is also true for Qt. If the rendering surface is a QOpenGLWidget, the application must call ```GLuint QOpenGLContext::defaultFramebufferObject()``` to get the name of the default FBO. See (http://doc.qt.io/qt-5/qopenglcontext.html#defaultFramebufferObject). If it is a QOpenGLWindow, then the default FBO is "0".

GTK is the same, it doesn't allow rendering into the "0" framebuffer, the application must render into a framebuffer created by GTK. RetroArch does the same thing.

This was recently identified as an issue for the Provenance project (https://github.com/Provenance-Emu/Provenance) which is an iOS app that includes mupen64plus.

@braindx @JoeMatt can you please review this and comment whether this will meet the requirements for your project?

### DIFF
--- a/doc/emuwiki-api-doc/Mupen64Plus-v2.0-API-Versioning.mediawiki
+++ b/doc/emuwiki-api-doc/Mupen64Plus-v2.0-API-Versioning.mediawiki
@@ -112,5 +112,7 @@ This is the most complicated interface, because it involves 3 components: the vi
 * '''VIDEXT_API_VERSION''' version 3.0.0:
 ** add VidExt_ResizeWindow() function in video extension.  This function is called by the video plugin to notify the window manager (SDL if no video extension is registered by the front-end) that the OpenGL render window size should change.
 ** add m64p_video_flags parameter to the VidExt_SetVideoMode() function.  Currently the flags are only used to notify the window manager that resizing is supported by the video plugin, and it should create a resizable window if possible.  This may be extended in the future to support other features.
+* '''VIDEXT_API_VERSION''' version 3.1.0:
+** add VidExt_GL_GetDefaultFramebuffer() function in video extension. This function should be called to get the name of the default FBO.
 * '''INPUT_API_VERSION''' version 2.0.1:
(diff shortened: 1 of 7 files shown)


braindx:
Yes, this would work for us. Alternatively, video plugins could just call glGet* using GL_FRAMEBUFFER_BINDING to cache the default framebuffer before the first time they change it. I'm unsure as to whether any platforms have a need to change this during runtime.

JoeMatt:
I'll need to update our project to pull mupen64plus-core as a submodule since we're pretty out of date too. I'll work on that next so we can actually use / test this change.

loganmc10:
@braindx Yes, but that is still assuming that the video subsystem is binding the default framebuffer for you when it creates the GL context. That isn't necessarily the case. They could just leave the binding at "0" and expect you to call ```QOpenGLContext::defaultFramebufferObject()``` (for instance) to get the correct framebuffer to bind. So I think this is the safer approach.

loganmc10:
Although it would be pretty crazy for them to leave it at "0", I doubt Qt does this, I'm just saying it's not technically "illegal" for them to do, so we should probably deal with it in the API

loganmc10:
I'm not familiar with what mutex rendering is but perhaps you can open a different issue for that? In my experience getting API changes merged can take time so we should work through this issue if we can

loganmc10:
On that note though, if you already have some speed limiting mechanism, you can disable the built-in speed limit by calling ```CoreDoCommand(M64CMD_CORE_STATE_SET, M64CORE_SPEED_LIMITER, 0);``` before the ROM starts

loganmc10:
@richard42 @bsmiles32 any comment on this?

loganmc10:
Just a reminder to front-end developers like @fzurita , now that this has been merged, the front-end VidExt API portion needs to be updated to include this new function, or else the core will reject the VidExt implementation from the front-end

PR #557, opened 2018-04-03
Increase DebugMessage buffer size to prevent overflows

JoeMatt:
When running on iOS simulator, some paths are really long. This was causing mupen64plus core to crash DebugMessage with a buffer overflow when printing out statements related to paths.


loganmc10:
Doesn't vsnprintf exist? We should probably use that as well to ensure there aren't any buffer overflows

PR #561, opened 2018-04-18
Update xxHash to 0.6.5

loganmc10:
https://github.com/Cyan4973/xxHash/releases/tag/v0.6.5


fzurita:
Did you see any performance improvement from the update to improve performance on "small" hashes?

loganmc10:
I didn't actually do any testing for this update, I think they do more/better testing than I'd be capable of so I tend to just take their word for it.

I don't know if XXH_INLINE_ALL would add any benefit to us, I guess we'd have to test, you would remove xxhash.c from the source list in that case and just define XXH_INLINE_ALL if I understand correctly

PR #564, opened 2018-05-11
GB Camera support

bsmiles32:
m64p_gbcam

Following discussion with @LuigiBlood in PR #446, this PR adds GB camera support.
See commit message for more information.

NOTE: I haven't updated the msvc files yet, I'd like to get some feedback first.

@AntonioND: big thanks for sharing your RE work on the GB camera !

### DIFF
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -271,6 +271,19 @@ endif
 CFLAGS += $(LIBPNG_CFLAGS)
 LDLIBS += $(LIBPNG_LDLIBS)
 
+ifeq ($(OPENCV), 1)
+  # OpenCV lib
+  ifeq ($(origin OPENCV_CFLAGS) $(origin OPENCV_LDLIBS), undefined undefined)
+    ifeq ($(shell $(PKG_CONFIG) --modversion opencv 2>/dev/null),)
+      $(error No OpenCV development libraries found!)
+    endif
+    OPENCV_CFLAGS += $(shell $(PKG_CONFIG) --cflags opencv)
+    OPENCV_LDLIBS += $(shell $(PKG_CONFIG) --libs opencv)
+  endif
+  CFLAGS += $(OPENCV_CFLAGS)
+  LDLIBS += $(OPENCV_LDLIBS)
+endif
+
 # test for presence of SDL
 ifeq ($(origin SDL_CFLAGS) $(origin SDL_LDLIBS), undefined undefined)
   SDL_CONFIG = $(CROSS_COMPILE)sdl2-config
@@ -446,6 +459,7 @@ SOURCE = \
     $(SRCDIR)/backends/plugins_compat/input_plugin_compat.c \
     $(SRCDIR)/backends/clock_ctime_plus_delta.c \
     $(SRCDIR)/backends/file_storage.c \
+    $(SRCDIR)/backends/dummy_video_backend.c \
     $(SRCDIR)/device/cart/cart.c \
     $(SRCDIR)/device/cart/af_rtc.c \
     $(SRCDIR)/device/cart/cart_rom.c \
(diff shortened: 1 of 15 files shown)


LuigiBlood:
Oh wait you really went that far to support a camera as well. I didn't even do that.

bsmiles32:
Yep, full GB camera support (as far as documented by @AntonioND) with live capture.
I know the quality of the GB camera is not great, but I see that as a first step... I have greater ambitions for the video input backend :) If you have some info on the A/V capture cart we could look into that as well

LuigiBlood:
This is as far as I got about the Capture Cartridge:
https://github.com/LuigiBlood/64dd/wiki/Capture-Cartridge

I don't understand the video portion a single bit. It's very confusing and each game handles it slightly differently.

bsmiles32:
Updated travis-ci file.

bsmiles32:
@LuigiBlood Thanks for the link, will look into it

src/backends/api/video_backend.h
```diff
@@ -0,0 +1,44 @@
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+ *   Mupen64plus - video_backend.h                                         *
+ *   Mupen64Plus homepage: http://code.google.com/p/mupen64plus/           *
```
Jj0YzL5nvJ: Obsolete URL...

AntonioND:
Thanks for telling me, @bsmiles32! It's nice to see that other people find that documentation useful. :)

bsmiles32:
- Added vs2013 support (require https://github.com/mupen64plus/mupen64plus-win32-deps/pull/6)
- Support older OpenCV version from ubuntu used in travis


@Jj0YzL5nvJ Obsolete URL fixed. Thanks.

bsmiles32:
Few things to consider before considering merging this :
- I've put the opencv specific configuration parameter in the Core section, but since it's backend specific it may be better to put it in a opencv video backend specific section
- OpenCV is quite a big dependency that I don't want to force upon our users, maybe we should allow to make it optional at compile time (or completely disable camera support). Better, we should be able to enable available backends at compile-time, and for enabled backends be able to select which one to use at run-time. This not just for the video input backends, but for all backends when multiple implementation are available.
- Only 1 camera can be used for all emulated gb cameras (up to 4 because one per game controller).
- I went with OpenCV because it was easy to implement in a cross plateform way video capture, but if you happen to know a simpler cross-plateform video capture library I'm all ears :)

richard42:
I don't like having such a big dependency for something that most of our users will never use, so a compile-time make flag is probably necessary. You're using this for cross-platform capability of video capture with whatever cameras happen to be attached to the computer?

bsmiles32:
> I don't like having such a big dependency for something that most of our users will never use, so a compile-time make flag is probably necessary.

Agreed.

> You're using this for cross-platform capability of video capture with whatever cameras happen to be attached to the computer?

Yes. Note that the OpenCV capture API allow to use webcams but also video/image files transparently (not tested but advertised on their documentation). Internally it wraps different backends depending on the plateform it is run.
Beside the capture stuff I also use some functions to resize the captured image to the desired resolution (and to display images for debug purpose, but that's disabled by default). But I don't use any "image processing" OpenCV stuff.
The ideal solution would be to work with SDL upstream to have them integrate a cross-plateform video capture API, so we can benefit from that without all the OpenCV burden. But way too much work for a rarely used feature... Or find a much lighter library, or do an optimized windows build with only video capture and image resizing enabled so we don't bring the full opencv burden.

bsmiles32:
- Updated with compile-time make flag (OPENCV=1 with Makefile). Not done for VS2013 because editing by hand msvc files is really asking for troubles, so some other Window dev will have to port this compile-time {en,dis}ablement. [For now, OpenCV is required when using VisualStudio].
- Fixed a bug with savestates (forgot to rename {M64282FP -> POCKET_CAM}_REGS_COUNT value).

projects/unix/Makefile
```diff
@@ -646,7 +666,7 @@ targets:
 	@echo "    WARNFLAGS=flag == compiler warning levels (default: -Wall)"
 	@echo "    PIC=(1|0)      == Force enable/disable of position independent code"
 	@echo "    OSD=(1|0)      == Enable/disable build of OpenGL On-screen display"
-	@echo "    NEW_DYNAREC=1  == Replace dynamic recompiler with Ari64's experimental dynarec"
```
richard42: was this an accident? I think we should keep this line for NEW_DYNAREC in the makefile

bsmiles32:
@richard42 Yep, silly mistake on my part. Updated with Makefile line re-added.

bsmiles32:
Disabling opencv stuff in MSVC so that Appveyor has a chance to build it.

PR #566, opened 2018-05-16
Initial bio pak support.

bsmiles32:
For now, the bio pak only report hard-coded BPM value (but is impacted by fast-forwarding).

In future work remote heart rate monitoring methods (rPPG) could be
implemented to provide an experience similar to the Bio Sensor pak using
a regular webcam. However it could be quite CPU intensive and may prove
challenging in multi-player context.

Another idea to create an "equivalent" Bio Sensor experience without the
original device, could be to derive the reported heart rate from the
BPM (or any relevant quantity) from a user specified audio clip.

tetris64-000


bsmiles32:
Fixed outdated homepage link.

Issue #520, opened 2018-01-15
Still some crashes, probably when loading save states


fzurita:
The fix for https://github.com/mupen64plus/mupen64plus-core/issues/489 fixed a VERY large number of crashes that were being reported to me. Unfortunately, it was not a silver bullet, and there are still some crashes left. They probably happen when loading save states as well and seemed to point to bad function pointers still. Here are two of them:

```
********** Crash dump: **********
pid: 0, tid: 0 >>> org.mupen64plusae.v3.fzurita <<<
#00  pc 00000000000b4250  ./libmupen64plus-core.so: Routine masked_write at ./mupen64plus-core/src/device/memory/memory.h:67
#01  pc 0000000000047f44  ./libmupen64plus-core.so: Routine mem_write32 at ./mupen64plus-core/src/device/memory/memory.h:87
#02  pc 00000000000aaf78  ./libmupen64plus-core.so: Routine write_word_new at ./mupen64plus-core/src/device/r4300/new_dynarec/new_dynarec.c:11223
#03  pc 0000000000000eac  <unknown>
Crash dump is completed
```

```
********** Crash dump: **********
pid: 0, tid: 0 >>> org.mupen64plusae.v3.fzurita <<<
#00  pc 00000000000b3ffc  ./libmupen64plus-core.so: Routine map_corrupt_rdram at ./mupen64plus-core/src/device/rdram/rdram.c:118
#01  pc 0000000000047f44  ./libmupen64plus-core.so: Routine mem_write32 at ./mupen64plus-core/src/device/memory/memory.h:87
#02  pc 00000000000aad6c  ./libmupen64plus-core.so: Routine write_byte_new at ./mupen64plus-core/src/device/r4300/new_dynarec/new_dynarec.c:11193
#03  pc 000000000003bdf0  <anonymous>
```

If I remember correctly, these crashes still happened even before https://github.com/mupen64plus/mupen64plus-core/commit/99f00d78a2feec139a60ca7d59fb163daf8029b6 increased the crash rate by a large amount.

fzurita:
Here is another one:

```
********** Crash dump: **********
pid: 0, tid: 0 >>> org.mupen64plusae.v3.fzurita <<<
#00  pc 00000000000b4268 ./libmupen64plus-core.so: Routine rdram_dram_address at ./mupen64plus-core/src/device/rdram/rdram.h:67
[...]

PR #569, opened 2018-05-30
Cleanup tpak savestate loading.

bsmiles32:
This may help fixing some edge-cases of crashes while loading
savetstates.

We isolate the savestate data extraction part from the state committing
part so that curr offset don't depends on external factors which could
lead to crashes.

@fzurita Can you test that to see if it helps with https://github.com/mupen64plus/mupen64plus-core/issues/520 ?


fzurita:
I think #520 was happening before transfer pak support was added, so it will probably not help. I will include it though on my next pak to the play store to see if it improves the situation.

fzurita:
I didn't see any changes in the crash rate after this was included in the play store build of mupen64plus AE.

PR #579, opened 2018-08-03
Make --nosaveoptions not ever save options.

bugfood:
As a new mupen64plus user, I struggled with the configuration, due to two major categories of unexpected behavior:

1. Under some circumstances, the configuration was changed and saved even when I specified `--nosaveoptions`, potentially leading to undesired options being persisted in the configuration.
2. The program runs when there is no configuration, and it generates a configuration file, but command-line parameters are not always respected unless the configuration is present first.

I do not yet have a proposed solution for problem 2 above, but this should help for problem 1, at least when `--nosaveoptions` is specified. I would like to first get this patchset accepted, if possible. I will show an example of the problems below. Please let me know if I have not explained things clearly... it's kind of tricky.

I will also post a corresponding patchset for ui-console.

```
# Set up an alias for convenience; this will be used in examples:
$ alias mupen64plus='/usr/local/src/mupen64plus/install/bin/mupen64plus --plugindir /usr/local/src/mupen64plus/install/lib/mupen64plus/ --datadir /usr/local/src/mupen64plus/install/share/mupen64plus/ --corelib /usr/local/src/mupen64plus/install/lib/libmupen64plus.so.2 --testshots 0'
```

Example from master branch of core and ui-console:
```
# Clean out configuration:
$ rm ~/.config/mupen64plus/mupen64plus.cfg

# Do a test run with --nosaveoptions; expected behavior is for configuration to
# not be saved:
$ mupen64plus --nosaveoptions /tmp/test.rom 2>&1 | tail -n 1
Core Status: Rom closed.

# Test if configuration was saved:
$ test -f ~/.config/mupen64plus/mupen64plus.cfg ; echo $?
0

# The configuration was saved, contrary to expectation.
# Check joystick order in preparation for next test:
$ grep 'name = ' ~/.config/mupen64plus/mupen64plus.cfg
name = "Twin USB Joystick"
name = "Twin USB Joystick"
name = "DragonRise Inc.   Generic   USB  Joystick  "
name = ""

# Try a different joystick order:
$ mupen64plus --nosaveoptions --set 'Input-SDL-Control1[name]=DragonRise Inc.   Generic   USB  Joystick  ' --set 'Input-SDL-Control1[mode]=1' ~/files/roms/N64/Zelda64.rom 2>&1 | tail -n 1
Core Status: Rom closed.

# Check joystick order again; the expected behavior is that the order is
# unchanged, since --nosaveoptions was present:
$ grep 'name = ' ~/.config/mupen64plus/mupen64plus.cfg
name = "DragonRise Inc.   Generic   USB  Joystick  "
name = "Twin USB Joystick"
name = "Twin USB Joystick"
name = ""

# The joystick order was changed, contrary to expectation.
```

Now the same example using patched UI and core:
```
# (using the same alias as before)

# Clean out configuration:
$ rm ~/.config/mupen64plus/mupen64plus.cfg

# Do a test run with --nosaveoptions; expected behavior is for configuration to
# not be saved:
$ mupen64plus --nosaveoptions /tmp/test.rom 2>&1 | tail -n 1
Core Status: Rom closed.

# Test if configuration was saved:
$ test -f ~/.config/mupen64plus/mupen64plus.cfg ; echo $?
0

# Configuration was not saved, as expected. Now allow the configuration to be
# saved in preparation for the next test.
$ mupen64plus /tmp/test.rom 2>&1 | tail -n 1
Core Status: Rom closed.

# Check the joystick order before running the test:
$ grep 'name = ' ~/.config/mupen64plus/mupen64plus.cfg
name = "Twin USB Joystick"
name = "Twin USB Joystick"
name = "DragonRise Inc.   Generic   USB  Joystick  "
name = ""

# Try a different joystick order, with --nosaveoptions present:
$ mupen64plus --nosaveoptions --set 'Input-SDL-Control1[name]=DragonRise Inc.   Generic   USB  Joystick  ' --set 'Input-SDL-Control1[mode]=1' ~/files/roms/N64/Zelda64.rom 2>&1 | tail -n 1
Core Status: Rom closed.


# Check joystick order again; the expected behavior is that the order is
# unchanged, since --nosaveoptions was present:
$ grep 'name = ' ~/.config/mupen64plus/mupen64plus.cfg
name = "Twin USB Joystick"
name = "Twin USB Joystick"
name = "DragonRise Inc.   Generic   USB  Joystick  "
name = ""

# The order is indeed unchanged, as expected.
```

Thanks,
Corey

### DIFF
--- a/src/main/eventloop.c
+++ b/src/main/eventloop.c
@@ -509,7 +509,6 @@ void event_initialize(void)
 int event_set_core_defaults(void)
 {
     float fConfigParamsVersion;
-    int bSaveConfig = 0;
 
     if (ConfigOpenSection("CoreEvents", &l_CoreEventsConfig) != M64ERR_SUCCESS || l_CoreEventsConfig == NULL)
     {
@@ -522,14 +521,12 @@ int event_set_core_defaults(void)
         DebugMessage(M64MSG_WARNING, "No version number in 'CoreEvents' config section. Setting defaults.");
         ConfigDeleteSection("CoreEvents");
         ConfigOpenSection("CoreEvents", &l_CoreEventsConfig);
-        bSaveConfig = 1;
     }
     else if (((int) fConfigParamsVersion) != ((int) CONFIG_PARAM_VERSION))
     {
         DebugMessage(M64MSG_WARNING, "Incompatible version %.2f in 'CoreEvents' config section: current is %.2f. Setting defaults.", fConfigParamsVersion, (float) CONFIG_PARAM_VERSION);
         ConfigDeleteSection("CoreEvents");
         ConfigOpenSection("CoreEvents", &l_CoreEventsConfig);
-        bSaveConfig = 1;
     }
     else if ((CONFIG_PARAM_VERSION - fConfigParamsVersion) >= 0.0001f)
     {
@@ -537,7 +534,6 @@ int event_set_core_defaults(void)
         float fVersion = CONFIG_PARAM_VERSION;
         ConfigSetParameter(l_CoreEventsConfig, "Version", M64TYPE_FLOAT, &fVersion);
         DebugMessage(M64MSG_INFO, "Updating parameter set version in 'CoreEvents' config section to %.2f", fVersion);
-        bSaveConfig = 1;
     }
 
     ConfigSetDefaultFloat(l_CoreEventsConfig, "Version", CONFIG_PARAM_VERSION,  "Mupen64Plus CoreEvents config parameter set version number.  Please don't change this version number.");
@@ -575,9 +571,6 @@ int event_set_core_defaults(void)
     ConfigSetDefaultString(l_CoreEventsConfig, JoyCmdName[joyForward], "",    "Joystick event string for fast-forward");
     ConfigSetDefaultString(l_CoreEventsConfig, JoyCmdName[joyAdvance], "",    "Joystick event string for advancing by one frame when paused");
     ConfigSetDefaultString(l_CoreEventsConfig, JoyCmdName[joyGameshark], "",  "Joystick event string for pressing the game shark button");
-
-    if (bSaveConfig)
-        ConfigSaveSection("CoreEvents");
 
     return 1;
 }
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -266,21 +266,19 @@ static void main_check_inputs(void)
 int main_set_core_defaults(void)
 {
     float fConfigParamsVersion;
-    int bSaveConfig = 0, bUpgrade = 0;
+    int bUpgrade = 0;
 
     if (ConfigGetParameter(g_CoreConfig, "Version", M64TYPE_FLOAT, &fConfigParamsVersion, sizeof(float)) != M64ERR_SUCCESS)
     {
         DebugMessage(M64MSG_WARNING, "No version number in 'Core' config section. Setting defaults.");
         ConfigDeleteSection("Core");
         ConfigOpenSection("Core", &g_CoreConfig);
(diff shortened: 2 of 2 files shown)


bugfood:
The companion ui-console PR is here:
https://github.com/mupen64plus/mupen64plus-ui-console/pull/47

Thanks,
Corey

richard42:
There are a couple of things that I don't like about the design of this pull request. One is taking the static variable l_CoreInit and exposing it to the rest of the core library by making it global. The other is controlling the behavior of the configuration subsystem with specially named configuration variables. This is kind of a hack. It's not the config system's fault that it's saving the file; it is being told to do so by higher level code.

I think it would be better to achieve the desired behavior by adding some sort of core startup parameter instructing it not to save any options, so that it wont call ConfigSaveSection in main.c or event.c.

bugfood:
Thank you for your feedback. My goals in building it this way were:
1. Maintain backwards/forwards compatibility between UI and core.
2. Retain the same default behavior.

Are these priorities? If so, then I would appreciate a bit more guidance; I haven't seen another way to do it without breaking either of these goals somehow.

Specifically, the problematic chain of events is:
The UI starts the core via `CoreStartup()`, which calls `main_set_core_defaults()`, which calls `ConfigSaveSection()` (if it thinks it needs to change anything); all this happens before the UI can do anything to specify that configuration should not be saved.

I have seen two alternative approaches that I think could work:
1. Add a parameter to `CoreStartup()` so that the UI can specify whether configuration should be saved or not; I think this would break forward/backward compatibility with the UI.
2. Remove the call to `ConfigSaveSection()` from `main_set_core_defaults()` and rely on the UI to call `ConfigSaveSection()` if desired; configuration could either be saved unconditionally or, if desired, add a way to track whether the internal configuration state is "dirty" and only flush to the file when needed. This might change the default behavior for older UIs, but realistically, at least for now, plugins seem to call `ConfigSaveSection()` pretty often anyway.

Option 1 seems like it matches your suggestion above. Option 2 seems like it might be a better long-term approach, and I think it would allow me to also address your concern about the configuration variable, if it would be acceptable for me to (eventually) remove `ConfigSaveSection()` from non-UI plugins.

Please let me know how you would like me to proceed. I really care about trying to make this work, if I can.

Thanks,
Corey

richard42:
I think the best way forward is to remove the call to `ConfigSaveSection()` from inside of the `main_set_core_defaults()` function. Then, inside of the ui-console front-end, use the ConfigHasUnsavedChanges() on the Core section, and if there are unsaved changes (and if the `--nosaveoptions` was not specified), then save the section.

bugfood:
Thanks for your feedback. I have updated this PR to match your suggestion with respect to the core changes. I also had to change `event_set_core_defaults()` similarly.

The scope is less ambitious than before, but the change does allow the UI to be modified so that configuration saving does not happen automatically until plugin initialization--which doesn't fully fix the problem, but it gets further. If you will accept this change, then I can move onward to modifying plugins to not call `ConfigSaveSection()`.

I will update the corresponding UI-console PR shortly.

Thanks,
Corey

bugfood:
UI-console PR updated:
https://github.com/mupen64plus/mupen64plus-ui-console/pull/47

As long as `--nosaveoptions` is not present, the ui-console will call `ConfigSaveFile()`. Based on my reading of this, that makes it not necessary to check `ConfigHasUnsavedChanges()` on the core sections; if the entire configuration is about to be saved, there is no need to check individual sections. Is my understanding correct?

Thanks,
Corey

richard42:
yes you are correct, it saves the whole file in 2 places after starting the core.

PR #588, opened 2018-09-19
added support for aarch64

bkeys:
I added support for aarch64 with this addition to the makefile

### DIFF
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -1,3 +1,4 @@
+
 #/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
 # *   Mupen64plus - Makefile                                                *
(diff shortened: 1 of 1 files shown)


projects/unix/Makefile
```diff
@@ -131,7 +139,7 @@ OPTFLAGS ?= -O3 -flto
 WARNFLAGS ?= -Wall
 CFLAGS += -ffast-math -fno-strict-aliasing -fvisibility=hidden -I../../src -I../../src/asm_defines -DM64P_PARALLEL
 CXXFLAGS += -fvisibility-inlines-hidden
-LDLIBS +=  -lm
+LDLIBS +=  -lm -fPIC
```
richard42: The "-fPIC" option doesn't belong in the LDLIBS variable. You should add it to LDFLAGS instead, and put it in a system or cpu-specific section

bkeys:
Fixed it for ya.

PR #598, opened 2018-11-13
Fix errors introduced in #444

loganmc10:
These changes allow Rugrats and Xena to boot. Tested booting many other games, no regressions found.


fzurita:
Xena still doesn't seem to be booting, at least in Android.

PR #602, opened 2019-01-09
new_dynarec: Initial support for x64 and arm64

Gillou68310:


Zapeth:
Is there a description somewhere that explains the difference compared to the current implementation?

Like, is there anything in particular that this implementation aims to improve or does different in terms of performance, accuracy and/or portability?

fzurita:
arm64 performance seems to be about 10% worse than arm32 performance, at least in CBFD. Although, it's much better than nothing.

It's hard to determine where the performance difference is coming from since this is just the FPS counter I'm looking at in a static area. It could be other plugins are running slower when compiled in 64 bit vs 32 bit.

Gillou68310:
@fzurita your issue with clang is weird could you send me the verbose compiler+linker output?

Concerning the slowdown, did you try with non TLB games? Wondering if this is a TLB specific issue? For games running fullspeed in both situations, try with dummy audio and framelimiter off ;-)

@Zapeth this is just a basic port to support arm64 and x64 architectures. It's using the same original new_dynarec design. So compatibility and performance are expected to be the same as arm32 and x86 new_dynarec.

fzurita:
I attached the build output of the x86_64 build.

[ndkBuild_build_output.txt](https://github.com/mupen64plus/mupen64plus-core/files/2753512/ndkBuild_build_output.txt)

I will look into the TLB theory.

Gillou68310:
Thanks, just out of curiosity did you try building the old dynarec 64bits?

fzurita:
I have not tried yet.

Gillou68310:
ok you should definitely get the same issue, not sure where it comes from. Is GCC still an option with latest NDK?

fzurita:
GCC was removed in Android NDK 18: https://developer.android.com/ndk/downloads/revision_history

Gillou68310:
Ok

Gillou68310:
Hum on a second look I have a few non pc relative access to g_dev. Try commenting out those:
```mov  reg,   g_dev_*```

**Comment by fzurita** [comment]
Sure, will do. By the way, the performance difference may have nothing to do with 32 vs 64 bit. I was comparing a build before this pull request to after this pull request. When comparing after the pull request, 32 bit vs 64 bit performance is identical.

When comparing 32 bit only, before and after the pull request, there is about a 5% performance loss after the pull request of the app overall in CBFD. I will need to grab more data point in other games. This was with audio plugin disabled.

**Comment by Gillou68310** [comment]
That's good news! Could you bisect which commit causes the slowdown?

**Comment by fzurita** [comment]
I commented out all the mov operations in linkage_x64.asm and in dyna_start.asm and it allows the core library to compile.

I will try to bisect what commit created the slowdown in the 32 bit arm dynarec next.

**Comment by Gillou68310** [comment]
Thanks I will push a fix for those non pc relative access to g_dev tomorrow ;-)

**Comment by Gillou68310** [comment]
BTW I'm not sure if android is using the x64 or amd64 ABI see assem_x64.h

**Comment by Gillou68310** [comment]
See https://github.com/mupen64plus/mupen64plus-core/issues/603

**Comment by fzurita** [comment]
Good question, I'm not sure how to tell. I would guess AMD64 since it's clang and linux based.

Also, once it's compiling in Android, I don't have a device to test with unfortunately, although Android emulator may work.

**Comment by fzurita** [comment]
This commit introduced the slowdown:

5ff85aac1d59e5d717e419b9867de2a9d0014f71

**Comment by Gillou68310** [comment]
@fzurita Thanks for bisecting! I'll investigate the issue.
https://github.com/mupen64plus/mupen64plus-core/pull/606/commits/1e6d8bdf99efd6e6de6460492399bdedd9de39c0 should fix the build issue on android, I'll also add a fix for the slowdown in this PR once I find the cause.
Could you send me the core disassembly I'll check the calling convention to find out which ABI is used.

**Comment by fzurita** [comment]
Here is the disassembly:

[disassembly.txt](https://github.com/mupen64plus/mupen64plus-core/files/2754964/disassembly.txt)

**Comment by Gillou68310** [comment]
Thanks but could you generate the disassembly in intel format I hate AT&T

**Comment by fzurita** [comment]
Hmmm, any idea on how to make objdump do that?

**Comment by fzurita** [comment]
Never mind, found it: objdump -M intel -d program_name

**Comment by fzurita** [comment]
New one:

[disassembly.txt](https://github.com/mupen64plus/mupen64plus-core/files/2755152/disassembly.txt)

**Comment by Gillou68310** [comment]
Thanks it's much more readable!
Anyway definitely amd64 ABI. If _WIN32 is not defined on android we should be good ;-)
Do you have an android emulator to test?

**Comment by Gillou68310** [comment]
Damn I cannot reproduce the slowdown issue on my odroid do you have a savestate for CBFD where the issue is noticeable for you?

**Comment by fzurita** [comment]
Yes, it turns out I can make a x86_64 android emulator. It shouldn't be an issue. I'll get a save state.

**Comment by Gillou68310** [comment]
Thanks!

**Comment by fzurita** [comment]
I wonder if the slowdown is compiler dependent. Have you tried compiling using clang?

**Comment by Gillou68310** [comment]
Reading from the dynarec cache seems to be the issue.

CBFD is doing a lot of code invalidation thus get_bounds is called really often. My parse_dirty_stub function is doing more cache memory access than the previous get_bounds function.

Could you check that this is the issue by replacing get_bounds by the previous version of it?

**Comment by fzurita** [comment]
Sure, I'll do that some time tonight once I'm out of work.

**Comment by fzurita** [comment]
@Gillou68310 is the above backwards? The old code matches your first statement, so I will assume replace old code with new code. After doing that and testing the same scene in CBFD, performance goes to less than 1FPS, am I doing something wrong?

**Comment by fzurita** [comment]
I did it correctly the second time around. Making that fix does put performance back to previous levels.

**Comment by fzurita** [comment]
I just tested in an Android x86_64 emulated image. The dynarec seems to run fine. I didn't do any extensive testing since it was so slow though.

**Comment by Gillou68310** [comment]
@fzurita could you test this branch?
https://github.com/Gillou68310/mupen64plus-core/commit/5eaf9a1e640c2398f8dd8d4f674e2f033bc1c215

I tried to reduce dynarec cache access to a maximum, let me know if it helps

**Comment by fzurita** [comment]
@Gillou68310 That code doesn't seem to compile for x86. Either way, I tested with arm64 and now it runs faster than the original! Nice job. It's about a 5% to 10% faster than before adding the 64 bit new dynarecs.

**Comment by fzurita** [comment]
I put this update in the playstore. I'm seeing some crashes in memcmp, possibly this:

```
static int verify_dirty(void *addr)
{
  uintptr_t source=0;
  uintptr_t copy=0;
  u_int len=0;
  uintptr_t verifier=0;

  void *ptr=parse_dirty_stub(addr,&source,&copy,&len,&verifier);
  assert(ptr!=NULL);
  assert((verifier==(uintptr_t)verify_code)||(verifier==(uintptr_t)verify_code_vm)||(verifier==(uintptr_t)verify_code_ds));

  if(verifier==(uintptr_t)verify_code_vm||verifier==(uintptr_t)verify_code_ds) {
    unsigned int page=source>>12;
    uintptr_t map_value=g_dev.r4300.new_dynarec_hot_state.memory_map[page];
    if((intptr_t)map_value<(intptr_t)0) return 0;
    while(page<((source+len-1)>>12)) {
      if((g_dev.r4300.new_dynarec_hot_state.memory_map[++page]<<2)!=(map_value<<2)) return 0;
    }
    source = source+(map_value<<2);
  }
  //DebugMessage(M64MSG_VERBOSE, "verify_dirty: %x %x %x",source,copy,len);
  return !memcmp((void *)source,(void *)copy,len); // HERE
}
```

It's hard to get a backtrace, but if you think that's not an issue, I'll generate one.

Gillou68310:
I had the crash too I already pushed a fix:
https://github.com/mupen64plus/mupen64plus-core/pull/606/commits/49573d21c1af9fd6cf1d54d75eb440977599b334



> @Gillou68310 That code doesn't seem to compile for x86. Either way, I tested with arm64 and now it runs faster than the original! Nice job. It's about a 5% to 10% faster than before adding the 64 bit new dynarecs.

Thanks for testing! I forgot to say it's x64/arm64 only for now ;-)
Now that it I know it gives a small boost I will add support for x86/arm32

Issue #603, opened 2019-01-13
New dynarec not working on x64


loganmc10:
The following is from Mario 64 but other games are affected as well:
```
Core Error: Unknown MTC0 write: 64
Core Error: Unknown MTC0 write: 68
Core Error: Unknown MTC0 write: 72
```

Tested on Windows, compiled with MSYS2. 32-bit version is working normally
@Gillou68310

Gillou68310:
Probably MSVC macro not defined see assem_x64.h

PR #604, opened 2019-01-13
Fix mingw build

loganmc10:
Fixes https://github.com/mupen64plus/mupen64plus-core/issues/603


loganmc10:
Tested and working on my PC

Gillou68310:
Is there no macro shared between both mingw and visual studio? WIN32 maybe?

loganmc10:
_WIN32 seems to work on MinGW, I assume it works in Visual Studio as well. I've updated the PR

Gillou68310:
Thanks I'll check on visual studio

Gillou68310:
Looks good

PR #614, opened 2019-01-29
Make stop_after_jal a save state parameter instead of configuration

fzurita:
That way it can be used when loading save states instead of having speculate recompilation always disabled for compatibility with save states.

### DIFF
--- a/doc/emuwiki-api-doc/Mupen64Plus-Core-Parameters.mediawiki
+++ b/doc/emuwiki-api-doc/Mupen64Plus-Core-Parameters.mediawiki
@@ -59,10 +59,6 @@ These are standard parameters which are used by the Mupen64Plus Core library.  T
 |M64TYPE_INT
 |Force number of cycles per emulated instruction when set greater than 0.
 |-
-|DisableSpecRecomp
-|M64TYPE_BOOL
-|Disable speculative precompilation in new dynarec.
-|-
 |}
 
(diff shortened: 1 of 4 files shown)


fzurita:
@Gillou68310 Here you go. It seems to work ok in my ARM devices, but it should probably be tested in a build with NEW_DYNAREC disabled.

fzurita:
Rebased against master branch.

richard42:
I was actually not going to merge this pull request. It has a problem. That 4k buffer at the end (data_0001_0200) was designed to be a fixed block for adding new parameters as necessary. You shouldn't have added a new stop_after_jal_data block after that data_0001_0200 block, you should have just used the empty space which still remains in that block.  The big new block of code which begins with:
`else if (version == 0x00010500)`
doesn't need to be there.

fzurita:
Ok, gotcha, I missed that part of the design by not reading the code carefully enough. Feel free to rebase to remove to change from the master branch. I can make a new pull request tomorrow with the requested change.

Gillou68310:
Sorry @richard42 I missed that too when reviewing the PR. Do you want me to revert it in the meantime?

richard42:
@fzurita it's not your fault. There ought to be some comments in that code explaining that feature of the design. It's not obvious from just looking at the code. Thanks for re-doing it.
@Gillou68310 if you don't mind reverting this you would save me the trouble. Thanks!

fzurita:
@Gillou68310 you don't have to revert if you don't want... I'll have an update to this code in https://github.com/mupen64plus/mupen64plus-core/pull/615

Also... I found a bug in the original code that makes save states produced with the NEW_DYNAREC defined to not load when the NEW_DYNAREC is not defined.

PR #615, opened 2019-01-30
Fixes for last save state update

fzurita:
Minor cleanup from last pull request. This disables Speculative recompilation when loading pj64 save states.

### DIFF
--- a/src/main/savestates.c
+++ b/src/main/savestates.c
@@ -195,7 +195,6 @@ static int savestates_load_m64p(struct device* dev, char *filepath)
     char queue[1024];
     unsigned char using_tlb_data[4];
     unsigned char data_0001_0200[4096]; // 4k for extra state from v1.2
-    unsigned char stop_after_jal_data[4];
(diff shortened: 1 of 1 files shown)


fzurita:
I made some more corrections. The additional stop_after_jal data was not being stored in the correct place.  Because of that, the correct value was also not being read.

I did a quick test of writing a random value for the ```stop_after_jal``` flag and the same value was correctly read back.

In addition, we seem to be relying on ```unsigned int``` being 32 bits and ```short``` being 16 bits, which is not guaranteed by the C standard. So potentially, a save state created in one platform may not be compatible in another platform. I don't think this is the problem with not being able to load ```NEW_DYNAREC``` save states on build without ```NEW_DYNAREC```.

fzurita:
@richard42 can you review and make comments or merge?

richard42:
yep, looks good to me.

fzurita:
Thanks!

PR #618, opened 2019-02-03
__clear_cache fix for ARM 64 devices plus ANDROID cleanup

fzurita:
This implements a fix for a bug in some arm 64 bit CPUs as seen here:

https://github.com/dolphin-emu/dolphin/pull/4204

Also, removed some ANDROID specific code that is no longer needed.

@Gillou68310 you should review this please. I didn't see any performance penalties.

Should we attribute the source of the fix in the code?

### DIFF
--- a/src/device/r4300/new_dynarec/arm/assem_arm.c
+++ b/src/device/r4300/new_dynarec/arm/assem_arm.c
@@ -230,6 +230,11 @@ static u_int jump_table_symbols[] = {
   (int)breakpoint
 };
 
+static void cache_flush(char* start, char* end)
+{
+    __clear_cache(start, end);
+}
+
 /* Linker */
 static void set_jump_target(int addr,u_int target)
 {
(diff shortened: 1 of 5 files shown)


Gillou68310:
Well I don't like intrinsics but I guess I won't ask you to write it in assembly ;-)
You already gave the source in this PR I don't think it's necessary to add it in the source code.

fzurita:
Feel free to rewrite it in assembly :-)

Gillou68310:
Maybe later

fzurita:
Ok to merge then?

Gillou68310:
Yep

fzurita:
@Gillou68310 I'm getting a warning from static analysis with clang tidy (It was added on Android Studio 3.3):
```
Line 4453: Local variable 'memtarget' might not have been initialized
````

It may explain all the breakpoint crashes I have been seeing in the play store.
```signal 5 (SIGTRAP), code 1 (TRAP_BRKPT)```
 Thoughts?

**Comment by Gillou68310** [comment]
That uninitialized variable is bad and should be fixed but I don't think it's the cause of the breakpoint.

Currently SWL/SWR/SDL/SDR instructions have no fallback to interpreted opcode, which means that a store with no valid TLB map or an address outside RDRAM (eg: 0xa4000000) will trigger a breakpoint. This is true for all platform except ARM which has no breakpoint.

We really need to find the game that triggers this so I can implement interpreted fallback properly.

**Comment by fzurita** [comment]
Also, I'm still seeing some of these crashes with bad core dumps in arm64 devices:
```
signal 4 (SIGILL), code 1 (ILL_ILLOPC)
```

Any chance we are emitting some bad instruction? I'm not seeing it in arm32.

Gillou68310:
that's possible indeed

fzurita:
Too bad I can't find which game

Gillou68310:
¯\_(ツ)_/¯

PR #620, opened 2019-02-04
safer config file writing

bugfood:
The configuration file writing is subject to two problem situations that
could end up leaving the file empty or with incomplete data:
1. If the program were to terminate while writing to the configuration
   file.
2. If the system were to crash or lose power shortly after the program
   has finished writing the configuration file.

These are covered in detail in:
https://lwn.net/Articles/457667/
https://lwn.net/Articles/322823/

Use a temporary file to mitigate issue #1.

Use fsync() on the file and the directory to mitigate issue #2. This is
not implemented for Windows because I was not able to find evidence that
Windows supports the necessary system calls (or wrappers); I can change
this if I get some feedback giving me direction.

### DIFF
--- a/src/api/config.c
+++ b/src/api/config.c
@@ -28,6 +28,12 @@
 #include <stdlib.h>
 #include <string.h>
 
+#if (!defined(_WIN32))
+#include <fcntl.h>
+#include <sys/types.h>
+#include <unistd.h>
+#endif
+
 #define M64P_CORE_PROTOTYPES 1
 #include "callbacks.h"
 #include "config.h"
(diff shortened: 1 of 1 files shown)


richard42:
This seems unnecessarily complicated. Have you ever experienced any corruption or data loss of the configuration file?

bugfood:
I have only noticed it when running under gdb and restarting the program during the config writing.

This was mostly an exercise to "do the right thing". In my experience, rare things eventually happen, given enough chances.

richard42:
I share your philosophy with doing the right thing, and of course rare things aren't impossible.. But at the same time, the harm is pretty small (just delete the file and a new one gets created). And complexity is the enemy of software development; everything that gets added has to justify its added complexity with new or improved functionality. And I don't think this one makes the cut, sorry.

PR #606, opened 2019-01-14
new_dynarec fixes

Gillou68310:

PR #621, opened 2019-02-09
Allow 16MB of RDRAM

fzurita:
So allegedly the N64 allowed for up to expansions paks of 16MB. I modified the code to allow that. @bsmiles32 Does everything look right?

### DIFF
--- a/src/main/savestates.c
+++ b/src/main/savestates.c
@@ -57,7 +57,7 @@ enum { GB_CART_FINGERPRINT_OFFSET = 0x134 };
 enum { DD_DISK_ID_OFFSET = 0x43670 };
 
 static const char* savestate_magic = "M64+SAVE";
-static const int savestate_latest_version = 0x00010500;  /* 1.5 */
+static const int savestate_latest_version = 0x00010600;  /* 1.6 */
 static const unsigned char pj64_magic[4] = { 0xC8, 0xA6, 0xD8, 0x23 };
 
 static savestates_job job = savestates_job_nothing;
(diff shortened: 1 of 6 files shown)


src/main/savestates.c
```diff
@@ -417,7 +429,16 @@ static int savestates_load_m64p(struct device* dev, char *filepath)
     dev->dp.dps_regs[DPS_BUFTEST_ADDR_REG] = GETDATA(curr, uint32_t);
     dev->dp.dps_regs[DPS_BUFTEST_DATA_REG] = GETDATA(curr, uint32_t);
 
-    COPYARRAY(dev->rdram.dram, curr, uint32_t, RDRAM_MAX_SIZE/4);
```
fzurita: This is an issue in the existing code, we always copy the max even though we may have only allocated 4MB.

fzurita:
@richard42 Do you think this is the right approach for save states?

richard42:
In order to retain backwards compatibility, the new data (extra 8MB of RDRAM) should be at the end of the savestate.  Also, I think we shouldn't merge this change until after the release.

fzurita:
Do you mean for older versions of the program to be able to load newer save states?

I did test loading older save states after this change and it works ok.

fzurita:
Also, if you think the save state modification is too much before the next release. I'll make another pull request that will fix some crashes I'm seeing without allowing the user to select 16MB of RDRAM or modifying save states.

richard42:
I'm okay with a pull request to fix crashes. Are there any games or homebrew software which uses 16MB of RAM?

fzurita:
The only homebrew that I'm aware of is Super Mario 64 online, there may be others though. This pull request was only made because I saw the opportunity to allow 16MB of RDRAM while debugging some crash reports I got for my app in the Google Play store, Like this one: https://github.com/mupen64plus/mupen64plus-core/pull/606#issuecomment-458388545 and see the follow up post as to how it could crash.

fzurita:
I think this will better protect against times when a save state is created when the core is running with 8MB of RDRAM and then it's loaded when the core is running with 4MB of RDRAM.

fzurita:
Will redo this once the other pull request is accepted

PR #623, opened 2019-02-09
Always allocate 16MB RDRAM even though only 8MB is supported

fzurita:
Also, correctly check RDRAM size when saving and loading save states.

I think this will fix the crash that was reported here:

https://github.com/mupen64plus/mupen64plus-core/pull/606#issuecomment-458388545

### DIFF
--- a/src/device/memory/memory.c
+++ b/src/device/memory/memory.c
@@ -215,7 +215,7 @@ void apply_mem_mapping(struct memory* mem, const struct mem_mapping* mapping)
 
 enum {
     MB_RDRAM_DRAM = 0,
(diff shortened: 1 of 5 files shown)


fzurita:
@bsmiles32 Is this ok with you? I touched some of the areas you modified recently.

fzurita:
There is something wrong here, it causes South Park to not boot.

fzurita:
Ok, fixed.

loper12:
Wouldn't it just be much better to just select 16MB of ram? Why force it?

loper12:
It won't fix all games,  just a select few, so why the globalization?

fzurita:
I was only following the pattern of the 8MB implemention. Everything you say also applies to 8MB as well.

loper12:
hmmm

Issue #577, opened 2018-07-03
Segfault from: #446 "Add support for 64 Disk Drive"


hissingshark:
Hi, it's the first time this year I've rebuilt from master and I'm getting a segfault, which I've bisected back to commit e302bc2 "Add support for 64 Disk Drive".

Mario Kart shows the spinning logo at the start but segfaults just before the start screen.
Super Smash Bros segfaults during the intro animation.

I'm on armv8, Amlogic S905x (Mali 450 GPU). GLES2. And I was using the mupen64plus-video-n64 plugin.

Anything I can provide to assist debugging?

bsmiles32:
Hi @hissingshark 

thanks for reporting (and bissecting) your issue here. We will need some more information in order to assist in the debugging. See below:

- Which emulation mode do you use : interpreter, cached interpreter, or new_dynarec ?
- Can you provide a backtrace of said crash ? You will need to compile with debug symbols (eg DEBUG=1) for it to be usable. Then running mupen64plus under gdb using something along the lines of :

```
gdb --args mupen64plus --corelib <path_to_my_compiled_version_of_libmupen64plus> --datadir <path_to_corresponding_data_dir>  --emumode <{0,1,2}> my_rom.z64
```
then inside the gdb console type "r" (for run), and when the program crashes, you can type "bt" (backtrace) to get the backtrace at the moment of the crash.

hissingshark:
Thought it might be something like that!  Thanks for getting back to me.
Well, it's mixed results here.
The segfault doesn't reproduce if compiled as DEBUG, so I can't get a backtrace.  I think this is what the kids call a "Heisenbug".  When reading up on the term "optimisation" came up, so I thought I'd try compiling with the `-O2` CFLAG instead of the `-O3` I had been using and no segfault either.

bsmiles32:
When building did you started from a "fresh" compilation or not (eg. make clean && make all) ?
I ask because, in order to share C struct definitions between C and assembler, we use some "magic" which generate some header files with proper members offsets. These headers MUST be regenereted when compiler options get changed or when some structures get updated. Otherwise assembler will use outdated offsets and cause hard to debug crashes.

hissingshark:
That's clean.
I've just reaffirmed for you with an unadulterated git clone -> build -> run -> segfault.

fzurita:
[...]

PR #649, opened 2019-05-24
Build option to disable 64 Disk Drive support

hissingshark:
There is a longstanding [issue](https://github.com/mupen64plus/mupen64plus-core/issues/577)  since the addition of the 64 Disk Drive support.  This is causing some ARM devices to segfault at emulator start.

This commit adds a build option `64DISKDRIVE`.
It defaults to `64DISKDRIVE=1` as I expect most devices are unaffected.
Those platforms vulnerable to it can use `64DISKDRIVE=0`.

Many Thanks

### DIFF
--- a/README
+++ b/README
@@ -63,6 +63,7 @@ Type 'make' by itself to view all available build options:
      NO_ASM=1       == build without assembly (no dynamic recompiler or MMX/SSE code)
      USE_GLES=1     == build against GLESv2 instead of OpenGL
      VC=1           == build against Broadcom Videocore GLESv2
+     NO_64DD=1      == build without 64 Disk Drive support
      NEON=1         == (ARM only) build for hard floating point environments
      VFP_HARD=1     == (ARM only) full hardware floating point ABI
      SHAREDIR=path  == extra path to search for shared data files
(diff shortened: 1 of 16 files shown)


hissingshark:
Ah, sorry forgot about the Visual C++ files:
* mupen64plus-core.vcxproj
* mupen64plus-core.vcxproj.filters

I think those explain the AppVeyor build fail.  Not familiar with configuring those and no way to test locally (no Windows box here) so will need to read up.

hissingshark:
It seemed like there was a problem with the checks for desktop builds because the Makefile wasn't setting the pre-processor macro, but now I'm not so sure.
Regardless I've started again and made this a "disable feature" build option.  Now checking successfully.

`NO_64DD=1      == build without 64 Disk Drive support`

I hope this feature is acceptable.

richard42:
I'm okay with this capability in general (being able to disable 64DD support at build time) but I have a problem with this implementation. There are 50 new #ifdef macro blocks in this merge request (that's a lot), and the function interfaces (internal API) are different depending upon whether or not this macro is defined.

There must be a simpler way to achieve this with fewer #ifdefs, like setting the dd_controller pointer to NULL if it's not supported, and keeping the existing internal API.

loganmc10:
I think it would be a lot better to figure out why the new dynarec crashes with 64DD support, there shouldn't really be a reason to remove DD support like this

fzurita:
Also, Android does not appear to have this problem. I agree with @loganmc10, if @hissingshark could provide a backtrace to the core dump, I'm sure the issue could be figured out.

PR #662, opened 2019-07-12
Log error when VSync set fails.

dankcushions:
This helped me track down https://github.com/ricrpi/mupen64plus-video-gles2n64/pull/27


dankcushions:
Replaced by https://github.com/mupen64plus/mupen64plus-core/pull/670

PR #666, opened 2019-07-16
Add speed limiter toggle, reintroduce SDL_PumpEvents for non-android platforms

### LINKED ISSUE #646: Pausing causes the emulator to lock up
As of late, it seems pausing the emulator causes it to lock up regardless of plugin configuration and CPU emulation setting on both of my Linux machines (running Fedora 29 and Ubuntu 18.10) on any game.

I am building from master.

mountainflaw:
The SDL_PumpEvents change fixes #646 , and is also required for the speed limiter toggle to not cause the emulator to hang.


fzurita:
Android is stubbing out SDL now, so ```SDL_PumpEvents()``` can be safely introduced back without the ```ANDROID``` guard.

GhostlyDark:
3 years later, what's the reason it's not merged?

Issue #663, opened 2019-07-12
VidExt_GL_SetAttribute sets SDL_GL_SetSwapInterval prematurely


dankcushions:
The spec for VidExt_GL_SetAttribute says:
> This function is used to set certain OpenGL attributes which must be specified **before** creating the rendering window with VidExt_SetVideoMode

One of these attributes is `M64P_GL_SWAP_CONTROL`, which controls VSync via `SDL_GL_SetSwapInterval` See 
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/api/vidext.c#L431
SetSwapInterval requires a context: https://wiki.libsdl.org/SDL_GL_SetSwapInterval - if you don't, it will error with `No OpenGL context has been made current` (you can see this error via my PR https://github.com/mupen64plus/mupen64plus-core/pull/662)

However, when launching mupen64 via command line on linux (not sure if that's relevant), it appears that GL context is only set via `VidExt_SetVideoMode` - see https://github.com/mupen64plus/mupen64plus-core/blob/77a2f8813251b5257382ec477c3f0ef3ea49c37f/src/api/vidext_sdl2_compat.h#L470

...but the API for VidExt_SetVideoMode says:
> This function creates a rendering window or switches into a fullscreen video mode. Any desired OpenGL attributes should be set **before** calling this function.

Interesting thread here: http://forums.libsdl.org/viewtopic.php?p=43308

should SwapInterval really be considered an attribute? It seems like it would be better as a parameter in SetVideoMode, however I'm sure you don't want to change the API at this stage.

@loganmc10 had a suggestion in https://github.com/gonetz/GLideN64/pull/2070:
> instead of trying to call SDL_GL_SetSwapInterval during VidExt_GL_SetAttribute, you could just set a variable to "true" and then call SDL_GL_SetSwapInterval in VidExt_SetVideoMode based on whatever the variable was set to

sound good?

dankcushions:
I tried @loganmc10's idea - https://github.com/dankcushions/mupen64plus-core/commit/3e07fdd82d864ed22a1a665128d820948b56768e

seemed to work! I will PR after a proper test

loganmc10:
You should probably set l_SwapControl to it's default value in the VidExt Init function as well

dankcushions:
Done! see https://github.com/mupen64plus/mupen64plus-core/pull/670

PR #670, opened 2019-07-20
Set VSync (SDL_GL_SetSwapInterval) after GL context is available

dankcushions:
This fixes https://github.com/mupen64plus/mupen64plus-core/issues/663

In my case, before this fix VSync was always ON in linux and standalone commandline (`--noosd`) mupen64plus.

### DIFF
--- a/src/api/vidext.c
+++ b/src/api/vidext.c
@@ -47,6 +47,7 @@ static m64p_video_extension_functions l_ExternalVideoFuncTable = {12, NULL, NULL
 static int l_VideoExtensionActive = 0;
 static int l_VideoOutputActive = 0;
 static int l_Fullscreen = 0;
+static int l_SwapControl = 0;
 static SDL_Surface *l_pScreen = NULL;
 
 /* global function for use by frontend.c */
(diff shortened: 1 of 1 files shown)


dankcushions:
the sdl 2 ifdefs are wrong now. will fix soon!

dankcushions:
Building properly on non-SDL 2.0.0. now. Good to go!

dankcushions:
Anything I can do to get this over the line? Seems to have gone a little quiet around here :eyes:

psyke83:
This appears to break mupen64plus on RPI4 when launched via an Xorg context. No window will appear, and In the log, I see `Core Error: SDL swap interval (VSync) set failed: That operation is not supported`, followed by an application hang.

It's late here, so I'll try to provide more detail after I've had more time to check it out. Presumably the solution is to omit the `return M64ERR_SYSTEM_FAIL;` entirely when the swap interval is not successfully changed, as it shouldn't be a fatal error. I'm not certain why the function is failing in the first place, though.

psyke83:
Please disregard my last post - apologies. I didn't notice that my Xorg session was using the llvmpipe software driver, which doesn't support setting the swap interval. After fixing that, the PR works fine, and fixes vsync with both the `kmsdrm` and `x11` SDL2 drivers.

Without this PR, `SDL_GL_SetSwapInterval` returns the `No OpenGL context has been made current` SDL [https://github.com/spurious/SDL-mirror/blob/master/src/video/SDL_video.c#L3586](error). In a case such as this, the swap interval will depend on whatever default the internal driver sets... `x11` and `rpi` is either `1` or `-1` (adaptive sync), and `kmsdrm` defaults to `0`. So this change is highly recommended to guarantee the user setting is applied correctly.

Gillou68310:
@dankcushions I want to merge this, could you rebase and squash to one commit?

dankcushions:
@Gillou68310 done!

Gillou68310:
thanks

PR #682, opened 2019-10-21
MPAL's refresh rate is 60 Hz

Nabile-Rahmani:
See: https://en.wikipedia.org/wiki/PAL-M

>It is unique among analog TV systems in that it combines the 525-line 30 frames-per-second System M with the PAL colour encoding system (using very nearly the NTSC colour subcarrier frequency), unlike all other countries which pair PAL with 625-line systems and NTSC with 525-line systems.


Narann:
Looks OK to me.

PR #685, opened 2019-10-27
Allow carts of up to 244 MB

fzurita:
From some discord conversations, some ROMs that are greater than 64MB are causing the core to crash. These ROMs run on the real console, so it's definitely an issue in the core. Making the below two changes allows these ROMs to run with no perceived issues. From discord conversations, the maximum allowed ROM is 244 MB.

Also, searching through the code, there are a number of other places where 0x3FFFFFF is being used as a mask. It's hard to tell how many of these places refer to ROM address offsets. So this change may allow data to be stored past 64MB, but game code past that point may still have issues.

### DIFF
--- a/src/device/cart/cart_rom.c
+++ b/src/device/cart/cart_rom.c
@@ -31,7 +31,7 @@
 #define __STDC_FORMAT_MACROS
 #include <inttypes.h>
 
-#define CART_ROM_ADDR_MASK UINT32_C(0x03ffffff);
+#define CART_ROM_ADDR_MASK UINT32_C(0x0effffff);
 
 
 void init_cart_rom(struct cart_rom* cart_rom,
--- a/src/device/cart/cart_rom.h
+++ b/src/device/cart/cart_rom.h
@@ -42,7 +42,7 @@ struct cart_rom
 
 static osal_inline uint32_t rom_address(uint32_t address)
 {
-    return (address & 0x03fffffc);
+    return (address & 0x0efffffc);
 }
 
 void init_cart_rom(struct cart_rom* cart_rom,


fzurita:
These are a few other places where the code may need to be updated as well, but I'm not 100% sure:
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/new_dynarec/arm64/assem_arm64.c#L302
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/new_dynarec/arm64/assem_arm64.c#L960
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/new_dynarec/arm64/assem_arm64.c#L5171
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/cached_interp.c#L406
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/idec.c#L68
https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/pure_interp.c#L102

fzurita:
@Gillou68310 Do you agree with those other places?

fzurita:
Never mind, CBFD crashes when the ROM mask is changed.... I'll look more into it.

Gillou68310:
The changes in assem_arm64 are wrong for sure. Not sure about the others I can't check in details rn

fzurita:
Ok, thanks for checking.

Nabile-Rahmani:
@fzurita fpganugga mentioned in the 64drive documentation that due to limitations in the real parallel interface, some parts of the highest cartridge map isn't available, hence his decision to cap the ROM size to 240 MB and use the rest for MMIO, but without going in-depth in the explanation.

But technically, the cartridge space spans a whopping 252 MB (0x10000000 to 0x1FBFFFFF).

fzurita:
Gotcha, yeah, I couldn't get this to work in the end. This at the very least allows carts bigger than 64MB to not crash, but I couldn't get a test ROM that was provided to me from working as expected.

PR #691, opened 2019-12-17
Implemented additional trap opcodes

MrCheeze:
As far as I know, these instructions aren't used in retail roms, but they exist on real hardware and need to work for the purpose of an OoT ACE project that is currently in the works.

It's worth mentioning that I'm not sure the existing TEQ implementation is strictly correct, given that it just stops rather than follow what the N64 would do in that situation (crash screen)? Regardless I've made the implementation here exactly analogous to what what currently exists for TEQ.

For reference, see also a similar PR by krimtonz to projec64 earlier today:
https://github.com/project64/project64/pull/1684


Narann:
Have you anything to share to reproduce the problem ?

loganmc10:
This isn't really the same as the PJ64 PR. This will just stop the emulator when these opcodes are executed, it won't generate any kind of exception.

It would be good to have some kind of test ROM that we could test the trap opcode with to make sure it gets implemented correctly

loganmc10:
I'm guessing the right thing to do, instead of stopping the emulator, would be:

```
uint32_t* cp0_regs = r4300_cp0_regs(&r4300->cp0);
cp0_regs[CP0_CAUSE_REG] = CP0_CAUSE_EXCCODE_TR;
exception_general(r4300);
```

bsmiles32:
Replaced by PR #745 . Continue discussion there.

PR #692, opened 2019-12-22
Fix VI interrupt not being called in certain cases

Zapeth:
Not sure if this affects any commercial roms, but it at least enables certain homebrew roms to display something (like some of [these test roms](https://github.com/PeterLemon/N64))

### DIFF
--- a/src/device/rcp/vi/vi_controller.c
+++ b/src/device/rcp/vi/vi_controller.c
@@ -58,6 +58,17 @@ unsigned int vi_expected_refresh_rate_from_tv_standard(m64p_system_type tv_stand
     }
 }
 
+void set_vi_vertical_interrupt(struct vi_controller* vi)
+{
+    if (!get_event(&vi->mi->r4300->cp0.q, VI_INT) && (vi->regs[VI_V_INTR_REG] < vi->regs[VI_V_SYNC_REG]))
+    {
+        const uint32_t* cp0_regs = r4300_cp0_regs(&vi->mi->r4300->cp0);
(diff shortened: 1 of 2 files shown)


loganmc10:
There should be a lot of testing if this is going to be merged. Calling a VI interrupt at the wrong time can break a lot of games

Zapeth:
I haven't done extensive testing (and I'm not planning to), but I haven't noticed any issues with other roms that I have been using so far.

In theory this change shouldn't cause issues since it would only apply to cases where a VI interrupt is set up on MI but hasn't been registered at cp0 due to `vi->regs[VI_V_INTR_REG] < vi->regs[VI_V_SYNC_REG]` condition not holding (which is essentially what happens with these particular test roms).

Not to mention that these roms have reportedly been tested on real hardware, so in that regards the current behavior is incorrect anyway.

Zapeth:
Another issue with the current implementation is that the event count of a `VI_INT` is remains the same when `VI_V_SYNC_REG` changes.

Though this might not have any (noticable) effect on games, especially since the overall emulation is far from accurate anyway.

Best way to fix this would probably require something like an `update_interrupt_event_count` function in interrupt.c, but thats neither in the scope of this PR nor is it something I want to spend my time with.

loganmc10:
I'll test this out tomorrow, you are right that it should be a pretty safe change, and it would be good to get those test ROMs working better

loganmc10:
I think this is OK to merge, I tested a good number of commercial games and didn't see any issues. The way it is written should make it pretty harmless as well

PR #693, opened 2019-12-22
Fix appveyor VS project path

Zapeth:


Zapeth:
Closed in favor of https://github.com/mupen64plus/mupen64plus-core/pull/694

PR #694, opened 2019-12-23
Let MSBuild do its job and implement a proper AppVeyor with artifact packaging

Jj0YzL5nvJ:
~~P.S: Stop pleasing the human sacrifices and goats bleeding demanded by the horrendous VS2017 IDE. MSBuild is intelligent enough to understand which is the most appropriate SDK agnostically in base of "DefaultPlatformToolset" and "PropertySheets".~~

Hack for `WindowsTargetPlatformVersion` extracted from [here](https://developercommunity.visualstudio.com/content/problem/140294/windowstargetplatformversion-makes-it-impossible-t.html).


appveyor.yml
```diff
@@ -1,19 +1,41 @@
...
+    configuration: New_Dynarec_Release
+  - platform: x64
+    configuration: New_Dynarec_Release
 
 before_build:
-- cmd: git clone --depth 1 https://github.com/mupen64plus/mupen64plus-win32-deps.git ../mupen64plus-win32-deps
-build:
-  project: projects/VisualStudio2013/mupen64plus-core.vcxproj
-  verbosity: minimal
+  - git clone --depth 1 https://github.com/mupen64plus/mupen64plus-win32-deps.git ..\mupen64plus-win32-deps
```
Narann: Is there any rational about removing `cmd:` ?

projects/msvc/mupen64plus-core.vcxproj
```diff
@@ -677,13 +677,13 @@
       <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Release|Win32'">true</ExcludedFromBuild>
       <ExcludedFromBuild Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Release|x64'">false</ExcludedFromBuild>
       <FileType>Document</FileType>
-      <Command Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Debug|x64'">"..\..\..\mupen64plus-win32-deps\nasm-2.14.02\nasm.exe" -i ..\..\src\ -i ..\..\src\asm_defines\ -o $(IntDir)linkage_x64.obj -f win64 -d WIN64 %(FullPath)</Command>
+      <Command Condition="'$(Configuration)|$(Platform)'=='New_Dynarec_Debug|x64'">"..\..\..\mupen64plus-win32-deps\nasm-2.14.02\x64\nasm.exe" -i ..\..\src\ -i ..\..\src\asm_defines\ -o $(IntDir)linkage_x64.obj -f win64 -d WIN64 %(FullPath)</Command>
```
Narann: Where `x64` sub directory come from ?

https://github.com/mupen64plus/mupen64plus-win32-deps/tree/master/nasm-2.14.02

Looks like it’s not here.

Narann: I just realize you download `nasm-2.14.02-win32.zip` and `nasm-2.14.02-win64.zip` and extract them in their respective directory.

Shouldn’t we add those to win32 repo ? https://github.com/mupen64plus/mupen64plus-win32-deps

Narann:
I’ve opened a PR to the win32-deps repo so we could remove the NASM downloading. :)

https://github.com/mupen64plus/mupen64plus-win32-deps/pull/8

appveyor.yml
Jj0YzL5nvJ: The `echo` command behaves differently in PowerShell than it usually does in CMD. I'm not really familiar with PS, so I simply use the syntax that is most familiar to me (Batch).

projects/msvc/mupen64plus-core.vcxproj
Jj0YzL5nvJ: Indeed, I will make the appropriate arranges once the dependencies are in place.

appveyor.yml
Narann: I'm not sure to get it. I talk about the `git clone` command. Why removing the `cmd` ?

Jj0YzL5nvJ: Because it's not necessary. The `git` command is agnostic to the CLI, it has no special behaviours in any specific shell.

Check out other AppVeyor's:
https://github.com/PCSX2/pcsx2/blob/master/appveyor.yml
https://github.com/gonetz/GLideN64/blob/master/appveyor.yml

Narann:
@Jj0YzL5nvJ It's done! Please update the appveyor file and rebase on master. Then we should be able to merge!

Jj0YzL5nvJ:
It's done, this time I kept the original logic.

Narann:
Looks OK to me.

Narann:
Good catch, I didn't think about it. I don't think it's mandatory for appveyor.

@Jj0YzL5nvJ do you have any rationals about renaming the Visual Project folder name?

Jj0YzL5nvJ:
I changed the name because now the project files are more generic, they are no longer limited to Visual Studio 2017.
 In fact, it's necessary to make similar changes in related projects to be consistent. I will do that shortly...

Jj0YzL5nvJ:
It's done.

Gillou68310:
Thanks @Jj0YzL5nvJ ;-)

Did you see my above comment on the Configuration thing?

I'm personnaly switching configurations quite often and I don't want to manually copy the core dll to mupen64plus-console.exe folder every time I recompile.

Jj0YzL5nvJ:
> Did you see my above comment on the Configuration thing?

What I understood were the references in mupen64plus-ui-console.sln

> I'm personnaly switching configurations quite often and I don't want to manually copy the core dll to mupen64plus-console.exe folder every time I recompile.

I'm currently blind (without a Windows machine), and I don't know exactly what the mupen64plus-ui-console.sln file does. Seeing the code does not make it clear to me where the results of the slave projects end.
Personally, I prefer my own scripts than being putting commands into project files... the order of someone is the disorder of someone else. [Look at this little monstrosity](https://github.com/Jj0YzL5nvJ/angrylion-rdp-plus/commit/acf96d404be8ae9d83a33ddb73736910a46ac6b1).

Gillou68310:
mupen64plus-ui-console.sln embeds all vcxproj from ui, core and other plugins
From the mupen64plus-ui-console.sln config manager you can select a specific config for every vcxproj.

They usually all only have Debug and Release config, but the core have an additional debug and release config for the new_dynarec.

If I build debug for x86 all plugin dlls and ui executable will end up in ```mupen64plus-ui-console\projects\VisualStudio2017\Win32\Debug``` directory.

With your change if I decide to build with new dynarec support, the core dll will end up in ```mupen64plus-ui-console\projects\VisualStudio2017\Win32\New_Dynarec_Debug``` directory.

Thus I need to manually copy the dll into the Debug directory.

Jj0YzL5nvJ:
Okay, it currently makes more sense to build the New_Dynarec individually.
https://github.com/mupen64plus/mupen64plus-ui-console/pull/56

PR #698, opened 2020-02-05
Don't set DPC_STATUS_START/END_VALID bits

loganmc10:
As noted here:
https://github.com/libretro/mupen64plus-libretro-nx/commit/8c87b431e0fe447d7bf2e3666623ee1e57ca80f3

DPC_STATUS_START_VALID and DPC_STATUS_END_VALID are never cleared, causing problems with some LLE RSP plugins

### DIFF
--- a/src/device/rcp/rdp/rdp_core.c
+++ b/src/device/rcp/rdp/rdp_core.c
@@ -112,10 +112,8 @@ void write_dpc_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mas
     {
     case DPC_START_REG:
         dp->dpc_regs[DPC_CURRENT_REG] = dp->dpc_regs[DPC_START_REG];
-        dp->dpc_regs[DPC_STATUS_REG] |= DPC_STATUS_START_VALID;
         break;
     case DPC_END_REG:
-        dp->dpc_regs[DPC_STATUS_REG] |= DPC_STATUS_END_VALID;
         gfx.processRDPList();
         signal_rcp_interrupt(dp->mi, MI_INTR_DP);
         break;


m4xw:
I'd suggest adding a comment that the value is (supposedly?) cleared by hw.
Do you want to wait for the merge until we confirmed the proper hw behavior for this?
I still think transaction_start/end should be the name of the bits, see my added comment, but if anyone knows for sure... let me know please.

Narann:
> I'd suggest adding a comment that the value is (supposedly?) cleared by hw.
If subtle, answerable things are confirmed by hw, I’m always in favor of one line comment for knowledge purpose.

About the flag name, I will let @bsmiles32 settles.

Gillou68310:
Capture

m4xw:
> If subtle, answerable things are confirmed by hw, I’m always in favor of one line comment for knowledge purpose.

Yea, fully agree, but well the docs that Gillou quoted are weirdly written and I am not sure if this is supposed to be set by a asynchronous task or by hw..
I've seen chips where actually reading the bit will unset it by hw and simliar weird stuff.
Don't think I get around some hw tests to rule all concerns out.
Also I want to throw a dry baguette at the person that wrote that document as well as picked the bit names..

Gillou68310:
From what I understand those flags are used to avoid a new RDP command transfer while another is still in progress. It's not very clear when those flags are going to be cleared.

Since m64p is not doing asynchronous transfer between CPU, RSP and RDP I think it's actually safer to ignore those flags.

loganmc10:
If I had to guess, the proper way to do it would be to set START_VALID at DPC_START_REG and clear it at DPC_END_REG, and do the opposite for END_VALID.

I'm going to update the PR to work that way and do some testing

m4xw:
> 
> 
> If I had to guess, the proper way to do it would be to set START_VALID at DPC_START_REG and clear it at DPC_END_REG, and do the opposite for END_VALID.
> 
> I'm going to update the PR to work that way and do some testing

Yea I thought so too at first, however I didn't end up trying it because from our discussion it was spinning so i wouldnt know how the emu will write that flag, if it isnt handled via a interrupt.

loganmc10:
@m4xw can you share what code in the rsp plugin is hanging? That will help me understand what the RSP plugin expects these bits to be

m4xw:
@loganmc10 am at work rn, can do so in 5-6h

Gillou68310:
If I should guess I would say START_VALID=1 and END_VALID=1 until CMD_CURRENT = CMD_END

Here's a use case:
```
#ifdef OUTPUT_DRAM
	# reset output pointers
		lw	outp, OS_TASK_OFF_OUTBUFF($1)
		lw	$3, OS_TASK_OFF_OUTBUFF_SZ($1)
		sw	outp, RSP_STATE_DRAM_OUTP(rsp_state)
		sw	$3, RSP_STATE_DRAM_OUT_LENP(rsp_state)

		addi	$4, zero, DPC_CLR_XBUS_DMEM_DMA
		mtc0	$4, CMD_STATUS
 		addi	outp, zero, RSP_OUTPUT_OFFSET

#else
#ifdef OUTPUT_FIFO
.name	bufend, $3
	# reset output pointers
		lw	outp, OS_TASK_OFF_OUTBUFF($1)
		lw	bufend, OS_TASK_OFF_OUTBUFF_SZ($1)
		sw	outp, RSP_STATE_FIFO_BUF_TOP(rsp_state)
		sw	bufend, RSP_STATE_FIFO_BUF_END(rsp_state)

		mfc0	$4, CMD_STATUS
		andi	$4, $4, DPC_STATUS_XBUS_DMEM_DMA
		bne	$4, zero, restart_fifo
		mfc0	$4, CMD_END
		sub	outp, outp, $4
		bgtz	outp, restart_fifo

	mfc0	$5, CMD_CURRENT
	beq	$5, zero, restart_fifo
	nop
	beq	$5, $4, restart_fifo
	

		nop
		j	set_fifo
		ori	bufend, $4, 0

restart_fifo:
		mfc0	$4, CMD_STATUS
		andi	$4, $4, DPC_STATUS_START_VALID
		bne	$4, zero, restart_fifo
		addi	$4, zero, DPC_CLR_XBUS_DMEM_DMA
 ### LOOP TO restart_fifo: UNTIL START_VALID=0
		mtc0	$4, CMD_STATUS
		mtc0	bufend, CMD_START
		mtc0	bufend, CMD_END
set_fifo:
		sw	bufend, RSP_STATE_FIFO_OUTP(rsp_state)
 		addi	outp, zero, RSP_OUTPUT_OFFSET
.unname	bufend
```

Gillou68310:
BTW @loganmc10 is this supposed to fix any issue?

loganmc10:
@Gillou68310 no, not that I know of, it was just an attempt at improving accuracy. Given that we don't do asynchronous transfers like you say, it probably is safer to just leave the bits at 0

Gillou68310:
I think it's best indeed

loganmc10:
Ok I've updated the PR to just stop setting these bits, I agree with @Gillou68310 that it's the best option in this case

m4xw:
Sounds sufficient for now, but ideally this should be implemented if hw behavior gets documented.
I'd at least leave it commented for now.

loganmc10:
In real hardware the RDP and CPU run asynchronously, so it's probably not something that can be properly emulated in mupen64plus

loganmc10:
Actually, now that I think about it, clearing both bits inside rdp_interrupt_event() may be the proper thing to do, but I think the current PR is good, just leaving them at 0 doesn't seem to cause any issues

PR #687, opened 2019-11-17
Database updates

Mastergatto:
By comparing with No-Intro database (see #599 ), I have corrected some errors according to No-Intro, because IMHO No-Intro is more trustworthy than GoodSet. I have also added new iQue games and manuals, some previously undumped revisions and some new prototypes.  

Advices are welcome.

PR #700, opened 2020-02-09
Whoops

Jj0YzL5nvJ:
https://github.com/mupen64plus/mupen64plus-core/pull/687#issuecomment-583770386


Narann:
@fzurita Is this good for you?

fzurita:
That's good with me, it doesn't repeat any more.

Issue #689, opened 2019-11-26
Aero Fighters Assault - Startup Error


StenApp:
Seems to be a mupen64plus core issue...

#https://github.com/gonetz/GLideN64/issues/2069

Tasosgemah:
The older mupen64 core doesn't have this issue, don't know if that's useful information.

loganmc10:
I think this is fixed now @StenApp can you confirm?

StenApp:
@PixelCrusader please check if it works now

> When booting the game, the expected behavior is for the company logos and licensing information to fade from black, before transitioning into white and then into the planes flying.
> 
> ![AeroFighters Assault (USA)-190711-021651](https://user-images.githubusercontent.com/40935002/61031281-323bb200-a38d-11e9-8d47-18b19fbd9e2a.png)
> 
> Currently, booting the game almost completely cuts the company logos and licensing, prematurely changing into the planes flying.
> 
> This behavior was observed on Mupen64Plus-gui using the GLideN64 Public Release 4.0 as well as RetroArch's latest version of Mupen64Plus-Next, which uses a port of GLideN. Real hardware was used to confirm the discrepancy.
> 
> The example picture above was taken from RetroArch running an older version of their mupen64plus core (af7a4bf). This could mean that this is a regression.
> 
> Video of how the game should look:
> 
> https://www.youtube.com/watch?v=0ceyDL-bSR0&t=09s
> 
> https://www.youtube.com/watch?v=MhYAA8niTQM
> 
> https://www.youtube.com/watch?v=2C-ulZRmTT8

PixelCrusader:
RetroArch 1.8.4, Mupen 64 Plus-Next Open GL (2.0 81a58df)

The problem persists, at least on my end. Logos begin to fade in, smash cut to planes.

StenApp:
Could you build a Mupen64plus version? I don't know if the Mupen64plus-next core has fixed that issue...

Jj0YzL5nvJ:
[Mupen64Plus-Next](https://github.com/libretro/mupen64plus-libretro-nx) is a completely different emulator and it's so heavily modified that save states are incompatible. The default core (master) has not been synchronized since 2019/03/03.
The most updated branch (rebase-core-gliden) was last synchronized on 2020/03/01 but you must compile it manually, and who knows if it works.

[...]

PR #701, opened 2020-02-10
Remove special_int (DO NOT MERGE)

loganmc10:
There was some discussion about removing SPECIAL_INT in this PR:

https://github.com/mupen64plus/mupen64plus-core/pull/444

The most reasonable solution proposed (in my mind) is to make the counter a 64 bit register, and use the lower 32 bits for COUNT_REG.

I have tested this PR with both interpreters.

Left to be done:

- Dynarecs -> I'm really not great at assembly, I'm hoping someone else can take on this task

I tested a number of games, including ones that I know are sensitive to timing and interrupt changes, I didn't notice any regressions. In fact, this PR fixes https://github.com/mupen64plus/mupen64plus-core/issues/689

I think this change is worthwhile, it simplifies the interrupt code, and hopefully might remove any bugs that the SPECIAL_INT "hack" (for lack of a better term) might have caused

### DIFF
--- a/src/device/r4300/interrupt.c
+++ b/src/device/r4300/interrupt.c
@@ -98,38 +98,6 @@ static void clear_queue(struct interrupt_queue* q)
     clear_pool(&q->pool);
 }
 
-static int before_event(const struct cp0* cp0, unsigned int evt1, unsigned int evt2, int type2)
-{
-    const uint32_t* cp0_regs = r4300_cp0_regs((struct cp0*)cp0); /* OK to cast away const qualifier */
-    uint32_t count = cp0_regs[CP0_COUNT_REG];
-
-    if (evt1 - count < UINT32_C(0x80000000))
-    {
-        if (evt2 - count < UINT32_C(0x80000000))
-        {
-            if ((evt1 - count) < (evt2 - count)) return 1;
-            else return 0;
-        }
-        else
-        {
-            if ((count - evt2) < UINT32_C(0x10000000))
-            {
-                switch(type2)
-                {
-                    case SPECIAL_INT:
-                        if (cp0->special_done) return 1;
-                        else return 0;
-                        break;
-                    default:
-                        return 0;
-                }
-            }
-            else return 1;
-        }
-    }
-    else return 0;
-}
-
 unsigned int add_random_interrupt_time(struct r4300_core* r4300)
 {
     if (r4300->randomize_interrupt) {
@@ -146,23 +114,15 @@ unsigned int add_random_interrupt_time(struct r4300_core* r4300)
 
 void add_interrupt_event(struct cp0* cp0, int type, unsigned int delay)
 {
-    const uint32_t* cp0_regs = r4300_cp0_regs(cp0);
-    add_interrupt_event_count(cp0, type, cp0_regs[CP0_COUNT_REG] + delay);
+    const uint64_t* cp0_count = r4300_cp0_get_count(cp0);
+    add_interrupt_event_count(cp0, type, *cp0_count + delay);
 }
 
-void add_interrupt_event_count(struct cp0* cp0, int type, unsigned int count)
+void add_interrupt_event_count(struct cp0* cp0, int type, uint64_t count)
 {
     struct node* event;
     struct node* e;
-    int special;
-    const uint32_t* cp0_regs = r4300_cp0_regs(cp0);
-    unsigned int* cp0_next_interrupt = r4300_cp0_next_interrupt(cp0);
-
-    special = (type == SPECIAL_INT);
-
-    if (cp0_regs[CP0_COUNT_REG] > UINT32_C(0x80000000)) {
-        cp0->special_done = 0;
-    }
+    uint64_t* cp0_next_interrupt = r4300_cp0_next_interrupt(cp0);
 
     if (get_event(&cp0->q, type)) {
         DebugMessage(M64MSG_WARNING, "two events of type 0x%x in interrupt queue", type);
@@ -184,7 +144,7 @@ void add_interrupt_event_count(struct cp0* cp0, int type, unsigned int count)
         event->next = NULL;
         *cp0_next_interrupt = cp0->q.first->data.count;
     }
-    else if (before_event(cp0, count, cp0->q.first->data.count, cp0->q.first->data.type) && !special)
+    else if (count < cp0->q.first->data.count)
     {
         event->next = cp0->q.first;
         cp0->q.first = event;
@@ -194,7 +154,7 @@ void add_interrupt_event_count(struct cp0* cp0, int type, unsigned int count)
     {
         for (e = cp0->q.first;
             e->next != NULL &&
-            (!before_event(cp0, count, e->next->data.count, e->next->data.type) || special);
+            (!(count < e->next->data.count));
             e = e->next);
 
         if (e->next == NULL)
@@ -204,8 +164,7 @@ void add_interrupt_event_count(struct cp0* cp0, int type, unsigned int count)
         }
         else
         {
-            if (!special)
-                for(; e->next != NULL && e->next->data.count == count; e = e->next);
+            for(; e->next != NULL && e->next->data.count == count; e = e->next);
 
             event->next = e->next;
             e->next = event;
@@ -216,22 +175,20 @@ void add_interrupt_event_count(struct cp0* cp0, int type, unsigned int count)
 void remove_interrupt_event(struct cp0* cp0)
 {
     struct node* e;
-    const uint32_t* cp0_regs = r4300_cp0_regs(cp0);
-    uint32_t count = cp0_regs[CP0_COUNT_REG];
-    unsigned int* cp0_next_interrupt = r4300_cp0_next_interrupt(cp0);
+    const uint64_t* cp0_count = r4300_cp0_get_count(cp0);
+    uint64_t* cp0_next_interrupt = r4300_cp0_next_interrupt(cp0);
 
     e = cp0->q.first;
     cp0->q.first = e->next;
     free_node(&cp0->q.pool, e);
 
     *cp0_next_interrupt = (cp0->q.first != NULL
-         && (cp0->q.first->data.count > count
-         || (count - cp0->q.first->data.count) < UINT32_C(0x80000000)))
+         && (cp0->q.first->data.count > *cp0_count))
         ? cp0->q.first->data.count
         : 0;
 }
 
-unsigned int get_event(const struct interrupt_queue* q, int type)
+uint64_t get_event(const struct interrupt_queue* q, int type)
 {
     const struct node* e = q->first;
 
@@ -287,17 +244,20 @@ void remove_event(struct interrupt_queue* q, int type)
 void translate_event_queue(struct cp0* cp0, unsigned int base)
 {
     struct node* e;
-    const uint32_t* cp0_regs = r4300_cp0_regs(cp0);
+    uint32_t* cp0_regs = r4300_cp0_regs(cp0);
+    uint64_t* cp0_count = r4300_cp0_get_count(cp0);
 
     remove_event(&cp0->q, COMPARE_INT);
-    remove_event(&cp0->q, SPECIAL_INT);
 
     for (e = cp0->q.first; e != NULL; e = e->next)
     {
-        e->data.count = (e->data.count - cp0_regs[CP0_COUNT_REG]) + base;
+        e->data.count = (e->data.count - *cp0_count) + base;
     }
-    add_interrupt_event_count(cp0, COMPARE_INT, cp0_regs[CP0_COMPARE_REG]);
-    add_interrupt_event_count(cp0, SPECIAL_INT, 0);
+
+    uint32_t delay = cp0_regs[CP0_COMPARE_REG] - (base + cp0->count_per_op);
+    *cp0_count = base + cp0->count_per_op;
+    add_interrupt_event(cp0, COMPARE_INT, delay);
+    *cp0_count -= cp0->count_per_op;
 }
 
 int save_eventqueue_infos(const struct cp0* cp0, char *buf)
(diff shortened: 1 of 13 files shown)


loganmc10:
Actually a bug I just noticed: the DK intro freezes not long into the song, I'll look into it later in the week

loganmc10:
@Gillou68310 @bsmiles32 maybe you can answer this for me, is CP0_COUNT_REG a signed int? I always assumed it was unsigned and rolled over at 0xFFFFFFFF, but looking at the code, and doing some testing, it seems that it might be a signed int, that is expected to roll over at 0x80000000, is that right?

bsmiles32:
cp0_count is unsigned and roll over when reaching 2^32-1.

loganmc10:
@bsmiles32 are you 100% sure though, look at this:

image

And the current interrupt handler has code like this:
```
if (cp0_regs[CP0_COUNT_REG] > UINT32_C(0x80000000)) {
    cp0->special_done = 0;
}
```

loganmc10:
Actually I guess "sign extended" is supposed to mean that it preserves the sign while still using all 32 bits somehow?

The reason I suspect this, is that DK64 is freezing for me right when COUNT_REG (and COMPARE_REG) get larger than 0x80000000

bsmiles32:
6.3.3 Count Register (9)

The read/write Count register acts as a timer, incrementing at a constant rate—half 
the PClock speed—whether or not instructions are being executed. This register 
is a free-running type. When the register reaches all ones, it rolls over to zero and 
continues counting. This register can be used for diagnostic purposes, system 
initialization or synchronization between the processes.

From User’s Manual U10504EJ7V0UM00 p164

loganmc10:
I feel like there is something missing, even in PJ64 I see this:

https://github.com/project64/project64/blob/f20dca31fdc43e71fd50c4c0888352e4c78a36c2/Source/Project64-core/N64System/Mips/SystemTiming.cpp#L270-L282

They don't let "NextCompare" get any larger than 2^31-1

loganmc10:
@bsmiles32 I pushed the change to wrap CP0_COUNT_REG around at 2^31-1 in a separate commit:

https://github.com/mupen64plus/mupen64plus-core/pull/701/commits/d266a5d3b27b6ef9a8e1644c467926d58574e1ad

This does fix the freeze in DK64 (Indiana Jones also froze when starting a new game without this change).

I'm really not too sure on this, it would maybe help if someone had some way to test on a real N64 what happens when cp0_count exceeds 2^31-1?

m4xw:
> I'm really not too sure on this, it would maybe help if someone had some way to test on a real N64 what happens when cp0_count exceeds 2^31-1?


This should be 2^32-1, shouldnt it?
> When the register reaches all ones

As per bsmiles quote.
Pretty sure this should be treated as uint32 since according to some datasheets it uses the full 32bits for counting.
I would consider the possibility that this might trigger another bug if implemented properly, maybe register endianess breaking something when bitcounting reaches a certain pattern or stuff like that.

loganmc10:
I'm not really sure, the way this PR is written currently, I don't see any bugs, I've tested quite a few games.

It would be a really easy thing for somebody that had experience writing test ROMs could check, just  keep reading CP0_COUNT_REG and see what it does when it exceeds 0x7fffffff. The other thing to test would be to write a value greater than 0x7fffffff to CP0_COMPARE_REG, and then read it back, and see what the value is

Gillou68310:
Capture

It's actually clear from the docs that the count register roll over when reaching 2^32-1.
I'll try to take time this afternoon to review your 1st commit on this PR.

loganmc10:
Thanks @Gillou68310 . I've read the same document and I totally agree about what it says. With the first commit, DK64 freezes when the count register exceeds 0x80000000. Everything seems to work perfectly when I roll the counter over at that point

Gillou68310:
I started looking at the code and I'm not sure the COMPARE_INT event is generated correctly


I would do something like that for the handler:
````
void compare_int_handler(void* opaque)
{
    struct r4300_core* r4300 = (struct r4300_core*)opaque;
    uint32_t* cp0_regs = r4300_cp0_regs(&r4300->cp0);
    uint64_t* cp0_count = r4300_cp0_get_count(&r4300->cp0);
	
    uint32_t delay = cp0_regs[CP0_COMPARE_REG] - (cp0_regs[CP0_COUNT_REG] + r4300->cp0.count_per_op);
    *cp0_count += r4300->cp0.count_per_op;
    add_interrupt_event(&r4300->cp0, COMPARE_INT, delay);
    *cp0_count -= r4300->cp0.count_per_op;

    raise_maskable_interrupt(r4300, CP0_CAUSE_IP7);
}
```

translate_event_queue and DECLARE_INSTRUCTION(MTC0) needs to be adapted too

Gillou68310:
Adding r4300->cp0.count_per_op to the count is necessary in the case cp0_regs[CP0_COMPARE_REG] == cp0_regs[CP0_COUNT_REG] to have a delay of 0x100000000 instead of 0

Gillou68310:
translate_event_queue would be:

```
void translate_event_queue(struct r4300_core* r4300, unsigned int base)
{
    struct node* e;
    struct cp0* cp0 = &r4300->cp0;
    uint32_t* cp0_regs = r4300_cp0_regs(cp0);
    uint64_t* cp0_count = r4300_cp0_get_count(cp0);

    remove_event(&cp0->q, COMPARE_INT);

    for (e = cp0->q.first; e != NULL; e = e->next)
    {
        e->data.count = (e->data.count - *cp0_count) + base;
    }

    uint32_t delay = cp0_regs[CP0_COMPARE_REG] - (base + cp0->count_per_op);
    *cp0_count = base + cp0->count_per_op;
    add_interrupt_event(cp0, COMPARE_INT, delay);
    *cp0_count -= cp0->count_per_op;
}
```

I removed the VI thing, I think this should be in another PR

Gillou68310:
And MTC0

```
    case CP0_COMPARE_REG:
        cp0_update_count(r4300);
        remove_event(&r4300->cp0.q, COMPARE_INT);
        uint32_t delay = rrt32 - (cp0_regs[CP0_COUNT_REG] + r4300->cp0.count_per_op);
        *cp0_count += r4300->cp0.count_per_op;
        add_interrupt_event(&r4300->cp0, COMPARE_INT, delay);
        *cp0_count -= r4300->cp0.count_per_op;
        cp0_regs[CP0_COMPARE_REG] = rrt32;
        cp0_regs[CP0_CAUSE_REG] &= ~CP0_CAUSE_IP7;
        break;
```

loganmc10:
```
uint32_t delay = cp0_regs[CP0_COMPARE_REG] - (cp0_regs[CP0_COUNT_REG] + r4300->cp0.count_per_op
```
I don't think this is right, picture a scenario where COMPARE_REG == 0. We just interrupted, so COUNT_REG is also 0. The result of that equation would be -2

Gillou68310:
unsigned so 0xFFFFFFFE actually

Gillou68310:
scheduling the comparing interrupt at 0x100000000

Gillou68310:
Regarding the VI thing I think the next_vi variable should just be removed all together. As far as I know we can get the next VI event at any time using the get_event function.

loganmc10:
Ok makes sense, I'll try it in a bit, but then why do:
```
cp0_count += r4300->cp0.count_per_op;
cp0_count -= r4300->cp0.count_per_op;
```
As well, isn't that doing the same thing?

loganmc10:
I think the next_vi variable existed to provide precision timing. The VI_INT handler could be called a few cycles after next_vi, so if you just add vi_delay and schedule it again in the handler, the distance between the 2 interrupts would actually be slightly greater than next_vi. That was my understanding anyway

Gillou68310:
it's to revert the effect of adding count_per_op to ```uint32_t delay = rrt32 - (cp0_regs[CP0_COUNT_REG] + r4300->cp0.count_per_op);```

Gillou68310:
> I think the next_vi variable existed to provide precision timing. The VI_INT handler could be called a few cycles after next_vi, so if you just add vi_delay and schedule it again in the handler, the distance between the 2 interrupts would actually be slightly greater than next_vi. That was my understanding anyway

Ok let's discuss that later in another PR, I didn't look in detail rn

Gillou68310:
@loganmc10 Had time to test what we talked about?

loganmc10:
Yes, sorry it's early here. I tried out your changes and it does work, for some reason I can't wrap my head around what the difference is, but I'll update the PR with the new logic and wrapping around at 2^32-1 and do some more testing

Gillou68310:
Sorry no pressure here, I was just wondering if my logic was correct ;-)
Nice to hear it works!

loganmc10:
Ok I've updated the PR with the new logic, I left the next_vi stuff in there for now, I can remove it I suppose, but it's needed to make translate_event_queue work properly.

Everything seems to be working normally. One minor thing that actually seems fixed in this PR, in Twisted Edge snowboarding, with the current master, about halfway through the race I usually got an earful of static for about a second. That problem seems to be resolved in this PR.

I've got savestates working as well, I think I've written it in a way that still allows save states to be forward and backward compatible.

Gillou68310:
I don't have time to look at the savestate thing rn, I'll do that tomorrow!
Thanks for your work @loganmc10

src/device/r4300/interrupt.c
```diff
@@ -273,13 +273,16 @@ int save_eventqueue_infos(const struct cp0* cp0, char *buf)
...
+    unsigned int short_count;
 
     len = 0;
 
     for (e = cp0->q.first; e != NULL; e = e->next)
     {
         memcpy(buf + len    , &e->data.type , 4);
         memcpy(buf + len + 4, &e->data.count, 4);
+        short_count = e->data.count & UINT64_C(0xFFFFFFFF);
+        memcpy(buf + len + 4, &short_count, 4);
```
Gillou68310: This is not necessary ```memcpy(buf + len + 4, &e->data.count, 4);``` is already doing it

src/device/r4300/interrupt.c
```diff
@@ -290,13 +293,15 @@ int save_eventqueue_infos(const struct cp0* cp0, char *buf)
...
 {
     int len = 0;
+    uint64_t* cp0_count = r4300_cp0_get_count(cp0);
 
     clear_queue(&cp0->q);
 
     while (*((const unsigned int*)&buf[len]) != 0xFFFFFFFF)
     {
         int type = *((const unsigned int*)&buf[len]);
-        unsigned int count = *((const unsigned int*)&buf[len+4]);
```
Gillou68310: type == SPECIAL_INT shouldn't not be added to the event list

Gillou68310: We should probably also need to re-schedule the COMPARE_INT here

src/main/savestates.c
```diff
@@ -476,6 +477,7 @@ static int savestates_load_m64p(struct device* dev, char *filepath)
 
     *r4300_cp0_next_interrupt(&dev->r4300.cp0) = GETDATA(curr, uint32_t);
     dev->vi.next_vi = GETDATA(curr, uint32_t);
+    if (dev->vi.next_vi <= cp0_regs[CP0_COUNT_REG]) dev->vi.next_vi += UINT64_C(0x100000000);
```
Gillou68310: this won't be necessary if https://github.com/mupen64plus/mupen64plus-core/pull/702 is merged

Gillou68310:
Storing/loading next_interrupt in savestate is actually not necessary since it's overwritten by load_eventqueue_infos. It makes even less sense now since having only the low 32bits part doesn't mean anything.

Gillou68310:
```PUTDATA(curr, uint32_t, get_event(&dev->r4300.cp0.q, VI_INT) - cp0_regs[CP0_COUNT_REG]); // vi_timer```

needs to be updated with 64bits count in savestates_save_pj64

File: src/device/rcp/vi/vi_controller.c [hunk 4]
```diff
@@ -89,15 +89,15 @@ void read_vi_regs(void* opaque, uint32_t address, uint32_t* value)
...
+    const uint64_t* cp0_count = r4300_cp0_get_count(&vi->mi->r4300->cp0);
 
     if (reg == VI_CURRENT_REG)
     {
         /* XXX: update current line number */
         cp0_update_count(vi->mi->r4300);
 
         if (vi->regs[VI_V_SYNC_REG] != 0)
-            vi->regs[VI_CURRENT_REG] = (vi->delay - (vi->next_vi - cp0_regs[CP0_COUNT_REG])) / vi->count_per_scanline;
+            vi->regs[VI_CURRENT_REG] = (vi->delay - (vi->next_vi - *cp0_count)) / vi->count_per_scanline;
```
Gillou68310: cast to uint32_t to avoid warning

Gillou68310:
BTW we need to increment version since newer savestates won't work on older core.

src/device/r4300/interrupt.c
loganmc10: I'm not sure what you mean by re-schedule COMPARE_INT, won't it be in the event queue?

loganmc10:
@Gillou68310 ok I've rebased the PR and incorporated your feedback. The one question I had was:

> We should probably also need to re-schedule the COMPARE_INT here

I don't quite understand that, the event queue should have a pending COMPARE_INT inside when it is loaded from the save state

loganmc10:
Oh and I still need to fix:

> PUTDATA(curr, uint32_t, get_event(&dev->r4300.cp0.q, VI_INT) - cp0_regs[CP0_COUNT_REG]); // vi_timer

> needs to be updated with 64bits count in savestates_save_pj64

Although IMO, we should just get rid of savestates_save_pj64, I doubt it works in 80% of cases. Loading PJ64 save states hardly works from the ones I've tested

Gillou68310:
@loganmc10 I was just thinking out loud, I wasn't sure the COMPARE_INT event from older savestate are going to be 100% compatible with the new core

loganmc10:
Oh ok, it should be fine, I tested loading a few old save states. It shouldn't need any special reordering in a load, it operates the same as the other interrupts

loganmc10:
closed in favour of #708

Issue #554, opened 2018-03-31
VI_CURRENT_REG not calculated properly if CP0_COUNT_REG changed by software


pdrome:
I'm building an N64 test ROM where I occasionally reset CP0_COUNT_REG to zero. This register is writable and such a scenario is possible on real hardware. After doing this, mupen will then take a very long period of time before it reaches the next VI.

I believe this is because the VI_CURRENT_REG isn't properly calculated when I change the CP0_COUNT_REG value.

https://github.com/mupen64plus/mupen64plus-core/blob/5fb51e8d268a1a666244506d97574665d56901c2/src/device/rcp/vi/vi_controller.c#L90

I'm not sure if this behavior affects any games, but if anything out there is modifying CP0_COUNT_REG it could cause timing issues.  Here is a test ROM that resets CP0_COUNT_REG to zero and demonstrates the issue:

https://github.com/pdrome/n64-asm-timing/blob/master/TIMING_R0/CPUTIMING_COUNT.N64

Here is the same test ROM but without the lines which reset CP0_COUNT_REG:

https://github.com/pdrome/n64-asm-timing/blob/master/TIMING_R0/CPUTIMING_COUNT_NO_MTC0.N64

Gillou68310:
Thanks for testing! Could you test the following branch to see if it makes any difference?
https://github.com/Gillou68310/mupen64plus-core/tree/int
It's a wip refactoring to get rid of the special interrupt used to deal with the count register overflow.

pdrome:
I'm out and about today, but I can test this evening

If you want to test in the meantime, the main way to see the problem will be that the first ROM (CPUTIMING_COUNT.N64) will run slowly and the second ROM (CPUTIMING_COUNT_NO_MTC0.N64) will run significantly faster. This is for CountPerOp>=2.

Additionally,, if you set CountPerOp==1 then the first ROM will show blank screen before everything pops in at once while the second ROM will incrementally fill the screen (like with CountPerOp>=2).

pdrome:
@Gillou68310 

I tried the WIP code and it does not appear to affect the issue.

pdrome:
Check that last comment.  I realized today that I didn't have the latest commit from the WIP version.  git clone gave me an old version so I had to download the zip instead.

[...]

PR #702, opened 2020-02-14
Remove next_vi

Gillou68310:
Should fix https://github.com/mupen64plus/mupen64plus-core/issues/554

### DIFF
--- a/src/device/r4300/interrupt.c
+++ b/src/device/r4300/interrupt.c
@@ -213,7 +213,7 @@ void add_interrupt_event_count(struct cp0* cp0, int type, unsigned int count)
     }
 }
 
-static void remove_interrupt_event(struct cp0* cp0)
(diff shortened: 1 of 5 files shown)


loganmc10:
I still don't really see the benefit. It doesn't really simplify the code, and:
```
uint32_t next_vi = get_event(&vi->mi->r4300->cp0.q, VI_INT) + vi->delay;
```
Will need to be reworked when get_event returns a 64 bit value (probably not a big deal)

It's just sort of a different way of doing the same thing

Gillou68310:
Sure it is, but cleaner IMO. And yes it simplify the code because you don't need to update next_vi all over the place.

loganmc10:
OK, I doubled check that there were no regressions by diff'ing the interrupt timings between the current master and this PR. No issues, all interrupts are timed the same, so this is OK to be merged by me.

I'll update my PR once this is merged

Issue #600, opened 2019-01-03
Command and conquer in-game save corruption


fzurita:
It seems that Mupen64plus is unable to create in-game saves for command and conquer. Whenever this is attempted, the following error occurs within the game:

![command conquer-000](https://user-images.githubusercontent.com/11581687/50637542-6048c100-0f28-11e9-85ef-4dfcc2ff842f.png)

This may just be a game database issue. The wrong save type may be specified.

fzurita:
Save type doesn't seem to match the project 64 rdb, but fixing it doesn't allow the game to save.

loganmc10:
I spent some time looking into this. At least now, mupen64plus and PJ64 save game formats match (flash ram), maybe the .ini has been updated since this report.

It still doesn't work, it does work in PJ64. Theoretically should be easy enough to fix? Just need to compare the implementation in PJ64 to mupen64plus, there are definite differences

fzurita:
This is probably fixed.

PR #705, opened 2020-02-16
Only align cart address in PI DMA

loganmc10:
This PR fixes:

https://github.com/mupen64plus/mupen64plus-core/issues/600

Cart and DRAM addresses were aligned to fix an issue in Taz Express:

https://github.com/mupen64plus/mupen64plus-core/pull/541

However, this broke saving in Command & Conquer. I tried a couple things, like just aligned PI DMA Writes, Command & Conquer wouldn't boot at all then.

The thing that makes both games (Taz and C&C) work is just aligning the cart addresses.

I'll admit that I don't really understand why, the documentation says:

> We summarize restrictions on DMA transfers as follows:
> 1) RDRAM address must be 8-byte aligned---the three least significant bits of address
must be zeroes.
> 2) BSD address must be 2-byte aligned---the least significant bit of address must be zero.
> 3) Length of DMA must be a multiple of 8 bytes (specified as the byte count - 1)---the
three least significant bits of the length register should always be ones.
> 4) Maximum transfer size is 16 MByte.

### DIFF
--- a/src/device/rcp/pi/pi_controller.c
+++ b/src/device/rcp/pi/pi_controller.c
@@ -54,7 +54,7 @@ enum
 static void dma_pi_read(struct pi_controller* pi)
 {
     uint32_t cart_addr = pi->regs[PI_CART_ADDR_REG] & ~UINT32_C(1);
-    uint32_t dram_addr = pi->regs[PI_DRAM_ADDR_REG] & ~UINT32_C(7);
+    uint32_t dram_addr = pi->regs[PI_DRAM_ADDR_REG];
     uint32_t length = (pi->regs[PI_RD_LEN_REG] & UINT32_C(0x00fffffe)) + 2;
     const uint8_t* dram = (uint8_t*)pi->ri->rdram->dram;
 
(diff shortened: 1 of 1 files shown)


Gillou68310:
Well from the docs RDRAM address must be 8-byte aligned, not sure what to say here...

loganmc10:
Yes this is where hardware tests are useful, the docs say that, but it doesn't define what the behavior is when you violate that instruction.

For what it's worth, when cen64 implemented this fix, they only aligned the cart address, not the DRAM address:

https://github.com/n64dev/cen64/pull/123

m4xw:
The whole forcing alignments on the side of hw seems problematic for me.
These are limitations imposed by the silicone, so shouldn't the requests be dropped instead of forcing the alignment?
Usually if u do unaligned dma where it's not supported these chips just hang or set a error bit.
Would be weird if real hw would align it down..

loganmc10:
@m4xw the N64 does that, in a lot of the docs you'll read things like "the least 3 significant bits are assumed to be 0", so it aligns the DMAs by basically not caring what you write into certain bits.

The CEN64 folks did some actual hardware tests as well to verify the cart address alignment behavior: https://github.com/n64dev/cen64/issues/86

m4xw:
What the actual fuck.
They should state that then, I don't wanna know how many cycles were wasted back in the day just masking lower bits..
Some engineer is screaming in his grave rn at the guy that wrote the technical documentation..

Issue #609, opened 2019-01-21
WCW Backstage Assault locks up


fzurita:
WCW Backsage Assault locks up after getting past the start screen.

It locks up with interpreter or dynarec.
It also locks up with either LLE or HLE plugins.

It feels a lot like some kind of timing issue.

loganmc10:
In case it is helpful to someone, this game didn't boot at all until this change:

https://github.com/mupen64plus/mupen64plus-core/pull/324

That allowed it to boot, but I guess it doesn't get far.

loganmc10:
Probably related to this issue https://github.com/project64/project64/issues/1634

loganmc10:
I've traced this problem to it's source, it happens because of write_cart_rom:

https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/cart/cart_rom.c#L70-L75

Commenting out the code inside that function makes the game work.

There was some discussion a few years back about this function, and someone actually did some hardware tests:

https://github.com/mupen64plus/mupen64plus-core/pull/511#issuecomment-443264685

I'll spend some time this week trying to get a proper implementation going

PR #707, opened 2020-02-17
Implement cart ROM writes as described by Jan Goldacker

loganmc10:
Fixes https://github.com/mupen64plus/mupen64plus-core/issues/609

@jago85 did some investigation on how cart ROM writes are handled a while ago (https://github.com/mupen64plus/mupen64plus-core/pull/511#issuecomment-443264685)

The basic gist I got from that is that our current "rom_written" code should only happen if the game tries to read back the value when PI_STATUS_IO_BUSY is set, otherwise it should just read the regular value.

It does seem to work that way, this change fixes WCW Backstage Assault, and other games that depend on this code (A Bug's Life, Toy Story 2) still work. The only oddity with A Bug's Life/Toy Story 2 is that you get a few ```Core Warning: two events of type 0x10 in interrupt queue``` messages because there are multiple PI_INT events in the queue at the same time.

A big thanks to @jago85, there is no way we would have figured this out without those actual hardware tests.


loganmc10:
I'm not sure why AppVeyor isn't happy, but it doesn't seem related to this change. I've tested all the games that I know of that attempt to write to the cartridge ROM, I haven't found any regressions

Jj0YzL5nvJ:
Yeah, AppVeyor has been having very weird issues lately.
All configurations are being compiled properly on my machine.

loganmc10:
Hi all, any thoughts on this?

Jj0YzL5nvJ:
See #729

loganmc10:
Sorry I meant specifically related to this PR, I trust the AppVeyor issues will get sorted out

bsmiles32:
Also tested said games and found no obvious regression.
I still think with some refactoring, we could move this IO_BUSY logic inside the PI instead of spreading it in cart/cart_rom. It would also probably best reflect how the hardware works. But for now it is better than what we have and fixes real issues. So merging this work.

Thanks to all involved especially to @jago85 for the hardware investigation.

PR #708, opened 2020-02-17
Rework interrupt scheduling

Gillou68310:
@loganmc10 
I didn't realized it at first but https://github.com/mupen64plus/mupen64plus-core/pull/701 is breaking the new dynarec very badly.

The main limitation is that the number of cycles between current cycle count and next interrupt event cannot exceed 2^31

I reworked interrupt scheduling a bit:

- cp0_next_interrupt is now a negative number representing the
number of cycles left until next interrupt is taken.
Next interrupt is taken whether cp0_next_interrupt value is positive or null

- Special interrupt is now a fake interrupt which porpose is
to ensure the number of cycles between current cycle count
and next interrupt will never exceed 2^31

I still need to fix both dynarecs.
In the meantime this PR needs to be tested to see if there's no regression

### DIFF
--- a/src/device/r4300/cached_interp.c
+++ b/src/device/r4300/cached_interp.c
@@ -94,7 +94,7 @@ void cached_interp_##name(void) \
         cp0_update_count(r4300); \
     } \
     r4300->cp0.last_addr = *r4300_pc(r4300); \
-    if (*r4300_cp0_next_interrupt(&r4300->cp0) <= r4300_cp0_regs(&r4300->cp0)[CP0_COUNT_REG]) gen_interrupt(r4300); \
+    if (*r4300_cp0_cycle_count(&r4300->cp0) >= 0) gen_interrupt(r4300); \
 } \
  \
 void cached_interp_##name##_OUT(void) \
@@ -127,24 +127,26 @@ void cached_interp_##name##_OUT(void) \
         cp0_update_count(r4300); \
     } \
     r4300->cp0.last_addr = *r4300_pc(r4300); \
-    if (*r4300_cp0_next_interrupt(&r4300->cp0) <= r4300_cp0_regs(&r4300->cp0)[CP0_COUNT_REG]) gen_interrupt(r4300); \
+    if (*r4300_cp0_cycle_count(&r4300->cp0) >= 0) gen_interrupt(r4300); \
 } \
   \
 void cached_interp_##name##_IDLE(void) \
 { \
     DECLARE_R4300 \
     uint32_t* cp0_regs = r4300_cp0_regs(&r4300->cp0); \
+    int* cp0_cycle_count = r4300_cp0_cycle_count(&r4300->cp0); \
     const int take_jump = (condition); \
-    int skip; \
     if (cop1 && check_cop1_unusable(r4300)) return; \
     if (take_jump) \
     { \
         cp0_update_count(r4300); \
-        skip = *r4300_cp0_next_interrupt(&r4300->cp0) - cp0_regs[CP0_COUNT_REG]; \
-        if (skip > 3) cp0_regs[CP0_COUNT_REG] += (skip & UINT32_C(0xFFFFFFFC)); \
-        else cached_interp_##name(); \
+        if(*cp0_cycle_count < 0) \
+        { \
+            cp0_regs[CP0_COUNT_REG] -= *cp0_cycle_count; \
+            *cp0_cycle_count = 0; \
+        } \
     } \
-    else cached_interp_##name(); \
+    cached_interp_##name(); \
 }
 
 /* These macros allow direct access to parsed opcode fields. */
--- a/src/device/r4300/interrupt.c
+++ b/src/device/r4300/interrupt.c
@@ -102,31 +102,13 @@ static int before_event(const struct cp0* cp0, unsigned int evt1, unsigned int e
 {
     const uint32_t* cp0_regs = r4300_cp0_regs((struct cp0*)cp0); /* OK to cast away const qualifier */
     uint32_t count = cp0_regs[CP0_COUNT_REG];
+    int* cp0_cycle_count = r4300_cp0_cycle_count((struct cp0*)cp0);
 
-    if (evt1 - count < UINT32_C(0x80000000))
-    {
-        if (evt2 - count < UINT32_C(0x80000000))
-        {
-            if ((evt1 - count) < (evt2 - count)) return 1;
-            else return 0;
-        }
-        else
-        {
-            if ((count - evt2) < UINT32_C(0x10000000))
-            {
-                switch(type2)
-                {
-                    case SPECIAL_INT:
-                        if (cp0->special_done) return 1;
-                        else return 0;
-                        break;
-                    default:
-                        return 0;
-                }
-            }
-            else return 1;
-        }
-    }
+    /* At least one other interrupt is pending */
+    if (*cp0_cycle_count > 0)
+        count -= *cp0_cycle_count;
+
+    if ((evt1 - count) < (evt2 - count)) return 1;
     else return 0;
 }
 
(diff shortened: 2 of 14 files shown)


Gillou68310:
@loganmc10 sorry for wasting your time on the other PR BTW

loganmc10:
No problem, I'll take a look and do some testing. Just so I understand

> The main limitation is that the number of cycles between current cycle count and next interrupt event cannot exceed 2^31

You mean this is a limitation of the new dynarec? It still doesn't seem great to me, the SPECIAL_INT is still a sort of hack in this PR, even if it is more harmless now

Gillou68310:
> You mean this is a limitation of the new dynarec?

Yes mainly. It's also a limitation with the new design here if we want to keep next_interrupt as 32bits.

> It still doesn't seem great to me, the SPECIAL_INT is still a sort of hack in this PR, even if it is more harmless now

I agree but like I said, without it, it means drop support for 32bits new_dynarec rn.

loganmc10:
Ok I'll take a look. IMO we have way too much dynarec going on as-is. I would drop the old dynarec and the x86 new dynarec completely. Most half decent android devices support arm64 as well. It's so much code to maintain and you are probably the only person who understands how it works.

Gillou68310:
If only new dynarec was perfect ;-)

Gillou68310:
Added support for Hacktarux dynarecs

loganmc10:
Just did a bit of testing with the Cached Interpreter. Twisted Edge Extreme Snowboarding freezes right as the race is starting with this PR.

src/device/r4300/cached_interp.c
```diff
@@ -127,24 +127,26 @@ void cached_interp_##name##_OUT(void) \
...
-        if (skip > 3) cp0_regs[CP0_COUNT_REG] += (skip & UINT32_C(0xFFFFFFFC)); \
-        else cached_interp_##name(); \
+         if(*cp0_next_interrupt < 0) \
+         { \
+           cp0_regs[CP0_COUNT_REG] -= *cp0_next_interrupt; \
+           *cp0_next_interrupt = 0; \
+         } \
     } \
-    else cached_interp_##name(); \
+    cached_interp_##name(); \
```
loganmc10: Is this right? The logic doesn't seem to make how it is currently.

src/device/r4300/interrupt.c
```diff
@@ -296,8 +263,17 @@ void translate_event_queue(struct cp0* cp0, unsigned int base)
     {
         e->data.count = (e->data.count - cp0_regs[CP0_COUNT_REG]) + base;
     }
+
+	cp0_regs[CP0_COUNT_REG] = base;
```
loganmc10: Too many spaces here

src/device/r4300/interrupt.c
```diff
@@ -331,21 +308,23 @@ void load_eventqueue_infos(struct cp0* cp0, const char *buf)
...
 }
 
 void init_interrupt(struct cp0* cp0)
 {
-    cp0->special_done = 1;
-
     clear_queue(&cp0->q);
-    add_interrupt_event_count(cp0, SPECIAL_INT, 0);
+    add_interrupt_event_count(cp0, SPECIAL_INT, 0x80000000);
+    add_interrupt_event_count(cp0, COMPARE_INT, 0);
```
loganmc10: I guess we'd need a real hardware test to be sure, but I don't think this is right. The COUNT_REG starts at 0x5000 in our code, so I would think that means the first COMPARE_INT happens when the COUNT_REG reaches 2^32, not at 0

Gillou68310:
> Just did a bit of testing with the Cached Interpreter. Twisted Edge Extreme Snowboarding freezes right as the race is starting with this PR.

I guess it's not with freezing with https://github.com/mupen64plus/mupen64plus-core/pull/701?

src/device/r4300/interrupt.c
Gillou68310: 0 is actually when COUNT_REG reaches 2^32

src/device/r4300/interrupt.c
Gillou68310: yep

loganmc10:
Correct, as far as I could tell, there is even some improvement with #701, perhaps you could verify.

With the current master, if you just run through a race (click A through the menu and complete a race), about halfway through the race, the audio usually glitches out for about a second, and it goes all static.

With #701 I've never experienced that. The game appears to use the COMPARE_INT to time the audio, when I was working on #701, when it wasn't quite working right, the audio in Twisted Edge actually cut out completely part way through the race.

src/device/r4300/cached_interp.c
Gillou68310: Yeah I changed the logic here, I actually don't understand the original logic.
What I do here is, if in an idle loop just update the count to the next interrupt event count

loganmc10:
It's probably common to BOSS games, World Driver Championship appears to freeze as well. It froze for me when trying to start a race, and once just when I hit start for the first time

Gillou68310:
Weird I cannot reproduce the freeze

loganmc10:
U version with the cshced interpreter?

Gillou68310:
Nevermind I was able to reproduce. I pushed a fix for the issue.

Gillou68310:
@loganmc10 @m4xw I update the PR, all dynarecs should work now.
Please test and report any regression.

loganmc10:
I could have done something wrong, but it seems like Conkers doesn't boot (tested with cached interpreter). This could also be me (or my computer) as well, but I still get the big static sound halfway through a race in Twisted Edge, something that appears to be fixed in 701

This still seems like the wrong approach to me, the goal of removing the special int was to simplify the interrupt code and make the interrupt timing more precise/correct.

There must be some way to do 64 bit math using 32 bit ASM? I assume using 2 registers? I'm not really sure why the dynarecs are involved with interrupt timing/incrementing the count register, could that not be left to the regular cp0_update_count function?

Gillou68310:
I indeed broke Conkers, I will fix that.
I'm pretty sure the audio issue in twisted edge is caused by :
```
if (next_ai_event <= cp0_regs[CP0_COUNT_REG])
    return 0;
```

this should actually be :
```
if ((int)(cp0_regs[CP0_COUNT_REG] - next_ai_event) >= 0)
    return 0;
```

> This still seems like the wrong approach to me, the goal of removing the special int was to simplify the interrupt code and make the interrupt timing more precise/correct.

I understand that the special int looks hacky and it actually was in the old design. Now it's pretty clear what it is used for. Having 64 count does simplify the code but does not make interrupt timing more precise/correct in any way so I'm not sure it's worth it.

> There must be some way to do 64 bit math using 32 bit ASM? I assume using 2 registers?

ofc we can, but that's actually the issue with the new_dynarec, I need one additional register for cycle count and I don't have one.

> I'm not really sure why the dynarecs are involved with interrupt timing/incrementing the count register, could that not be left to the regular cp0_update_count function?

Calling C function from recompiled code involve saving currently used register to stack and restore on function return. Dynarecs usally try to avoid doing that for speed. For this and a few other reasons cycle count update is done via recompiled code.

Gillou68310:
@loganmc10 I pushed a few fixes, could you test again?

Gillou68310:
Thanks for the feedback! I rebased, fixed the warnings and remove tabs.

src/device/r4300/cached_interp.c
```diff
@@ -127,24 +127,26 @@ void cached_interp_##name##_OUT(void) \
...
     const int take_jump = (condition); \
-    int skip; \
     if (cop1 && check_cop1_unusable(r4300)) return; \
     if (take_jump) \
     { \
         cp0_update_count(r4300); \
-        skip = *r4300_cp0_next_interrupt(&r4300->cp0) - cp0_regs[CP0_COUNT_REG]; \
-        if (skip > 3) cp0_regs[CP0_COUNT_REG] += (skip & UINT32_C(0xFFFFFFFC)); \
-        else cached_interp_##name(); \
+         if(*cp0_cycle_count < 0) \
```
loganmc10: I know I'm a real stickler for indenting, but this still looks off by 1 space

Gillou68310: lol good catch

loganmc10:
I think it looks good, I would still probably wait for a review from @bsmiles32 if possible before merging, it touches a fairly significant portion of the code

Gillou68310:
Ok I fixed the indentation

loganmc10:
Unless someone else is intending on reviewing this I would say it should be merged. I've tested a decent number of games, I don't see any regressions. This will hopefully fix some random freezing that the current SPECIAL_INT causes

PR #709, opened 2020-02-18
Proper handling of divide by zero

loganmc10:
I was testing these test ROMs:

https://github.com/PeterLemon/N64/tree/master/CPUTest

We pass all the CPU tests except for division.

I took this implementation from https://github.com/n64dev/cen64/pull/79. So credit to @sp1187 for this code.

After this PR, all the division test ROMs pass. The only commercial game that I know of that consistently tries to do division by zero is Toy Story 2 (it currently spits out a bunch of "DIV by zero" warnings). I tested that game and it still seems to work fine.

```rrs32 == INT32_MIN && rrt32 == -1``` is a special case, the result of that equation is actually an integer overflow, which was crashing the emulator, so it needs to be dealt with in an if statement, same as the cen64 implementation.

### DIFF
--- a/src/device/r4300/mips_instructions.def
+++ b/src/device/r4300/mips_instructions.def
@@ -815,10 +815,22 @@ DECLARE_INSTRUCTION(DIV)
     DECLARE_R4300
     if (rrt32)
     {
-        *r4300_mult_lo(r4300) = SE32(rrs32 / rrt32);
-        *r4300_mult_hi(r4300) = SE32(rrs32 % rrt32);
+        if (rrs32 == INT32_MIN && rrt32 == -1)
+        {
+            *r4300_mult_lo(r4300) = SE32(rrs32);
+            *r4300_mult_hi(r4300) = 0;
+        }
+        else
+        {
+            *r4300_mult_lo(r4300) = SE32(rrs32 / rrt32);
+            *r4300_mult_hi(r4300) = SE32(rrs32 % rrt32);
+        }
     }
-    else { DebugMessage(M64MSG_ERROR, "DIV: divide by 0"); }
+    else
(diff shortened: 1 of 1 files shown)


Gillou68310:
The DIV/DIVU implem does not match cen's one, why is that?

loganmc10:
I assume you mean:
```rrs32 == INT32_MIN && rrt32 == -1``` in DIV

cen64 does this for DDIV (using INT64_MIN), but not DIV. I'm not really sure why that was omitted, the test ROM includes the case of INT32_MIN/-1, which causes an integer overflow, so it's a valid case for DIV as well.

I'm not sure what other differences you mean, CEN64 combines DIV and DIVU into one function, so the test:
```div = rs < 0 ? 1 : -1;``` is always -1 for DIVU, since ```rs``` must be positive.

Gillou68310:
> `rrs32 == INT32_MIN && rrt32 == -1` in DIV

I was talking about that indeed, anyway if the test ROM pass, I guess the implementation is correct

loganmc10:
Yes, without that the emulator crashes when running the test ROM because of the integer overflow, I'm not aware of any game that actually attempts that equation besides this test ROM though

sp1187:
If I remember correctly the fix wasn't necessary for DIV since the actual division operation performed by the emulator is 64-bit, therefore there can not be an overflow in this case. (see https://github.com/n64dev/cen64/pull/79/commits/65b5a08cd612dcfead72cfb12f7b405df74d7061#diff-ceb3e456120f6dff3fc50512d81546a5R711).

loganmc10:
Yes in our case rrs32 is an int32_t, which I guess explains why I needed it

Issue #493, opened 2017-12-01
FBInfo missing some data with dynarec


loganmc10:
To reproduce: Use the latest master of both mupen64plus-core and GLideN64. Launch Super Mario 64 with the dynarec, the head is missing:

![super_mario_64-000](https://user-images.githubusercontent.com/848146/33468844-60037190-d61c-11e7-92f6-3b7af8927e97.png)

It appears when using the Interpreter, therefore the dynarec must be missing some of the reads/writes that FBInfo should be intercepting

bsmiles32:
Uncommenting INTERPRET_LWC1 in x86_64/interpret.h makes Mario Face appear again.

Issue #214, opened 2016-12-04
FB Info is not fully implemented


fzurita:
I'm inspecting the code and it seems like the FB Info API is only partially implemented. I can see that the hooks are in place in memory.c, I can also see that gr4300.c for the old dynarec is setup to be able to use those hooks but doesn't.

For the new dynarec (arm or x860, the fb info hooks do not seem adaptable at all.

Gillou68310:
When ```protect_framebuffers``` is called the corresponding memory region is mapped to ```read_rdramFB/b/h/d``` and ```write_rdramFB/b/h/d``` functions. So any read word/byte/hword/dword and write word/byte/hword/dword instructions that tries to access this region will jump to the corresponding handler.

```
for (j=start; j<=end; j++)
{
    map_region(0x8000+j, M64P_MEM_RDRAM, RW(rdramFB));
    map_region(0xa000+j, M64P_MEM_RDRAM, RW(rdramFB));
}
```

```unprotect_framebuffers``` will map the corresponding memory region back to ```read_rdram/b/h/d``` and ```write_rdram/b/h/d``` functions.

The only problem I see here is that the new_dynarec maps the memory region 0x80000000 to 0x807FFFFF with it's own ```write_rdram/b/h/d_new``` functions. So the ```unprotect_framebuffers``` function should restore this mapping if it was overwritten by ```protect_framebuffers```.

fzurita:
Interesting. I could not find hours that was working. I wonder why it's not working with GLideN64.

Gillou68310:
Is it working with project64?

fzurita:
Project 64 doesn't implement it.

Regular Mupen64 and 1964 both implement it correctly (I think).

Here is some backstory:

https://github.com/gonetz/GLideN64/issues/808

Gillou68310:
Thanks, I compared the current code with regular Mupen64 and saw no differences. Not sure what broke it.

fzurita:
I can some debug to determine if the API calls are ever being made. There must be something missing somewhere.

Gillou68310:
> I can some debug to determine if the API calls are ever being made

Already checked that, they are being called.

Gillou68310:
Tested with banjo-kazooie

fzurita:
Hmmm, does the puzzle transition work?

Gillou68310:
Nope

rcaridade145:
Any news on this with the recent merges?

[...]

PR #710, opened 2020-02-19
Disable FBInfo support for dynarecs

loganmc10:
Right now FBInfo doesn't work properly on either dynarec, see:

https://github.com/mupen64plus/mupen64plus-core/issues/493

https://github.com/mupen64plus/mupen64plus-core/issues/214

It's better to not advertise support for FBInfo at all if it isn't working, since plugins like GLideN64 have it enabled by default for certain games.

From my testing it works pretty reliably with the interpreters

### DIFF
--- a/src/device/rcp/rdp/fb.c
+++ b/src/device/rcp/rdp/fb.c
@@ -185,7 +185,8 @@ void protect_framebuffers(struct fb* fb)
     struct mem_mapping fb_mapping = { 0, 0, M64P_MEM_RDRAM, { fb, RW(rdram_fb) } };
 
     /* check API support */
-    if (!(gfx.fBGetFrameBufferInfo && gfx.fBRead && gfx.fBWrite)) {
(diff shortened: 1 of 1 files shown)


fzurita:
I agree that FB Info doesn't work the dynarecs. Looks good time. This allows not having to manually turn off FB Info in the graphics plugin when running with the dynarecs.

If FB Info does start working with the dynarecs, we will have to remember to disable the checks.

fzurita:
Even better, the check is in only one place now.

loganmc10:
Yeah I realized what I was doing was redundant, hopefully it can get sorted out in the dynarecs one day and this can be removed, it's a pretty cool feature

bsmiles32:
Please add a comment in the code to explain why it is disabled for dynarec.

loganmc10:
done

Gillou68310:
I have an idea to fix FB Info with dynarecs using exceptions, but in the meantime this should be acceptable.

PR #715, opened 2020-02-22
Save 64DD data to SRAM folder in a zip file on emulation end

fzurita:
This adds support for saving and loading 64DD data from a compressed zip file instead of the original image. The compressed zip file is stored in the SRAM folder like other save data but only once emulation finishes. The save file will be extracted and loaded when emulation starts up again for a disk with the same file name.

I tried using bsdiff before compressing the file, but that turned out to be way too slow. Creating a patch for a ~60MB 64DD image in my relatively high end phone took about 2 minutes to run, so it did not seem like an acceptable option to use bsdiff.

Also, I did not test the ```get_filename_from_path``` function I added for windows or MacOs, I'm hopefully someone else can test that.

### DIFF
--- a/src/backends/file_storage.c
+++ b/src/backends/file_storage.c
@@ -22,6 +22,7 @@
 #include "file_storage.h"
 
 #include <stdlib.h>
+#include <zlib.h>
 
 #include "api/callbacks.h"
 #include "api/m64p_types.h"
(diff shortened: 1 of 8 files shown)


fzurita:
@bsmiles32 are you ok with this?

fzurita:
Closing in favor of the work from @LuigiBlood

PR #716, opened 2020-02-22
Change some DD logging to only happen in verbose mode

loganmc10:
Excessive logging can really kill performance (depending on the frontend).

The DD controller is currently printing out a lot of logs just for things like DMA reads and writes. This converts the logs to "verbose" mode, so they'll only print out if you have verbose mode enabled


Narann:
Looks simple enough.

PR #717, opened 2020-02-23
Don't check event count to test if event is pending

loganmc10:
Right now we use the "count" of the next event to check if that type of event is pending.

This could lead to the (admittedly rare) situation where the next event is scheduled at COUNT_REG 0, and get_event() would return the incorrect answer.

### DIFF
--- a/src/main/savestates.c
+++ b/src/main/savestates.c
@@ -1890,7 +1890,11 @@ static int savestates_save_pj64(const struct device* dev,
     PUTARRAY(pj64_magic, curr, unsigned char, 4);
     PUTDATA(curr, unsigned int, SaveRDRAMSize);
     PUTARRAY(dev->cart.cart_rom.rom, curr, unsigned int, 0x40/4);
-    PUTDATA(curr, uint32_t, get_event(&dev->r4300.cp0.q, VI_INT) - cp0_regs[CP0_COUNT_REG]); // vi_timer
+    uint32_t* next_vi = get_event(&dev->r4300.cp0.q, VI_INT);
+    if (next_vi != NULL)
+        PUTDATA(curr, uint32_t, *next_vi - cp0_regs[CP0_COUNT_REG]); // vi_timer
+    else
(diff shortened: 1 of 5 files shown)


loganmc10:
rebased against master and simplified a bit

Gillou68310:
I would rather return a pointer to either ```node ``` or ```interrupt_event``` instead.
If pointer is not null just get the count from it, otherwise it means no event of coresponding type in event list.

Gillou68310:
Rare situation indeed but nice catch ;-)

Gillou68310:
Well on a second thought just return a pointer ```unsigned int```

Gillou68310:
```
unsigned int* get_event(const struct interrupt_queue* q, int type)
{
    const struct node* e = q->first;

    if (e == NULL) {
        return NULL;
    }

    if (e->data.type == type) {
        return &e->data.count;
    }

    for (; e->next != NULL && e->next->data.type != type; e = e->next);

    return (e->next != NULL)
        ? &e->next->data.count
        : NULL;
}
```

loganmc10:
Good suggestion, I've updated the PR

src/main/savestates.c
```diff
@@ -1890,7 +1890,7 @@ static int savestates_save_pj64(const struct device* dev,
     PUTARRAY(pj64_magic, curr, unsigned char, 4);
     PUTDATA(curr, unsigned int, SaveRDRAMSize);
     PUTARRAY(dev->cart.cart_rom.rom, curr, unsigned int, 0x40/4);
-    PUTDATA(curr, uint32_t, get_event(&dev->r4300.cp0.q, VI_INT) - cp0_regs[CP0_COUNT_REG]); // vi_timer
+    PUTDATA(curr, uint32_t, *get_event(&dev->r4300.cp0.q, VI_INT) - cp0_regs[CP0_COUNT_REG]); // vi_timer
```
Gillou68310: This probably will never happen but maybe test the pointer before dereferencing?

loganmc10:
Good suggestions, I've updated the PR

PR #718, opened 2020-02-24
Adjust DD DMA timing

loganmc10:
I believe the formula for DD DMA timing (length * 63 / 25) was taken from MAME:
https://github.com/mamedev/mame/blob/master/src/mame/machine/n64.cpp#L1573

They say "Measured as between 2.53 cycles per byte and 2.55 cycles per byte". 63/25 is 2.52.

We time our interrupts using the COUNT register, which increments at 1/2 the clock speed. To me this means we should be using 63/50.

This change allows F-Zero DD expansion to boot, before it was just sitting on the "DD Loading" screen. The only other game I have tested it the "Dawn to Dusk" Zelda OoT homebrew that was recently released.

The rest of the DD games should be tested to make sure there aren't regressions, I haven't found a really good way to boot them because of the requirement for a cartridge ROM. Maybe someone that knows how to get all those games booted can give this PR a try? I wouldn't merge this until the rest of the DD games can be tested for regressions

### DIFF
--- a/src/device/dd/dd_controller.c
+++ b/src/device/dd/dd_controller.c
@@ -623,14 +623,14 @@ unsigned int dd_dom_dma_read(void* opaque, const uint8_t* dram, uint32_t dram_ad
     else {
         DebugMessage(M64MSG_ERROR, "Unknown DD dma read dram=%08x  cart=%08x length=%08x",
             dram_addr, cart_addr, length);
-        return (length * 63) / 25;
+        return (length * 63) / 50;
     }
 
     for (i = 0; i < length; ++i) {
         mem[(cart_addr + i) ^ S8] = dram[(dram_addr + i) ^ S8];
     }
 
-    return (length * 63) / 25;
+    return (length * 63) / 50;
 }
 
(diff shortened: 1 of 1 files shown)


loganmc10:
I guess I should say to: I tested using the cached interpreter. F-Zero already worked with the dynarec for some reason, but not the interpreter. This change appears to make the same work in both modes. I'm still not 100% sure if that was just a fluke, or as the result of more accurate timing though.

fzurita:
Ok, I will test soon.

bsmiles32:
Note: To test non-combo disks, just put a regular japanese cart rom (which will be "ignored"). If you put a non japanese game, country check may fail.

So I tested most of DD disk and there are some improvements such as DPGJ that now works and EFZJ that works with pure interpreter, cached interpreter, dynarec.

Some new warnings are emitted with DKDJ, DKIJ, DKKJ, DRDJ about unaligned dword reads which is a bit surprising. And they don't really boot. But they didn't work much better before, so I guess we can merge this.

PR #721, opened 2020-02-25
Emulate RSP DMA FIFO queue

loganmc10:
image

I modelled the FIFO queue after the AI FIFO queue. There is not actually an interrupt that happens at the end of the DMA, but I've used the interrupt/event queue to schedule the end of the DMA, since that is the most convenient way.

I haven't noticed any difference in any game, either an improvement or a regression. As suggested, most games just check DMA_BUSY to see if the DMA is complete, since we never set that bit in our current code, games probably just assume the DMA was _really fast_. This should get us a little closer to timing accuracy, since the games will wait a bit for the DMA to complete, like on a real N64

### DIFF
--- a/src/device/rcp/rsp/rsp_core.c
+++ b/src/device/rcp/rsp/rsp_core.c
@@ -34,60 +34,103 @@
 #include "main/profile.h"
 #endif
 #include "plugin/plugin.h"
+#include "api/callbacks.h"
 
-static void dma_sp_write(struct rsp_core* sp)
+static void do_sp_dma(struct rsp_core* sp, const struct sp_dma* dma)
 {
     unsigned int i,j;
 
-    unsigned int l = sp->regs[SP_RD_LEN_REG];
+    unsigned int l = dma->length;
 
     unsigned int length = ((l & 0xfff) | 7) + 1;
     unsigned int count = ((l >> 12) & 0xff) + 1;
     unsigned int skip = ((l >> 20) & 0xfff);
 
-    unsigned int memaddr = sp->regs[SP_MEM_ADDR_REG] & 0xfff;
-    unsigned int dramaddr = sp->regs[SP_DRAM_ADDR_REG] & 0xffffff;
+    unsigned int memaddr = dma->memaddr & 0xfff;
+    unsigned int dramaddr = dma->dramaddr & 0xffffff;
 
-    unsigned char *spmem = (unsigned char*)sp->mem + (sp->regs[SP_MEM_ADDR_REG] & 0x1000);
+    unsigned char *spmem = (unsigned char*)sp->mem + (dma->memaddr & 0x1000);
     unsigned char *dram = (unsigned char*)sp->ri->rdram->dram;
 
-    for(j=0; j<count; j++) {
-        pre_framebuffer_read(&sp->dp->fb, dramaddr);
+    if (dma->dir == SP_DMA_READ)
+    {
+        for(j=0; j<count; j++) {
+            for(i=0; i<length; i++) {
+                dram[dramaddr^S8] = spmem[memaddr^S8];
+                memaddr++;
+                dramaddr++;
+            }
 
-        for(i=0; i<length; i++) {
-            spmem[memaddr^S8] = dram[dramaddr^S8];
-            memaddr++;
-            dramaddr++;
+            post_framebuffer_write(&sp->dp->fb, dramaddr - length, length);
+            dramaddr+=skip;
         }
-        dramaddr+=skip;
+    }
+    else
+    {
+        for(j=0; j<count; j++) {
+            pre_framebuffer_read(&sp->dp->fb, dramaddr);
+
+            for(i=0; i<length; i++) {
+                spmem[memaddr^S8] = dram[dramaddr^S8];
+                memaddr++;
+                dramaddr++;
+            }
+            dramaddr+=skip;
+        }
+    }
+
+    /* schedule end of dma event */
+    cp0_update_count(sp->mi->r4300);
+    add_interrupt_event(&sp->mi->r4300->cp0, RSP_DMA_EVT, (count * length) / 8);
+}
+
+static void fifo_push(struct rsp_core* sp, uint32_t dir)
+{
+    if (sp->regs[SP_DMA_FULL_REG])
+    {
+        DebugMessage(M64MSG_WARNING, "RSP DMA attempted but FIFO queue already full.");
+        return;
+    }
+
+    if (sp->regs[SP_DMA_BUSY_REG])
+    {
+        sp->fifo[1].dir = dir;
+        sp->fifo[1].length = dir == SP_DMA_READ ? sp->regs[SP_WR_LEN_REG] : sp->regs[SP_RD_LEN_REG];
+        sp->fifo[1].memaddr = sp->regs[SP_MEM_ADDR_REG];
+        sp->fifo[1].dramaddr = sp->regs[SP_DRAM_ADDR_REG];
+        sp->regs[SP_DMA_FULL_REG] = 1;
+        sp->regs[SP_STATUS_REG] |= SP_STATUS_DMA_FULL;
+    }
+    else
+    {
+        sp->fifo[0].dir = dir;
+        sp->fifo[0].length = dir == SP_DMA_READ ? sp->regs[SP_WR_LEN_REG] : sp->regs[SP_RD_LEN_REG];
+        sp->fifo[0].memaddr = sp->regs[SP_MEM_ADDR_REG];
+        sp->fifo[0].dramaddr = sp->regs[SP_DRAM_ADDR_REG];
+        sp->regs[SP_DMA_BUSY_REG] = 1;
+        sp->regs[SP_STATUS_REG] |= SP_STATUS_DMA_BUSY;
+
+        do_sp_dma(sp, &sp->fifo[0]);
     }
 }
 
-static void dma_sp_read(struct rsp_core* sp)
+static void fifo_pop(struct rsp_core* sp)
 {
-    unsigned int i,j;
+    if (sp->regs[SP_DMA_FULL_REG])
+    {
+        sp->fifo[0].dir = sp->fifo[1].dir;
+        sp->fifo[0].length = sp->fifo[1].length;
+        sp->fifo[0].memaddr = sp->fifo[1].memaddr;
+        sp->fifo[0].dramaddr = sp->fifo[1].dramaddr;
+        sp->regs[SP_DMA_FULL_REG] = 0;
+        sp->regs[SP_STATUS_REG] &= ~SP_STATUS_DMA_FULL;
 
-    unsigned int l = sp->regs[SP_WR_LEN_REG];
-
-    unsigned int length = ((l & 0xfff) | 7) + 1;
-    unsigned int count = ((l >> 12) & 0xff) + 1;
-    unsigned int skip = ((l >> 20) & 0xfff);
-
-    unsigned int memaddr = sp->regs[SP_MEM_ADDR_REG] & 0xfff;
-    unsigned int dramaddr = sp->regs[SP_DRAM_ADDR_REG] & 0xffffff;
-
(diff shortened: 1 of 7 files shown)


Gillou68310:
I'm not sure to understand the point of having a new fake interrupt here. Like you said DMA completion is triggered by a flag, this is usually done via idle loops eg:

```
04001478 mfc0   r2, SP_DMA_FULL
0400147C bne    r2, 0, $0400147C
04001480 mfc0   r2, SP_DMA_FULL
```

Idle loops are skiped in the emulator, which means cycle count is set to the next interrupt value.
So rsp_end_of_dma_event purpose is just to delay the count?

Cycle count is a big hack that is really far from accuracy (see count_per_op).
It's just my opinion but this is more likely to introduce new regressions than help timing accuracy.
It's even more true if there's no issue in the first place.

loganmc10:
It's just an attempt to make the emulator more accurate. It's very clear in the documentation that this is how the DMAs are supposed to be treated.

You're right that right now we don't emulate the count register very accurately, but we might one day, and even with what we do now, games still read the count register and look at compare interrupts for timing. Right now we basically emulate an RSP DMA as taking 0 cycles. This should bring it a lot closer to accuracy.

I've actually tested quite a few games with this, and I haven't found a single game that actually double buffers an RSP DMA (they don't start a new DMA when DMA_BUSY is set), so you are correct that every game I've seen probably spins at an idle loop waiting for DMA_BUSY to clear. But that doesn't mean there isn't some game out there that does something differently.

Gillou68310:
> You're right that right now we don't emulate the count register very accurately, but we might one day

This day will probably never happen! We already have a cycle accurate N64 emulator.
IMO mupen's purpose is not cycle accuracy, it's more a trade of between compatibility and speed.

While your implementation seems correct, it'll cost a bit more (probably negligible) than what we have rn.

I'm not questioning your implementation but is it really worth the overhead?

Gillou68310:
> I'm not questioning your implementation but is it really worth the overhead?

This is an open question BTW ;-)

loganmc10:
Yes this isn't urgent, and I wouldn't mind hearing the opinions of others.

Like you say, I think the performance impact of this would be minimal, since idle loops are basically skipped like you say. I think it's still important for accuracy, picture this scenario:

A game writes to SP_RD_LEN_REG/SP_DRAM_ADDR_REG twice to queue up 2 DMAs. It knows that it's impossible for the 2nd DMA to happen right away, so the DRAM contents at SP_DRAM_ADDR_REG for the 2nd DMA aren't quite right yet, but they will be in 4 cycles when the game executes some other instruction. This would be "risky" behaviour by a game developer, since you are counting on the speed of the DMA to be predictable, however I think it would be possible.

Like I said, I haven't come across a game that does something like that, but there are a lot of games out there, I think it's possible that at least a couple might try something like that.

Gillou68310:
> A game writes to SP_RD_LEN_REG/SP_DRAM_ADDR_REG twice to queue up 2 DMAs. It knows that it's impossible for the 2nd DMA to happen right away, so the DRAM contents at SP_DRAM_ADDR_REG for the 2nd DMA aren't quite right yet, but they will be in 4 cycles when the game executes some other instruction. This would be "risky" behaviour by a game developer, since you are counting on the speed of the DMA to be predictable, however I think it would be possible.

Scheduling a DMA transfert while the DRAM data to be transfered is not in the correct state is just nonsense, nobody will ever do that!

loganmc10:
BOSS games doing something similar for AI DMAs (you can read my explanation from a while back here: https://github.com/mupen64plus/mupen64plus-core/issues/200#issuecomment-316780301)

They only use 1 DRAM address for audio DMAs, and they modify the contents of the DRAM at that location while the DMA is happening. It's not exactly the same thing, it's safer because the game reads AI_LEN_REG to determine what point in the DMA the game is at, thus knowing which areas of DRAM are safe to modify.

Plus, I would argue that 

> nobody _will_ ever do that

May be true, we wouldn't know for sure without testing every game, but they _can_ do that. There is always the possibility that somebody creates a homebrew game that does this, that would work on a real N64, but not mupen64plus

src/device/rcp/rsp/rsp_core.h
```diff
@@ -75,6 +75,22 @@ enum sp_registers2
     SP_REGS2_COUNT
 };
 
+enum sp_dma_dir
+{
+    SP_NO_DMA,
```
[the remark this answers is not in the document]
loganmc10: It probably isn't a huge deal, but this is there so that when the struct is initialized by ```memset(sp->fifo, 0, SP_DMA_FIFO_SIZE*sizeof(struct sp_dma));```, the dma_dir is SP_NO_DMA instead of read/write, which in my mind is more correct.

src/main/savestates.c
```diff
@@ -839,6 +839,12 @@ static int savestates_load_m64p(struct device* dev, char *filepath)
             curr += sizeof(uint32_t);
 #endif
         }
+
+        if (version >= 0x00010700)
+        {
+            dev->sp.fifo[0] = GETDATA(curr, struct sp_dma);
```
bsmiles32, line 845: You cannot serialize/deserialize a struct in one shot, you must do it field by field, otherwise run into the problem a compiler-specific struct layout (adding padding bytes for instance).

Also, you must provide a fallback (for instance empty fifo) for savestates < 1.7.

src/device/rcp/rsp/rsp_core.c
```diff
@@ -89,6 +93,60 @@ static void dma_sp_read(struct rsp_core* sp)
...
+
+    /* schedule end of dma event */
+    cp0_update_count(sp->mi->r4300);
+    add_interrupt_event(&sp->mi->r4300->cp0, RSP_DMA_INT, (count * length) / 8);
+}
+
+static void fifo_push(struct rsp_core* sp, uint32_t dir)
+{
+    if (sp->regs[SP_DMA_BUSY_REG])
+    {
```
bsmiles32, line 105: Maybe add a warning if SP_DMA_FULL_REG != 0 ?

src/device/rcp/rsp/rsp_core.c
```diff
@@ -35,20 +35,20 @@
 #endif
 #include "plugin/plugin.h"
 
-static void dma_sp_write(struct rsp_core* sp)
+static void dma_sp_write(struct rsp_core* sp, struct sp_dma* dma)
```
bsmiles32, line 38: You might regroup dma_sp_write and dma_sp_read into a single do_sp_dma ?

src/device/rcp/rsp/rsp_core.c
```diff
@@ -89,6 +93,60 @@ static void dma_sp_read(struct rsp_core* sp)
...
+        sp->regs[SP_STATUS_REG] &= ~SP_STATUS_DMA_FULL;
+
+        if (sp->fifo[0].dir == SP_DMA_READ)
+            dma_sp_read(sp, &sp->fifo[0]);
+        else
+            dma_sp_write(sp, &sp->fifo[0]);
+    }
+    else
+    {
+        sp->regs[SP_DMA_BUSY_REG] = 0;
```
bsmiles32, line 147: You may emit a warning if SP_DMA_BUSY_REG == 0

loganmc10: I don't think this would be possible, DMA_BUSY_REG is set in fifo_push

src/main/savestates.c
loganmc10: I will update to reference each member. I don't think setting a fallback for savestates < 1.7 would be needed. If the save state is < 1.7, there can't be one of these DMAs in the event queue, so the fifo values have no meaning/use

loganmc10:
Alright I believe I incorported all the feedback, I also renamed it from "RSP_DMA_INT" to "RSP_DMA_EVT", since it's more of an "event" than an actual interrupt

src/device/rcp/rsp/rsp_core.c
```diff
@@ -34,60 +34,93 @@
...
     unsigned int skip = ((l >> 20) & 0xfff);
 
-    unsigned int memaddr = sp->regs[SP_MEM_ADDR_REG] & 0xfff;
-    unsigned int dramaddr = sp->regs[SP_DRAM_ADDR_REG] & 0xffffff;
+    unsigned int memaddr = dma->memaddr & 0xfff;
+    unsigned int dramaddr = dma->dramaddr & 0xffffff;
 
-    unsigned char *spmem = (unsigned char*)sp->mem + (sp->regs[SP_MEM_ADDR_REG] & 0x1000);
+    unsigned char *spmem = (unsigned char*)sp->mem + (dma->memaddr & 0x1000);
     unsigned char *dram = (unsigned char*)sp->ri->rdram->dram;
```
bsmiles32, line 54: Do a single if(dma->dir == SP_DMA_READ) at the top of the outermost for loop. Otherwise you get the branch penalty at each iteration. Something like:

```
if (dma->dir == SP_DMA_READ) {
// full dma read logic (no extra if)
}
else {
// full dma write logic (no extra if)
}
```

Note: If you remove the SP_NO_DMA, SP_DMA_READ will equal 0 and therefore might be better to compare against than SP_DMA_WRITE.

src/device/rcp/rsp/rsp_core.c
```diff
@@ -34,60 +34,93 @@
 #include "main/profile.h"
 #endif
 #include "plugin/plugin.h"
+#include "api/callbacks.h"
 
-static void dma_sp_write(struct rsp_core* sp)
+static void do_sp_dma(struct rsp_core* sp, struct sp_dma* dma)
```
bsmiles32, line 39:
```suggestion
static void do_sp_dma(struct rsp_core* sp, const struct sp_dma* dma)
```

src/device/rcp/rsp/rsp_core.c
bsmiles32: Indeed, given how the fifo_{push,pop} is currently called, but it was more for "symmetry" : we checked for full fifo in fifo_push, so I suggested to also check for empty fifo in fifo_pop.

But either way is fine by me.

src/device/rcp/rsp/rsp_core.h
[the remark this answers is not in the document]
bsmiles32: Not a huge fan either of SP_NO_DMA as it is "redundant" with the BUSY_FLAG :
- if BUSY flag is cleared, there is no DMA and fifo content has no meaning
- if BUSY flag is set, at least one DMA (or 2 if FULL flag is set) is in the fifo and it is either a READ or a WRITE DMA.

src/main/savestates.c
```diff
@@ -839,6 +839,18 @@ static int savestates_load_m64p(struct device* dev, char *filepath)
...
+        {
+            dev->sp.fifo[0].dir = GETDATA(curr, uint32_t);
+            dev->sp.fifo[0].length = GETDATA(curr, uint32_t);
+            dev->sp.fifo[0].memaddr = GETDATA(curr, uint32_t);
+            dev->sp.fifo[0].dramaddr = GETDATA(curr, uint32_t);
+            dev->sp.fifo[1].dir = GETDATA(curr, uint32_t);
+            dev->sp.fifo[1].length = GETDATA(curr, uint32_t);
+            dev->sp.fifo[1].memaddr = GETDATA(curr, uint32_t);
+            dev->sp.fifo[1].dramaddr = GETDATA(curr, uint32_t);
+        }
```
bsmiles32, line 853: Still, you might want to explicitely clear DMA_BUSY/DMA_FULL flags to make sure fifo content is not accessed if version is < 1.7

loganmc10:
OK I've incorporated the further feedback, I removed NO_DMA, and I've memset the structs to 0 in savestate load if the version is < 1.7

loganmc10:
I think I've incorporated all the suggested feedback into this PR. It's been included in https://m64p.github.io for months now, I haven't heard of any issues, I think it is good to be merged. It has a merge conflict with https://github.com/mupen64plus/mupen64plus-core/pull/743, so it would be good to merge this so that https://github.com/mupen64plus/mupen64plus-core/pull/743 can be updated to be mergable

PR #725, opened 2020-02-27
Implement PIF write flags properly

loganmc10:
Here is another DMA flag fix, similar to #721 . As is the case with that PR, I haven't actually come across a game that does a CPU write to PIF RAM, but the documentation is clear:

```dma_busy indicates a DMA is in progress (or an IO write operation)```

```dma_busy is set for all dma and io write operations```

```DMA and IO writes generate an si_interrupt upon completion to inform host software that the operation has concluded.```

I would really love to find a game that does this so I can verify this behaviour, I'll probably keep trying to find an example, unless somebody has a way to search through game ASM code for SW operations to the PIF RAM address.

### DIFF
--- a/src/device/device.c
+++ b/src/device/device.c
@@ -201,7 +201,8 @@ void init_device(struct device* dev,
         (uint8_t*)mem_base_u32(base, MM_PIF_MEM),
         jbds, ijbds,
         (uint8_t*)mem_base_u32(base, rom_base) + 0x40,
-        &dev->r4300);
+        &dev->r4300,
+        &dev->si);
 
     init_cart(&dev->cart,
             af_rtc_clock, iaf_rtc_clock,
--- a/src/device/pif/pif.c
+++ b/src/device/pif/pif.c
@@ -32,6 +32,7 @@
 #include "backends/api/joybus.h"
 #include "device/memory/memory.h"
 #include "device/r4300/r4300_core.h"
+#include "device/rcp/si/si_controller.h"
 #include "plugin/plugin.h"
 
 #define __STDC_FORMAT_MACROS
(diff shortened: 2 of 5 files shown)


Gillou68310:
I think you misinterpreted write to PIF RAM here. PIF RAM is not the actual PIF chip RAM it's a classic write to RDRAM.

You then need to write to SI_PIF_ADDR_RD64B_REG or SI_PIF_ADDR_WR64B_REG to actualy start a DMA transfer between RDRAM and PIF chip RAM.

So this implementation is wrong IMO.

However calling ```process_pif_ram``` from ```write_pif_ram``` does not makes sense.
@bsmiles32 maybe you're more familiar with PIF than I am?

loganmc10:
I don't think so, for everything connected to the PI (cartridge ROM, flash, sram), you have 2 ways to access to them. You can do a DMA, or you can do a CPU/IO read or write. The IO read/write allows you to access a single word (maybe dword?) See here: https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/cart/cart.c

That's why there is a write_cart_dom2 and a cart_dom2_dma_write function for example. The same goes for the SI/PIF. For instance the SI documentation shows:

image

The ```write_pif_ram``` function is that IO Write. It is a bit odd that dma_busy is set for an IO/CPU write, but that's what the documentation states should happen.

```write_pif_ram``` get mapped based on address space here: https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/device.c#L152

Gillou68310:
Actually you're right I read the SI specification carefully this time and it looks like your implementation reflects what the specification describes.

Mario 64 is doing a write_pif_ram BTW

loganmc10:
All games seem to do it a single time right at the beginning. I haven't found one yet that does it after that though

loganmc10:
@richard42 I've included this PR in my build since early this year and haven't had any reports of issues, I think it is safe to be merged

PR #728, opened 2020-02-28
Trying to understand how GAWK breaks (DO NOT MERGE)

Jj0YzL5nvJ:


Jj0YzL5nvJ:
My conclusion with this series of tests was that something prevented the correct execution of `gawk.exe` **after** processing the files generated by `cl.exe` with _Win32\Debug_ and/or _Release_ configuration, and that this problem coincidentally affected the processing more severely of the files generated by `cl.exe` with the configuration _x64\Debug_ and/or _Release_.
The potential culprit should be in the reuse of a cache, buffer or a program bug, who knows.

In the end, in this case, `gawk` isn't indispensable and potentially neither is it on other platforms if the command shell is used creatively.

Narann:
Thanks for the investigations!

Issue #722, opened 2020-02-25
AppVeyor is broken


Narann:
* New_Dynarec_Release; Platform: Win32 (OK)
* New_Dynarec_Release; Platform: x64 (OK)
* Release; Platform: Win32 (OK)
* Release; Platform: x64 (Broken)

I rerun the incomplete and the build PR multiple times, here is the log:

```
Build started 2/25/2020 9:29:00 AM.
The target "CopyLocalFilesOutputGroup" listed in a BeforeTargets attribute at "C:\Tools\vcpkg\scripts\buildsystems\msbuild\vcpkg.targets (84,82)" does not exist in the project, and will be ignored.
Project "C:\projects\mupen64plus-core\projects\msvc\mupen64plus-core.vcxproj" on node 1 (default targets).
PrepareForBuild:
  Creating directory "x64\Release\".
  Creating directory "x64\Release\mupen64plus-core.tlog\".
InitializeBuildStatus:
  Creating "x64\Release\mupen64plus-core.tlog\unsuccessfulbuild" because "AlwaysCreate" was specified.
PreBuildEvent:
  del ..\..\src\asm_defines\asm_defines_* & cl /c /Fox64\Release\ /D "__x86_64__" /I ..\..\src ..\..\src\asm_defines\asm_defines.c & "..\..\..\mupen64plus-win32-deps\gawk-3.1.6-1\bin\gawk.exe" -v dest_dir="../../src/asm_defines" -f ..\..\tools\gen_asm_defines.awk x64\Release\asm_defines.obj
  :VCEnd
  Could Not Find C:\projects\mupen64plus-core\src\asm_defines\asm_defines_*
  Microsoft (R) C/C++ Optimizing Compiler Version 19.00.24241.7 for x64
  Copyright (C) Microsoft Corporation.  All rights reserved.
  
  asm_defines.c
CustomBuild:
  Performing Custom Build Tools
  C:\projects\mupen64plus-core\src\device\r4300\x86_64\dyna_start.asm:21: fatal: unable to open include file `asm_defines_nasm.h'
C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.CppCommon.targets(171,5): error MSB6006: "cmd.exe" exited with code 1. [C:\projects\mupen64plus-core\projects\msvc\mupen64plus-core.vcxproj]
[...]

PR #488, opened 2017-11-25
[WIP] Preliminary support of meson builds.

bsmiles32:
In order to lower the maintenance burden of the build system (eg. keeping in sync both Makefile and Visual studio files) I've explored the possibility of migrating to the meson build system [1].
So, I've ported the original Makefile to meson with some minor arrangements to fit in the new paradigm.
What I'm looking for now is feedback from other developers to see if there is interest in that work and what can be done to ease the eventual transition to it. I've tested this on my machine (linux, x64) and I'd like to know if it works in other platforms (windows, bsd, mac), cpu (arm, x86, mips?, ppc?) and dev env (cross-compilation, msvc).

In order to test (assuming the use of the ninja backend):
```
cd <path_to_m64p-core>
mkdir build
cd build
meson ../
ninja
```
You can also inspect and change build options with 
```
meson configure
```
Note that this build system relies on python3 which might not be available by default in windows (should we add it to win32-deps repo ?).

Please report any issue, feedback or comment.
When/If we get this repo into shape with meson, I'll also port other repo to it (should be more easy).

cc: @richard42 

[1] http://mesonbuild.com/

PR #729, opened 2020-03-01
Improve glue code generation

Jj0YzL5nvJ:
These changes make us independent of `gawk` in Windows, and avoid issues with AppVeyor listed in #722

More info in #728


Gillou68310:
I like the idea of using a script for generating asm defines but batch really is an awfull scripting language IMO. Do you think we can use something that could be cross platform?

Jj0YzL5nvJ:
Hmm... the Travis CI indicator has glitched. But it passed.

> Do you think we can use something that could be cross platform?

Python, Lua, Perl, Ruby, JavaScript. Any language works, the task is stupidly easy.

Display an object as a list, look for the `@ASM_DEFINE` pattern, take element `$2` and `$3` on that line, print `#define "$2" ("$3")` to asm_defines_gas.h and print `%define "$2" ("$3")` to asm_defines_nasm.h, that's all.

> but batch really is an awfull scripting language IMO

It's funny, we're literally using batch for AppVeyor building, not to mention that Visual Studio uses **native functions** for commands (by default CMD). Batch _Inception_.

Gillou68310:
> It's funny, we're literally using batch for AppVeyor building, not to mention that Visual Studio uses **native functions** for commands (by default CMD). Batch _Inception_.
> P.S: Most of the code in my batch is an anchor ensuring the position of key elements.

I do not have control over those ;-)
If we can avoid having a batch script in core repo I would be a happy man.

> Python, Lua, Perl, Ruby, JavaScript. Any language works, the task is stupidly easy.

I do like python, I know it's an easy task that's why I think the batch script looks complicated for the task.

Jj0YzL5nvJ:
I prefer anything within VS commands that return a logical error code when they fail, that a mixture command string that returns random values.

> I do like python, I know it's an easy task that's why I think the batch script looks complicated for the task.

You're welcome, BTW... **don't forget** to use `cp1252` if you detect a Windows OS, `utf-8` is incorrect.

Mastergatto:
I'm not sure if this is the case, probably I misunderstood this whole thing, but some time ago @bsmiles32 made an effort to migrate the code to meson, python-based build system (http://mesonbuild.com/)
https://github.com/mupen64plus/mupen64plus-core/pull/488

Jj0YzL5nvJ:
> I think the batch script looks complicated for the task.

This is the code without security measures:
```
for /f "tokens=1,2*" %%a in ('echo %*') do set R1=%%a& set R2=%%b& set R3=%%c
del %R2%asm_defines_* 2>nul
cl /c /Fo%R1% %R3% /I ..\..\src %R2%asm_defines.c
for /f "tokens=2,3" %%a in ('type %R1%asm_defines.obj ^| find "@ASM_DEFINE"') do (
echo #define %%a ^(%%b^)>>%R2%asm_defines_gas.h
echo %%define %%a ^(%%b^)>>%R2%asm_defines_nasm.h
)
```
The proposed code contains the minimum security conditions "for monkeys at the keyboard".

@Mastergatto, Meson is another completely different beast.

Narann:
That's a big job, but wow, it's not nice to read.

Jj0YzL5nvJ:
I take that as a compliment, even knowing that it's not.

The part that requires a clear explanation even knowing the language is this:
```
set R1=1
:L1
set R2=
for /f "tokens=%R1% delims=\" %%a in ('CD') do set R2=%%a
if "%R2%" NEQ "" set /a R1=%R1%+1& goto L1
set /a R1=%R1%-3
if "%R1%" LEQ "1" exit /b 2
```
Many programs in Windows inherit the open path of the program that invokes them, in this case VS. If VS launch the script from the shortest possible logical path `x:\mupen64plus-core\projects\msvc\`, then at the end of the directory count, we subtract the 3 we know and the result can never be less than 2 (`R1`). With this test it's guaranteed that an erroneous relative route is not possible due to not being able to go back more by using `..\..\src\asm_defines\` later.

Gillou68310:
Ok let's not add python dependency to core building. Even if batch is awfull to read, it does the job without external dependencies so that's a pretty strong point.

Gillou68310:
If this is merged we can get rid of gawk in win32-deps!

Jj0YzL5nvJ:
Well, I made the code a little more descriptive, I have the bad habit of wanting to make the code take up little bytes... floppy mindset.
I hadn't noticed it, I barely noticed it when I woke up ^^u

Narann:
TBH, I think one comment per line, starting from the for loop, wouldn't be too much.

Jj0YzL5nvJ:
More syntax, more variables to mess up, new logic, more _intricated_ code, more robust, more mindfuck, BETTER.
After this, I need to look for latent errors in my old scripts and add new optimizations ／人◕ ‿‿ ◕人＼

The only way to understand the code is to play with it. This is already in "debug" mode, run this from CMD, change GAS_VS_ARG2 to whatever you want and move the script to different folder levels to see the differences, a USB is also fine. To see error codes, execute `echo% errorlevel%`.
Using `pause` after each command also works.
```
@echo on
setlocal enableextensions disabledelayedexpansion
:: Forced destruction of environment variables, these can also be modified to debug errors
set GAS_VS_ARG1=%CD%\
set GAS_VS_ARG2=..\..\Hi
set GAS_VS_ARG3TOINF=asdf
:: Adaptation of the parameters sent by VS to use them as variables, if an invalid parameter is sent (syntax conflict), this script will be closed and will not generate an error code for ERRORLEVEL
REM for /f "tokens=1,2*" %%A in ('echo %*') do set GAS_VS_ARG1=%%A& set GAS_VS_ARG2=%%B& set GAS_VS_ARG3TOINF=%%C
:: Making sure we are in the tools folder, it's not really necessary...
REM echo "%~dp0" | find /i "\mupen64plus-core\tools\" >nul
REM if errorlevel 1 exit /b 11
:: Initialized as 1 to execute "tokens", it will be subtracted later
set GAS_RPATH_COUNT=1
:: Initialized with the ABSOLUTE PATH of the VS project file as FIRST condition
set GAS_RPATH_CONDITION=%CD%
set GAS_RPATH_COUNT_SHADOW=0
set GAS_RPATH_MINUS_SHADOW=0
:ADD
:: Destroying GAS_RPATH in each loop because if it can't be reinitialized in 'for', it will return the last valid value
set GAS_RPATH=
for /f "tokens=%GAS_RPATH_COUNT% delims=\" %%a in ('echo %GAS_RPATH_CONDITION%') do set GAS_RPATH=%%a
:: The value '..' can ONLY come from GAS_RPATH_CONDITION = GAS_VS_ARG2, a RELATIVE PATH
if "%GAS_RPATH%"==".." set /a GAS_RPATH_MINUS_SHADOW=%GAS_RPATH_MINUS_SHADOW%+1
:: Addition to GAS_RPATH_COUNT if GAS_RPATH is defined
if defined GAS_RPATH set /a GAS_RPATH_COUNT=%GAS_RPATH_COUNT%+1& goto ADD
:: Correcting GAS_RPATH_COUNT, 1 to execute "tokens" and another one of the last destruction of GAS_RPATH
set /a GAS_RPATH_COUNT=%GAS_RPATH_COUNT%-2
:: Single use loop... to set GAS_RPATH_COUNT_SHADOW, GAS_RPATH_MINUS_SHADOW and the GAS_RPATH_CONDITION's SECOND condition as RELATIVE PATH with GAS_VS_ARG2
if "%GAS_RPATH_CONDITION%" NEQ "%GAS_VS_ARG2%" set GAS_RPATH_COUNT_SHADOW=%GAS_RPATH_COUNT%& set GAS_RPATH_CONDITION=%GAS_VS_ARG2%& set GAS_RPATH_COUNT=1& goto ADD
:: Theoretical error... artificial too, this script ONLY counts folders
if %GAS_RPATH_COUNT_SHADOW% LEQ 0 exit /b 9999
set /a GAS_RPATH_COUNT=%GAS_RPATH_COUNT_SHADOW%-%GAS_RPATH_MINUS_SHADOW%
:: Logical check for ABSOLUTE PATH vs RELATIVE PATH, isn't a truly valid check
if %GAS_RPATH_COUNT% LEQ -1 exit /b 22
pause
```

Jj0YzL5nvJ:
Yeah... too complex, good for me, bad for everything else. Take the Lite version.

Narann:
Does it make sense to add some comments in the current `gen_asm_script.cmd`, at leat to globally understand what it does?

If we can't add comments, I will merge it like this but a general explanation of the script would be great.

Narann:
Thanks a lot @Jj0YzL5nvJ, it's perfect !

PR #731, opened 2020-03-06
Disable FTZ (flush to zero) mode for SSE when FS bit is set

### LINKED ISSUE #724: Murfy doesn't show up in Rayman 2
In Rayman 2 standing on certain tiles should summon murfy but for some reason he doesn't appear. Testing on my raspberry pi 4.

A side question: is there a place that lists known issues? I'd hate to waste people's time reporting known issues.

loganmc10:
Fixes (only for SSE, not ARM) #724 

The c.eq functions are meant to compare 2 float (or double) values. Computers aren't very good at comparing floating numbers, and the emulator seems to return ```equal``` when the N64 expects the values to be ```not equal```.

Using memcmp seems to work properly, the emulator still passes the PeterLemon/krom N64 test ROM for CP1 C EQ functions, and Rayman/Donald Duck collision detection is fixed

### DIFF
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -144,7 +144,7 @@ endif
 # base CFLAGS, LDLIBS, and LDFLAGS
 OPTFLAGS ?= -O3 -flto
 WARNFLAGS ?= -Wall
-CFLAGS += -ffast-math -fno-strict-aliasing -fvisibility=hidden -I../../src -I../../src/asm_defines -DM64P_PARALLEL
+CFLAGS += -fno-strict-aliasing -fvisibility=hidden -I../../src -I../../src/asm_defines -DM64P_PARALLEL
 CXXFLAGS += -fvisibility-inlines-hidden
 LDLIBS +=  -lm
 
--- a/src/device/r4300/cp1.h
+++ b/src/device/r4300/cp1.h
@@ -51,6 +51,10 @@ struct cp1
      * using 32-bit stores. */
     uint32_t rounding_mode;
 
+#ifdef OSAL_SSE
+    uint32_t flush_mode;
+#endif
+
 #ifdef NEW_DYNAREC
 	/* New dynarec uses a different memory layout */
     struct new_dynarec_hot_state* new_dynarec_hot_state;
--- a/src/device/r4300/r4300_core.c
+++ b/src/device/r4300/r4300_core.c
@@ -121,6 +121,13 @@ void poweron_r4300(struct r4300_core* r4300)
 
 void run_r4300(struct r4300_core* r4300)
 {
+#ifdef OSAL_SSE
+    //Save FTZ/DAZ mode
+    unsigned int daz = _MM_GET_DENORMALS_ZERO_MODE();
+    unsigned int ftz = _MM_GET_FLUSH_ZERO_MODE();
+    _MM_SET_DENORMALS_ZERO_MODE(_MM_DENORMALS_ZERO_OFF);
+#endif
+
     *r4300_stop(r4300) = 0;
     g_rom_pause = 0;
 
@@ -195,6 +202,11 @@ void run_r4300(struct r4300_core* r4300)
 #if defined(COUNT_INSTR)
     if (r4300->emumode == EMUMODE_DYNAREC)
         instr_counters_print();
+#endif
+#ifdef OSAL_SSE
+    //Restore FTZ/DAZ mode
+    _MM_SET_DENORMALS_ZERO_MODE(daz);
+    _MM_SET_FLUSH_ZERO_MODE(ftz);
 #endif
 }
 
--- a/src/device/r4300/cp1.c
+++ b/src/device/r4300/cp1.c
@@ -27,6 +27,8 @@
 
 #include "new_dynarec/new_dynarec.h"
 
+#define FCR31_FS_BIT UINT32_C(0x1000000)
+
 void init_cp1(struct cp1* cp1, struct new_dynarec_hot_state* new_dynarec_hot_state)
 {
 #ifdef NEW_DYNAREC
@@ -41,6 +43,9 @@ void poweron_cp1(struct cp1* cp1)
     *r4300_cp1_fcr31(cp1) = 0;
 
     set_fpr_pointers(cp1, UINT32_C(0x34000000)); /* c0_status value at poweron */
+#ifdef OSAL_SSE
+    cp1->flush_mode = _MM_GET_FLUSH_ZERO_MODE();
+#endif
     update_x86_rounding_mode(cp1);
 }
 
@@ -119,6 +124,20 @@ void set_fpr_pointers(struct cp1* cp1, uint32_t newStatus)
 void update_x86_rounding_mode(struct cp1* cp1)
 {
     uint32_t fcr31 = *r4300_cp1_fcr31(cp1);
+
+#ifdef OSAL_SSE
+    uint32_t flush_mode;
+    if (fcr31 & 2)
+        flush_mode = (fcr31 & FCR31_FS_BIT) ? _MM_FLUSH_ZERO_OFF : _MM_FLUSH_ZERO_ON;
+    else
+        flush_mode = _MM_FLUSH_ZERO_ON;
+
+    if (flush_mode != cp1->flush_mode)
+    {
+        _MM_SET_FLUSH_ZERO_MODE(flush_mode);
+        cp1->flush_mode = flush_mode;
+    }
+#endif
 
     switch (fcr31 & 3)
     {
--- a/src/osal/preproc.h
+++ b/src/osal/preproc.h
@@ -42,6 +42,11 @@
   /* for isnan() */
   #include <float.h>
 
+#if defined(_M_X64) || (_M_IX86_FP > 0)
+  #include <immintrin.h>
+  #define OSAL_SSE
+#endif
+
 #else  /* Not WIN32 */
   /* for strcasecmp */
   #include <strings.h>
@@ -57,6 +62,11 @@
 
   /* string functions */
   #define osal_insensitive_strcmp(x, y) strcasecmp(x, y)
+
+#ifdef __SSE__
+  #include <immintrin.h>
+  #define OSAL_SSE
+#endif
 
 #endif
 


loganmc10:
I should say, I haven't tested this with the new dynarec, the issue in Rayman/Donald Duck was also reported there. I'm not sure if the new dynarec uses fpu.h or not

loganmc10:
The issue seems to possibly be related to comparing -0.0 to +0.0

In C, doing:
```
float float1 = 0.0f;
float float2= -0.0f;
if (float1 == float2)
```
Will return true, however doing a memcmp will return false. It would good to know the N64's behaviour when comparing -0.0 and +0.0, whether it is doing a bitwise compare of the registers, or IEEE 754 compare of floating point values. It seems based on my tests that it is doing a bitwise comparison (memcmp)

Gillou68310:
I recently discover a bug in CFC1, value is stored in 32bits register (rrt32) instead of sign extending to 64bits using rrt. Maybe this is related to the collision issue?

Gillou68310:
Dynarec's are using fucomip for floating point compare.

https://www.felixcloutier.com/x86/fcomi:fcomip:fucomi:fucomip

From the docs +0.0 should be equal to -0.0

Zapeth:
> Will return true, however doing a memcmp will return false. It would good to know the N64's behaviour when comparing -0.0 and +0.0, whether it is doing a bitwise compare of the registers, or IEEE 754 compare of floating point values. It seems based on my tests that it is doing a bitwise comparison (memcmp)

I'm a bit confused why memcmp gives a different result for you, IEEE 754 floats include the sign in their encoding, so doing a bitwise comparison should still return a non-equal result (a quick test in C confirms this).

@Gillou68310 is correct though, docs say that -0.0 should equal to +0.0, so maybe this special case should be treated separately

loganmc10:
I'll admit that every document I read seems to say that -0 should =0, so I'm not really sure why the memcmp works, we'll have to do some more investigation, so don't anyone merge this yet

Zapeth:
I did some more testing and I think the issue lies elsewhere. In terms of results, CEN64 seems to return the same values as mupen64 (tested with modified test roms for eq, lt and le instructions), yet there are no issues in the mentioned games when playing.

I assume the (incorrect) results of the memcpy operation just happen to mask the bug that would otherwise occur on mupen64 with the interpreter.

loganmc10:
I just tested mupen64 0.5.1 and the issue doesn't seem to exist there, so it must be a regression somewhere, I'll try and bisect it, it must be pretty old though, I think I tested back to 2017 before and the issue existed then in mupen64plus

loganmc10:
I tested all the way back to May 3, 2014 (which is as far back as I can go before compiling gets complex), and the issue still exists then, very perplexing

Zapeth:
Yet another reason to dislike the plugin system. Anyway, if there is a reason to assume that this is a regression from way back, then it might be worth at least checking the project's [first github commit](https://github.com/mupen64plus/mupen64plus-core/tree/4fcb71627ccb49c12c56145474c893660e2afb57).

If the issue still exists then its probably not be worth going even further back, if it doesn't then it would be possible (albeit tedious) to bisect the issue.

loganmc10:
Actually I don't think this is a regression. It has something to do with the compiler. I just tested the released version of mupen64plus for Windows, and this bug doesn't exist there! I assume that is because it is compiled with Visual Studio

I always compile with gcc, even on Windows, but no messing around with the compiler flags seemed to fix this (debug, different -f flags for math optimizations).

I'm not really a compiler pro, can anyone think of something that would cause this issue with gcc? Also, I don't have Visual Studio on my computer, maybe someone could test using Visual Studio to confirm my suspicion?

mackal:
What happens without fast-math?

Gillou68310:
I can comfirm this issue is non existent with vs compiler.

loganmc10:
I tried removing fast-math and all other compiler optimizations. It may not be related to optimizations, it could be how the rounding functions behave or something

Gillou68310:
Vs compiler is using ucomiss instruction for what it's worth

m4xw:
ucomiss lowers precision https://c9x.me/x86/html/file_module_x86_id_317.html
`Performs and unordered compare of the single-precision floating-point values in the low doublewords`
What mupen does and this pr, both are wrong for float comparisions and are affected by the fp precision of the host system, gcc generates the right behaviour IMO.
Did everyone forget how fp math works?

loganmc10:
I tried doing the comparison like ```return fabsf(a - b) < FLT_EPSILON;``` instead of ```a == b```, didn't fix the issue. It's not necessarily the ```==``` that is the issue, it could be some previous rounding error (or something else) is passing the wrong float to ```==``` to begin with

loganmc10:
I've done a bit more research, this seems to maybe be a 32-bit vs 64-bit problem (I assume the Virtual Studio builds are 32-bit). The problem is solved in 64-bit gcc by adding this compiler flag:
```
-mfpmath=387
```
I also tested 32-bit gcc in Windows, and the problem doesn't exist there at all.

I'm not a programmer by trade, so I'm not going to pretend to know the best way to solve this problem, but it does seem to be related to floating point precision in 64-bit/SSE mode.

Zapeth:
I also looked a bit deeper and wanted to share some results:
- Dynarec path indeed uses x87 fpu instructions for floating point comparisons
- in case of Donald Duck game, when running around in the actual world (with the stars in it), there are sometimes c_eq_s calls being made with `fs=-2.592206E-39` and `ft=0.0` which seems to be the reason for the issue (sse path returns `true` here, x87 `false`)
- switching to x87 code only in c_eq_s fixes the issue in the interpreter
- CEN64 uses sse instructions for c_eq_s too (`comiss`), however the issue is not present there

So while forcing more accurate floating point precision seems to be the immediate solution here, I don't think its the actual issue.
It might just be an inaccurate/incorrect implementation somewhere else that leads to these problematic c_eq_s calls. However tracing this back is tedious, so I'm not going to bother.

Zapeth:
This is bizarre: The same inline assembly from CEN64 doesn't work in mupen64.

The compiled assembly is practically the same (there are two more lines in mupen64's objdump, but I'm not sure if they are revelant to this issue), yet they return different values for the problematic comparison:
- [objdump of cen64](https://pastebin.com/AfcUECmA) (returns 0)
- [objdump of mupen64](https://pastebin.com/uMYg0qiK) (returns 1)

If someone has an explanation for this, I'd love to hear it (its basically just a copy&paste from CEN64's `fpu_cmp_eq_32` to mupen64's `c_eq_s` function).

As for now I'm pretty much done with investigating this issue.

Gillou68310:
Using -mfpmath=387 fixes donald's issue, I didn't tested rayman.

Gillou68310:
Lol @loganmc10 I missed your previous post!

To add to @Zapeth investigations fs=-2.592206E-39 has the suffix #DEN in visual studio which means the value is denormalized.

loganmc10:
I updated the PR to include the current "best fix". I'm not sure if it's really the best solution, my reading leads me to believe that using -mfpmath=387 could have performance impacts, it may be better to use inline assembly that works in fpu.h? Or somehow force the compiler to use x87 FPU instructions just inside the functions in fpu.h, maybe in the way the variables are declared?

Also I'm not really sure what the solution would be for ARM, the problem is present there as well, but I assume there isn't an option equivalent to x87 FPU there

loganmc10:
It seems like gcc has a __float80 type that can force it use a 10-byte type.

I assume for that to work, the cp1 regs would need to be stored as __float80 (or double where that doesn't exist), and the logic for reading/writing to those register would have to be changed around to cast them to float/double when needed.

Zapeth:
Since the issue seems to basically be just about denormalized floating point values I think the better solution would be to just [disable the CPUs DAZ mode](https://software.intel.com/en-us/articles/x87-and-sse-floating-point-assists-in-ia-32-flush-to-zero-ftz-and-denormals-are-zero-daz).

I haven't done any performance analysis, but it should at least perform better than using x87 instructions since the CPU would still execute only sse instructions.

This also seems to be the reason why CEN64 behaves differently with the same code, though I haven't figured out where/how the MXCSR register is configured accordingly in their code (and whether it is set permanently or only toggled).

I actually already tested this with the problematic roms and the issues are no longer present (even with -ffast-math).

loganmc10:
@Zapeth yes excellent suggestion! I updated the PR, I used ```__SSE__``` to determine whether to disable DAZ or not, I'm not sure if that works on Visual Studio

loganmc10:
By the way, it sounds like the same thing can be achieved in ARM by doing:
```
__asm__ volatile("vmrs r0, fpscr\n"
                     "bic r0, $(1 << 24)\n"
                     "vmsr fpscr, r0" : : : "r0");
```
http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.ddi0344k/Chdfafia.html

To set the FZ bit in FPSCR to 0, it sounds like the FZ bit controls FTZ and DAZ. I don't have an ARM device setup though, so I would look for someone that could test on ARM to submit that fix

loganmc10:
@Zapeth Do you think FTZ should be disabled as well? The Intel documentation says:

> DAZ and FTZ flags are not compatible with the IEEE 754 standard, and should only be enabled when compliance to the IEEE standard is not required.

It seems like DAZ and FTZ are enabled by default, but they are actually SSE optimizations that are not compliant with IEEE 754, which the N64 is supposed to follow

Zapeth:
I don't know how big the (additional) performance penalty would be if its also disabled, so its hard to say.

There are some resources online that talk about the penalties, but in our case it heavily depends on the games that are emulated.

Are there any known games that use a lot of floating point operations? If so one could do a before/after test to see if there are any noticable/measurable performance hits (preferably on an older CPU and/or an AMD one).

Though looking at the N64 docs I guess the FTZ mode could be tied to the FS bit of the FCR31 register (since it seems to serve the same purpose)?

Fwiw, CEN64 seems to have both modes disabled.

loganmc10:
Yes it seems you are right again, this is what I found:
image

So it sounds like DAZ is disabled when the FS bit is set. I've updated the PR to do the same. FTZ seems a little more complicated, the behavior is tied to the rounding mode, I'm not sure exactly how to emulate that using SSE, or if it would be worth the performance impact

loganmc10:
Also for FWIW, the games I've tested either seem to set the FS bit permanently, or quickly alternate with it on/off. I'm not sure if there is a performance impact of enabling/disabling DAZ like that, I haven't really benchmarked this

Zapeth:
> So it sounds like DAZ is disabled when the FS bit is set. I've updated the PR to do the same. FTZ seems a little more complicated, the behavior is tied to the rounding mode, I'm not sure exactly how to emulate that using SSE, or if it would be worth the performance impact

I don't think thats correct. DAZ only applies to explicit (i.e. existing) denormal values, FTZ applies to computed ones.
The FS bit seems to just affect computed values as well, hence my suggestion to bind _FTZ_ mode to the FS bit.
It might not be completely equivalent in behavior in regards to rounding mode, but
1) its a better approximation than completely ignoring it
2) RN/RZ rounding modes are much more likely to be used than the other two anyway (assuming there are even any applications that rely on this)

Regarding quick alternation: thats again a topic where it highly depends on the application (how often is it alternated vs how many (denormalized) floating point operations are happening in between).
I haven't found any conclusive information about how "expensive" the mode switch is, but I don't think its much worse than a single (or a few) denormalized operation (disclaimer: this is pure speculation).

As for DAZ mode: its probably safer to just leave it disabled permanently. If it turns out to be a big performance hit then it could always be made optional (as well as FTZ), especially since so far only two games turned out to be (noticably) relying on this behavior.

loganmc10:
That's not really how I understand it, it says that the FS bit controls "whether a denormal number can be flashed". That sounds like DAZ/input to me. It says that whether the result is flushed to zero or not depends on the rounding mode. Like you say, RN/RZ is the likely scenario anyway, so it is probably safe to just leave FTZ enabled, which is the default.

I think it's worth emulating DAZ correctly, from my limited testing I don't see any major performance impact. You never know what other problems this might fix. Some games may rely on DAZ being enabled, so I don't think it would be safe to just disable it permanently

loganmc10:
Given the results of the test ROM you created, I'm inclined to believe you are correct. I've updated the PR to permanently disable DAZ mode, and only disable FTZ when the rounding mode is RP or RM, does that sound correct @Zapeth ?

Zapeth:
I'm not the one who can decide whether to accept or reject this PR so my approval doesn't really matter here, but it sounds close enough to what I had in mind in terms of behavior.

The end result is what matters anyway, so as long as the issues in the problematic roms are no longer present and it causes no other noticable issues elsewhere it should be safe to include.

Though apart from the missing implementation for non-SSE platforms I'd also recommend to test this with other systems/compilers (are other compilers than gcc and msvc even supported?).

I don't know what your system setup is, but I'd suggest at least tests with an (older) AMD CPU and other OSs like 32-bit Windows, Linux and macOS ([this test rom](https://github.com/PeterLemon/N64/files/4309135/CP1CEQ.zip) should give immediate results, but testing the games would be good too of course).

loganmc10:
> I'm not the one who can decide whether to accept or reject this PR so my approval doesn't really matter here, but it sounds close enough to what I had in mind in terms of behavior.

mupen64plus is light on reviewers/maintainers, people come and go (myself included), you seem to understand the problem, so your opinion is valuable.

I've tested on WIndows/Linux 32 and 64 bit using GCC, I don't have MSVC to test (the only Windows machine I have is a work computer, and I will get a talking to from the licensing people if I install the community edition on it)

src/device/r4300/cp1.h
```diff
@@ -25,6 +25,11 @@
 #include <stdint.h>
 #include "new_dynarec/new_dynarec.h"
 
+#if defined(__SSE__) || defined(_M_X64) || (_M_IX86_FP > 0)
+#include <immintrin.h>
+#define SSE
```
Narann: I think we need a better name than simply `SSE` that could clash with other constants and could be easier to search (as a reference to this particular problem). `M64P_SSE_NO_FTZ`?

src/device/r4300/r4300_core.c
```diff
@@ -121,6 +121,13 @@ void poweron_r4300(struct r4300_core* r4300)
 
 void run_r4300(struct r4300_core* r4300)
 {
+#ifdef SSE
+    //Save FTZ/DAZ mode
+    unsigned int daz = _MM_GET_DENORMALS_ZERO_MODE();
+    unsigned int ftz = _MM_GET_FLUSH_ZERO_MODE();
+    _MM_SET_DENORMALS_ZERO_MODE(_MM_DENORMALS_ZERO_OFF);
+#endif
```
Narann: I think a more complete comment would be best here: `//Save FTZ/DAZ mode because ...`.

projects/unix/Makefile
```diff
@@ -144,7 +144,7 @@ endif
 # base CFLAGS, LDLIBS, and LDFLAGS
 OPTFLAGS ?= -O3 -flto
 WARNFLAGS ?= -Wall
-CFLAGS += -ffast-math -fno-strict-aliasing -fvisibility=hidden -I../../src -I../../src/asm_defines -DM64P_PARALLEL
+CFLAGS += -fno-strict-aliasing -fvisibility=hidden -I../../src -I../../src/asm_defines -DM64P_PARALLEL
```
Narann: I _suspect_ we could find a dedicated flag fixing this problem: https://scicomp.stackexchange.com/questions/20867/what-does-ffast-math-do

Disabling `fast-math` disable a lot of thing, maybe one or two specific flags would be enough.

src/device/r4300/cp1.h
loganmc10: I would maybe change it to SSE_CAPABLE or something, the flag is not specific to FTZ or anything, it's just to determine whether SSE is supported or not.

projects/unix/Makefile
loganmc10: Yes:
> It enables -fno-trapping-math, -funsafe-math-optimizations, -ffinite-math-only, -fno-errno-math, -fno-signaling-nans, -fno-rounding-math, -fcx-limited-range and -fno-signed-zeros. Each of these flags violates IEEE in a different way

The N64 follows the IEEE standard as far as I can tell. For instance, -fno-rounding-math disables all rounding modes except "round to nearest", so all that code we have to switch between the 4 rounding modes is meaningless. I think case I think the NaN comparison was probably broken by -ffinite-math-only, but I think we should be emulating the N64 as close as possible, not fixing game specific issues 1-by-1 (when it can be avoided)

src/device/r4300/cp1.h
Narann: I get the idea, but I don't think it's a valid point. My problem is about the meaning :+1: 

Having a `SSE_CAPABLE` in the middle of the code (device/r4300/cp1.h) doesn't explain what is the purpose of the variable.

As the reason of the check is to fix a FTZ problem, I would keep this information on the variable name.

loganmc10: The purpose of the variable is just to determine whether SSE is supported or not, I guess that is where we see it different? It's unrelated to FTZ/DAZ

Narann: You are partially correct. The purpose on the variable is to determine whether SSE is supported or not, but because of the placement (globally, in a particular component, under an include intrinsic), its purpose must be specific. That's why I was suggesting a specific name.

Can I suggest to put this in `osal.h` as `OSAL_SSE` with OS specific check ?

loganmc10: Sure, I've updated it to use osal

Narann:
@loganmc10 suggest this to be part of 2.6.0.

It's a quite intrusive PR. So if anyone have a better idea on how implement this properly, feel free to pop in the discussion.

richard42:
It looks okay to me.  But do we need to change the savestate file format, to support the new flush_mode member that you've added to cp1?

loganmc10:
No I wouldn't think so, the struct isn't copied/read in it's entirety, the members are read one at a time, and ```update_x86_rounding_mode``` is called in the savestate loading code, and that function sets flush_mode.

PS the only thing this is missing is an ARM implementation, this solves the issue with the Interpreter, but someone would have to write something in ARM ASM to fix the issue there, I think this could still be merged either way though

Narann:
I've created a ticket for ARM implementation. I will merge for now but it could be great if we also fix this problem for ARM.

PR #734, opened 2020-03-21
Support FBWrites with a size of 3

loganmc10:
I noticed that Resident Evil 2 does this, triggering an "Unknown mask" warning

### DIFF
--- a/src/device/rcp/rdp/fb.c
+++ b/src/device/rcp/rdp/fb.c
@@ -162,6 +162,16 @@ void write_rdram_fb(void* opaque, uint32_t address, uint32_t value, uint32_t mas
         size = 2;
         break;
 
+    case 0xffffff00:
+        addr += (0 ^ Sh16);
+        size = 3;
+        break;
+
+    case 0x00ffffff:
+        addr += (1 ^ Sh16);
+        size = 3;
+        break;
+
     case 0xffffffff:
         addr += 0;
         size = 4;


bsmiles32:
Can't reproduce the warning on my end, while playing the first few minutes of the game.
Also a write of 3 bytes in a single transaction looks really odd to me as CPU only support byte, hword, word and dword memory operations.

loganmc10:
Really? You're running it with the interpreter? This is the trace I get on the first occurance, it happens right after the N64 logo, even before the main menu:
```
#1  0x00007fffcaf1a3bf in write_rdram_fb (opaque=0x7fffcfe7da18 <g_dev+17830328>, address=6545928, value=197634, mask=16777215) at ../../src/device/rcp/rdp/fb.c:172
#2  0x00007fffcaf17dbf in mem_write32 (handler=0x7fffcfe7f058 <g_dev+17836024>, address=6545928, value=197634, mask=16777215) at ../../src/device/memory/memory.h:87
#3  0x00007fffcaf18543 in r4300_write_aligned_word (r4300=0x7fffced7c860 <g_dev>, address=6545928, value=197634, mask=16777215) at ../../src/device/r4300/r4300_core.c:353
#4  0x00007fffcaef45f7 in cached_interp_SWL () at ../../src/device/r4300/mips_instructions.def:369
#5  0x00007fffcaf05fae in run_cached_interpreter (r4300=0x7fffced7c860 <g_dev>) at ../../src/device/r4300/cached_interp.c:1009
```

You can see the mask value is 16777215/0xFFFFFF

bsmiles32:
Sorry, I had DisableFBInfo = True in my config, that's why I didn't reproduce.
And indeed, I was also wrong about load/store sizes, because I had forgotten about SWL and other unaligned operations.

Tested and it works good :)

PR #735, opened 2020-03-22
PIF binary boot rom support

### LINKED ISSUE #730: Binary PIF ROM support
One feature I think would be cool to add to the emulator is the ability to execute the binary PIF ROM, instead of the HLE version.

This may help us find potential issues with the HLE PIF implementation.

I made a quick attempt at it here:
https://github.com/loganmc10/mupen64plus-core/tree/pifrom

This adds a new API call, M64CMD_PIF_OPEN, which allows the frontend to load pifdata.bin, very similar to the M64CMD_ROM_OPEN command.

It doesn't work right now, I tried with both interpreters, and the emulator stops at: ```Core Error: reserved opcode: BFC00014:4A4083C```. I don't really know enough about the PIF or R4300 execution in general to know what steps I might have missed.

@bsmiles32 is there something small I need to do to fix this? Or is this more work than I think?

loganmc10:
Fixes #730 

This allows booting using a binary PIF ROM (pifdata.bin) instead of the HLE implementation.

* Tested with old dynarec, and both interpreters. Not currently working with new dynarec. I'm not exactly sure how to get the new dynarec to start at 0xbfc00000 instead of 0xa4000040, or how to make that dynamic at runtime instead of a compiled address. @Gillou68310 

@bsmiles32 given your previous work and study on the PIF, I would really like to get your input if possible


loganmc10:
Here is the PR in ui-console that allows loading the PIF ROM file: https://github.com/mupen64plus/mupen64plus-ui-console/pull/58

loganmc10:
I think this would be OK to merge, it doesn't support the new dynarec, but it doesn't break it either, it's an optional new feature

m4xw:
@loganmc10 https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/new_dynarec/x64/linkage_x64.asm#L344

loganmc10:
@richard42 this is another I've included for most of the year without issue. It doesn't change the default behavior, but does add an optional API function that allows booting from a binary PIF

richard42:
Ok that's cool. Can you add one more thing first though? We should bump up the core API point version number and update the API versioning document Mupen64Plus_v2.0_API_Versioning

loganmc10:
@richard42 Just to make sure, do you mean ```FRONTEND_API_VERSION``` or ```MUPEN_CORE_VERSION```. I assume it's FRONTEND_API_VERSION, since the only addition was another command that can be called from the frontend?

loganmc10:
Yes I'm pretty sure that's right. I've upped the ```FRONTEND_API_VERSION``` version.

PR #737, opened 2020-03-24
BattleTanx (U) games need CountPerOp=3 to boot

loganmc10:
These games are very sensitive to interrupt timing, since https://github.com/mupen64plus/mupen64plus-core/commit/5a1491505f65f7bdef8930d03f9890f6d7992145 they aren't booting.

I don't think it's really the fault of that PR, it's just that the interrupt timing changed a bit and now these games don't like it.

For whatever reason, the (E) version doesn't have this issue


CallistoNTG:
I'm pretty sure setting CPO to 3 absolutely destroys performance in this game to basically unplayable levels.

loganmc10:
It's probably better than the game not booting at all though

PR #743, opened 2020-04-02
Rework flashram implementation

bsmiles32:
By combining the following sources :
- reverse engineering of Majoras Mask flashram functions [1]
- @sanni's cart reader [2]
- and MX29L1611 datasheet ("close enough" flash memory from Macronix)

I was able to figure a better model for flashram than what we currently
have. As this new modeling of flashram doesn't match one-to-one with
old model, meaning it may possibly "break" older savestate (in fact it may break if the savestate has been taken during some flashram operation, otherwise it should be relatively safe),
I had to bump savestate version.

[1] pending PR in https://github.com/n64decomp/majora/
[2] https://github.com/sanni/cartreader/blob/master/Cart_Reader/N64.ino

I've tested mostly Majora's Mask, Jet Force Gemini and Derby Stallion and it seems to work OK.
If you want to test also, **be sure to backup your fla saves first**, so that you won't loose them, in case of a bug...

cc: @sanni. This work may (or may not) interest you. Feel free to comment or give suggestions if you want :)

### DIFF
--- a/src/device/cart/cart.c
+++ b/src/device/cart/cart.c
@@ -124,7 +124,7 @@ void init_cart(struct cart* cart,
 
     init_flashram(&cart->flashram,
         flashram_type,
-        flashram_storage, iflashram_storage, dram);
+        flashram_storage, iflashram_storage);
 
     init_sram(&cart->sram,
         sram_storage, isram_storage);
@@ -161,7 +161,7 @@ void read_cart_dom2(void* opaque, uint32_t address, uint32_t* value)
         }
 
         cart->use_flashram = 1;
-        read_flashram_status(&cart->flashram, address, value);
+        read_flashram(&cart->flashram, address, value);
     }
 }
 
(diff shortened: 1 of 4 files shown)


loganmc10:
Pretty cool, I tested Paper Mario and Command & Conquer and didn't see any obvious issues.

I also just wanted to point out that https://github.com/mupen64plus/mupen64plus-core/pull/721 is in the PR queue and also proproses a savestate version increase to 1.7

i30817:
Does this break runahead if started on boot, like happens in many other emulators if you can savestate on boot?

loganmc10:
@bsmiles32 I think this is really valuable work. I think if you can fix the merge conflict it should probably be merged

bsmiles32:
Rebased against master and resolved savestate conflicts as suggested by @loganmc10

bsmiles32:
I'm not sure this PR should be merged now... I have doubts about page_buf belonging to flashram (not clear from the datasheet) and I suspect the buffer belongs to the PI which is known to have such 128 byte buffer. This 128 byte PI buffer is what causes the write_cart_rom issue we found earlier.
To be clear, I think the code works OK, but I'd like to revisit it and see if I should move page_buf to PI instead. This would avoid breaking/bumping savestate version later.

bsmiles32:
@richard42 Are you OK to merge this ?

richard42:
yes, please go ahead

PR #746, opened 2020-04-04
Fix hard reset crash

bsmiles32:
Integrate work done by @flagrama


loganmc10:
@bsmiles32 we are now getting this AppVeyor failure:

```
..\..\src\device\r4300\interrupt.c(483): error C2039: 'save_rsp': is not a member of 'recomp' [
```

I think just on the Win32 build

Issue #112, opened 2015-03-27
Network play


ymartel06:
Hi,

To begin, thanks everyone for your great work on Mupen64plus, it's awesome.

I'm working to make the netplay for mupen64plus and I posted a message in google code few days ago.
https://code.google.com/p/mupen64plus/issues/detail?id=16&colspec=ID%20Type%20Component%20Status%20Priority%20Stars%20Milestone%20Owner%20Summary
It's my bad, you moved on github :)

I'm aiming to have the netplay for maximum 4 network players with all combinations available. (Multiple locals vs network etc...)

This morning I pushed a version with the support of 2 players (One by computer)
https://github.com/ymartel06/mupen64plus-core/commit/6fe81b909bf13dd4abbf611b20d615748290001d

To test it, get my console https://github.com/ymartel06/mupen64plus-ui-console and my core https://github.com/ymartel06/mupen64plus-core/.
Then you can launch the server with:
mupen64plus-ui-console.exe --server your-rom
For the client:
mupen64plus-ui-console.exe --client IP your-rom

Currently, it's only TCP (UDP will come really later) and the default is 6464 (you can change it using --server-port).

Before to work on the 4 players, I will work on:
* Fix the clock to take care of CLOCKS_PER_SEC for the client. I use a clock sync to be able to launch the emulator at around the same time.
* Don't allow to press pause, forward, backyard and more, when the emulator is in network
* Try to build it on Linux / MacOS and fix what is needed (I try to make everything cross platform but I can have a bad surprise)
* Add an option to force the frame sync between players (good to have the same simulation but can be laggy, looks like this concept http://www.gamasutra.com/view/feature/131503/1500_archers_on_a_288_network_.php)

Known issue:
* When you move the render window, you can have some disconnection.

Of course feedbacks are welcome. If you want some explanations of what I made (like flow or technical), I will try my best to answer.

Note: It seems I can't open a discussion.

tony971:
This could be good with 64DD emulation if mupen ever adds that functionality. I know that cen64 and MESS have made huge leaps in it.

wareya:
Time travel / rollback when?

(I jest. Even desync-free low lag emu netplay is an accomplishment at this point!)

Narann:
Hello @ymartel06! Nice to see you here! I encourage peoples to test and feedback on this precious development netplay is.

richard42:
Yeah this is great to see.  Users have been wanting netplay for a very long time.

[...]

PR #753, opened 2020-04-26
Netplay

loganmc10:
For starters, you can see a demo here:

https://www.youtube.com/watch?v=bkW_SbxdsUo

Here is the code for the server I wrote, the intention is that the netplay server wouldn't be part of the mupen64plus project, but anyone could write one based on the spec:
https://github.com/loganmc10/m64p-netplay-server

Here are the changes I had to make to mupen64plus-gui:
https://github.com/m64p/mupen64plus-gui/tree/netplay

I am submitting this in a WIP state in the hopes that I can get comments on the API and code, and any other philosophical discussions about netplay that need to happen. I am on Discord (https://discord.gg/8XRuJ4F) in the #emu_mupen channel as well, which might be a better place for longer back and forth discussion.

Some principles of this design:
* The server is the "source of truth", when Player 1 hits A, their local game does not register an A press, it sends that info to the server, and the server responds to all players with data on when to register an A press for player 1
* Key inputs are not guaranteed to be received by the server (all data is transferred via UDP). This is to keep input latency low
* The game will "punish" the player with a poor internet connection (they may experience higher input latency and missed input, but it won't affect the other players). I think this is an important principle in online gaming

The implementation was written in SDL_net, I haven't documented the all the packet formats, but I can do all that documentation work once there is a general consensus on the design

Responsibilities:
* Core: sends and receives key inputs directly from the netplay server, keeps the game running in sync with other players
* Frontend: ensures that each player is running the same plugins/versions/settings via communication with the netplay server (I have not implemented this yet)
* Netplay server: communicates with core and frontend to accomplish what was listed above

I'm sure I missed many details, please ask me any questions you'd like


src/main/netplay.c
```diff
@@ -0,0 +1,339 @@
...
+}
+
+int netplay_register_player(uint8_t player)
+{
+    UDPpacket *requestPacket = SDLNet_AllocPacket(6);
+    UDPpacket *responsePacket = SDLNet_AllocPacket(3);
+
+    requestPacket->data[0] = 4; //Packet type, 4 = player registration
+    requestPacket->data[1] = player;
+    SDLNet_Write32(random_number(), &requestPacket->data[2]);
```
Narann, line 297: I'm not sure to get the idea here. If this number is supposed to match between clients, why not simply use an incremented value? Using rand number could fail on client using various architectures. Am I wrong?

src/api/frontend.c
```diff
@@ -304,6 +309,27 @@ EXPORT m64p_error CALL CoreDoCommand(m64p_command Command, int ParamInt, void *P
...
+            else if (response == -1)
+                return M64ERR_SYSTEM_FAIL; // no response from server
+            else
+                return M64ERR_INPUT_INVALID; // player already in use
+        case M64CMD_NETPLAY_GET_VERSION:
+            memcpy(ParamPtr, git_version, 41);
+            if (ParamInt == NETPLAY_API_VERSION)
+                return M64ERR_SUCCESS;
+            else
+                return M64ERR_INCOMPATIBLE;
```
Narann, line 330: I'm not sure using git_version for this is a good idea. Some distributions could have security patches or anything that could make this fail.

Narann:
The code is pretty clean. There is still a lot of magic numbers, but as you say, it's not the purpose yet. Just the versioning query using `GIT_VER` looks weird to me.

I love your server to clients design.

Good job! This was a long time requested feature!

Mastergatto:
> Frontend: ensures that each player is running the same plugins/versions/settings via communication with the netplay server (I have not implemented this yet)

Don't forget the cheats! It would be annoying if one is able to use them while others don't. Thanks and keep up the good job!

Also, tracking https://github.com/mupen64plus/mupen64plus-core/issues/112

src/main/netplay.c
loganmc10: Before the game starts, each client needs to register 1 (or more) players to play as. If they don't register any players, they will be a spectator when the game starts. The random number is just a client ID/registration token.

The client sends a message to the server-> "Client (random number) would like to be player 3". Because it is all in UDP, I have to retransmit that message every now and then, using the same random number/client ID. The server will respond with either "yes" or "no" if the client can become that player. Hopefully that makes sense.

I just picked a random number because it was the simplest to implement. Other methods would be to use a MAC address, or some kind of computer ID, but I didn't see any clean, cross platform way to do those things.

src/api/frontend.c
loganmc10: I needed some way to verify that each client is using the exact same version of the core. Just checking for "version 2.5.9" won't cut it unfortunately.

This does mean that it would be almost impossible for somebody that just pulls mupen64plus from the Arch Linux repo to play with someone that got their mupen64plus from somewhere else, but that is how it should be, the clients will likely desync if they aren't running the exact same version.

loganmc10:
I thought git version would be good because it's fairly automatic, even something like a change to mupen64plus.ini to change the default CountPerOp for a game would bring players out of sync if they are playing that game, so almost every PR would require an increment to the version number.

I think it would be better to do something like "if GIT_VER" and then fallback to MUPEN_CORE_VERSION if the compiler is unable to obtain the GIT_VER

src/main/netplay.c
Narann: I just get it. I was thinking this was called all the time. All you say make sense, my bad.

Narann:
The git version does not provides guarantee `mupen64plus.ini` is the same.

If you need to rely on some information to avoid out of sync, a more robust system need to be found. My suggestion is to keep an internal netplay version for modifications that change the internal behavior (for server purpose), what you consider as changing the sync (core version, ROM hash, CountPerOp, etc), hash all of this and send the hash. If all players don't send the same hash to the server, it's over.

About cheaters: It's open source, anyone with some knowledge could beat the system anyway.

loganmc10:
Alright, I haven't put a ton of thought into keeping all the settings in sync yet, that's probably the last thing I'll work on (as it's the least exciting..). For now I've updated it to just use a "netplay_core_version" variable that can be incremented when changes to the core are made.

And you are correct about cheaters, it would be trivial to break this is any number of ways, it's just designed for casual friendly play

loganmc10:
I've gotten it to a point where I think it can be tested, you can download a build here:

https://github.com/loganmc10/m64p/releases/tag/netplayb1

Here is another demo showing some of the progress:

https://www.youtube.com/watch?v=XKehKifbK8U

Narann:
It's very nice ! I realize one PC seems to always beeing 1 frame before. I wonder if the system couldn't have a "friendly delay" option (aka, minimun delay of the other players / 2). So if you have this :

* Player 1: 10 ms
* Player 2: 50 ms
* Player 3: 53 ms
* Player 4: 75 ms

The best player (here, player 1) would be slighly delayed to better match the other best players ping : 50 - 10 = 40. 40 / 2 = 20:

* Player 1: 30 ms (10ms + friendly delay: 20ms)
* Player 2: 50 ms
* Player 3: 53 ms
* Player 4: 75 ms

As you say, better connection players should not be penalized instead of slow connection ones. I agree and it makes perfect sense for outside server (aka, both players connect to external server, like Steam).

But if, in practice, there is always the combo: Player 1 is both server and player, and player 2 (and 3, and 4) connects to player 1 PC. Player 1 will *always* have a better ping. This could ruin the game, even for player 1.

So here is my suggestion for a "friendly delay" option on server side : If one player is on the same network, its delay is compared with others and if it's better, he gets an delay to be a little bit consistent (aka, on par) with other players so everyone have fun.

Anyway it will always have a better delay than others, we just avoid a huge gap.

loganmc10:
I spent all this time trying to reduce the latency and now you want me to put it back? 😛 

In the initial testing I did with someone on Discord (not sure if he has a GitHub account), the latency differential didn't seem to negatively affect the gameplay. He had a ping of about 80-90ms and mine was about 35-40ms. We played a fairly long round of Perfect Dark, and some Mario Kart, the controls felt responsive for both of us.

I wouldn't rule it out completely, but I don't think it would be high on my list of things to implement, first I've got to iron out the bugs and do some work on syncing settings and such. To me, latency differentials are just a part of online gaming, a player with a higher latency would have a (small) disadvantage playing Call of Duty or other modern online games (if I understand how they work correctly)

loganmc10:
Keep in mind, the way I've written this is very client -> server oriented, with the idea that the server would be running somewhere on the Internet.

If a player is running the server locally, you are correct that they will probably have fairly significant difference in input latency, but that difference is reduced quite a bit when all players are connected to an Internet hosted server that is in roughly the same region as the players

Narann:
Of course, my suggestion was only for "local server" that I thought it would be the most common use case, but maybe I'm just wrong.

loganmc10:
> Of course, my suggestion was only for "local server" that I thought it would be the most common use case, but maybe I'm just wrong.

No I'm sure it's valid, I guess what I would say is that the server implementation I'm writing at https://github.com/loganmc10/m64p-netplay-server is meant to be hosted on the Internet. Anyone else could write their own implementation, or integrate one into ui-console, and the API in the core should allow the server to delay players as required.

> Any hope to get the vcxproj file fixed?

Yes for sure, the first step would be to add SDL2_net to win32-deps, are you able to merge this PR:
https://github.com/mupen64plus/mupen64plus-win32-deps/pull/11

Narann:
MSVC raise some errors:
```
..\..\src\main\netplay.c(372): error C2057: expected constant expression [C:\projects\mupen64plus-core\projects\msvc\mupen64plus-core.vcxproj]
..\..\src\main\netplay.c(372): error C2466: cannot allocate an array of constant size 0 [C:\projects\mupen64plus-core\projects\msvc\mupen64plus-core.vcxproj]
..\..\src\main\netplay.c(372): error C2133: 'output_data': unknown size [C:\projects\mupen64plus-core\projects\msvc\mupen64plus-core.vcxproj]
..\..\src\main\netplay.c(406): warning C4018: '<': signed/unsigned mismatch [C:\projects\mupen64plus-core\projects\msvc\mupen64plus-core.vcxproj]
..\..\src\main\netplay.c(407): error C2036: 'void *': unknown size [C:\projects\mupen64plus-core\projects\msvc\mupen64plus-core.vcxproj]
..\..\src\main\netplay.c(409): error C2057: expected constant expression [C:\projects\mupen64plus-core\projects\msvc\mupen64plus-core.vcxproj]
..\..\src\main\netplay.c(409): error C2466: cannot allocate an array of constant size 0 [C:\projects\mupen64plus-core\projects\msvc\mupen64plus-core.vcxproj]
..\..\src\main\netplay.c(409): error C2133: 'temp': unknown size [C:\projects\mupen64plus-core\projects\msvc\mupen64plus-core.vcxproj]
```

loganmc10:
Alright AppVeyor build is all fixed up :)

Narann:
Thanks. I think we should use `#define` constants instead of magic numbers (for `l_status` member).

The other discussion would be: If we merge this, We should put the [server](https://github.com/loganmc10/m64p-netplay-server) repo under the mupen64plus organization. Should I open a dedicated issue on this repo to talk about it?

loganmc10:
Yes my last step will be documentation and cleaning up the code (including defining the magic constants).

The server is written in Qt, and is designed to interact with https://github.com/m64p/mupen64plus-gui, which is also not part of mainstream mupen64plus. All the match making/lobby code is specific to that frontend. I think a proper "upstream" server would be written in C and designed to work with ui-console, it may be a bit simpler than the server I wrote.

Anyway I don't know if the server would really fit as an upstream project, but I'll definitely spend a good bit of time on documentation so if someone else wanted to write another server it wouldn't be too difficult

loganmc10:
It's getting to be fairly fully featured now, a reminder you can download it here:

https://github.com/loganmc10/m64p/releases

Some cool features it now supports: save syncing, core settings syncing, and raw input (Raphnet) support

loganmc10:
I've added comments in the code, and some documentation. I've been testing with a few people and the results have been quite promising. I think this is ready to start being considered for merging

src/api/frontend.c
```diff
@@ -304,6 +305,24 @@ EXPORT m64p_error CALL CoreDoCommand(m64p_command Command, int ParamInt, void *P
                 return M64ERR_INPUT_INVALID;
             g_media_loader = *(m64p_media_loader*)ParamPtr;
             return M64ERR_SUCCESS;
+        case M64CMD_NETPLAY_INIT:
+            return netplay_start(ParamPtr, ParamInt);
+        case M64CMD_NETPLAY_CONTROL_PLAYER:
+            if (netplay_register_player(ParamInt - 1, Controls[netplay_next_controller()].Plugin, Controls[netplay_next_controller()].RawData, *(uint32_t*)ParamPtr))
```
Narann, line 311: I suggest to add parameter check before calling function:
```c
if (ParamPtr == NULL)
    return M64ERR_INPUT_ASSERT;
if (ParamInt < 1 || ParamInt > 4)
    return M64ERR_INPUT_INVALID;
if (netplay_register_player(...
```

src/main/netplay.c
```diff
@@ -0,0 +1,602 @@
...
+}
+
+m64p_error netplay_stop()
+{
+    if (l_udpSocket == NULL)
+        return M64ERR_INVALID_STATE;
+    else
+    {
+        char output_data[5];
+        output_data[0] = 7; //Notify server that we are disconnecting
```
Narann, line 107: I still think we should use named constants instead of magic numbers. At least for packet type.

loganmc10:
Yes good feedback, I've made those changes. I also made it so building with netplay is optional (enabled with NETPLAY=1), since I know the SDL2_net dependency would be a pain for some downstream projects

src/main/netplay.c
```diff
@@ -0,0 +1,618 @@
...
+{
+    //This function is used to check if games have desynced
+    //Every 60 VIs, it sends the value of the CP0 registers to the server
+    //The server will compare the values, and update the status byte if it detects a desync
+    if (!netplay_is_init())
+        return;
+
+    const uint32_t* cp0_regs = r4300_cp0_regs(cp0);
+
+    if (l_vi_counter % 60 == 0)
```
Narann, line 511: Is `60` an universal value? Does it work for PAL games?

Narann:
I thinks it's nice. I wonder if packet sizes everywhere shouldn't use constants. Like `char output_data[PACKET_SIZE_PLAYER_OUTPUT]`. I've identified three or four of those *size numbers* in the code.

Buffer sizes are always sensitive variables that could benefit to be write once with a unique meaning.

For the rest, it looks OK to me.

src/main/netplay.c
loganmc10: Yes it'll work with PAL, it could be any number really, it's just there to make it transmit a desync check roughly every second.

loganmc10:
I think the sizes are probably OK as is? I did write this document as well: https://github.com/mupen64plus/mupen64plus-core/blob/3d6a823082892aaf54230a8248c938fe5ef9af9b/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Netplay-API.mediawiki

Which explains each packet, it's size, and what each byte is for

Narann:
@richard42 any idea?

richard42:
I haven't really looked over the code super closely but I've been following the conversation and seeing the updates. I see that you're using a NETPLAY_API version number instead of relying on the git repo commit GUID. That's good, as we discussed previously. Going forward we will need to remember to bump that version number whenever anything with the core changes.

loganmc10:
I just pushed the last change I was planning. From the testing I've seen and participated in, it seems to be working very well. I don't plan on making any further modifications, I think it is stable enough to be merged if everyone is happy with the code

src/backends/file_storage.c
```diff
@@ -42,7 +43,14 @@ int open_file_storage(struct file_storage* fstorage, size_t size, const char* fi
     }
 
     /* try to load storage file content */
-    return read_from_file(fstorage->filename, fstorage->data, fstorage->size);
+    if (!netplay_is_init())
```
bsmiles32, line 46: You should not modify backends like that, but instead provide another backend (netplay_storage_backend ?) and configure the core to use (in main_run()) it when in netplay mode.

src/backends/plugins_compat/input_plugin_compat.c
```diff
@@ -60,8 +61,28 @@ static m64p_error input_plugin_get_input(void* opaque, uint32_t* input_)
     int pak_change_requested = 0;
 
     /* first poll controller */
-    if (input.getKeys) {
-        input.getKeys(cin_compat->control_id, &keys);
+    if (!netplay_is_init())
```
bsmiles32, line 64: Same here, you should not modify input backends with netplay-specific code, but instead provide a new netplay input backend and make the core use it at core startup (main_run()). This new netplay input backend may rely the non-netplay input backends to get user input.

src/device/pif/pif.c
```diff
@@ -371,6 +372,12 @@ void update_pif_ram(struct pif* pif)
         input.readController(-1, NULL);
     }
 
+    if (netplay_is_init())
```
bsmiles32, line 375: This should go inside netplay input plugin. You may have to extend the input backend API (or use the readController(-1, NULL) trick).

src/device/rcp/vi/vi_controller.c
```diff
@@ -157,6 +158,9 @@ void write_vi_regs(void* opaque, uint32_t address, uint32_t value, uint32_t mask
 void vi_vertical_interrupt_event(void* opaque)
 {
     struct vi_controller* vi = (struct vi_controller*)opaque;
+
+    netplay_check_sync(&vi->mi->r4300->cp0);
```
bsmiles32, line 162: This code should go inside new_vi() function so that netplay code doesn't mix with n64 emulation code.

loganmc10: The reason I put it here is because it needs to read the CP0 register values, if I did it in new_vi, I would need to read the cp0 values using the global g_dev variable I would think, which method is preferable?

bsmiles32: I'd say using g_dev inside new_vi is better here because of better separation of concerns netplay vs emulation.

src/device/pif/pif.c
loganmc10: I'm not really sure how to do this. update_pif_ram calls ```input.readController(-1, NULL)```. This is the point that a raw input plugin like Raphnet updates the PIF ram. I need to intercept that data before update_pif_ram exits

bsmiles32: Probably in main_run you can store the value of input.readController, replace it with one you provides which calls the original input.reacController and adds your netplay logic after. Something like:
```
void netplay_read_controller(int control, unsigned char* command)
{
    original_read_controller(control, command);
   
    netplay_send_raw_input(&g_dev.pif);
    netplay_get_raw_input(&g_dev.pif);
}

// maybe inside main_run ? or somewhere after input plugin setup ?
...
if (netplay_is_init()) {
  original_read_controller = input.readController;
  input.readController = netplay_read_controller;
}
```

loganmc10: I think at that point (hijacking the call to input.readController), it's getting a little hacky just to remove a few lines of code. I consolidated the functions inside update_pif_ram to a single ```netplay_update_input``` function, so it's just a single line of code, similar to the new_vi call in the vi_controller.

richard42:
Okay @loganmc10 it looks pretty good.  If you can add documentation for the network protocol, then we'll merge it.

loganmc10:
@richard42 I think that is already done, unless I misunderstand what you're asking for, the packet formats are documented here:

https://github.com/loganmc10/mupen64plus-core/blob/netplay2/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Netplay-API.mediawiki

aacalfa:
Hello, I’m new to this so pardon me for lack of knowledge of things. I have retroarch on Xbox which has the core Mupen64Plus-Next 2.4-GLES3 installed. RetroArch says this core doesn’t support Netplay, but looking at this thread it leads me to think that it is supported. Could someone help explain what is the issue?

PR #762, opened 2020-06-12
[Disk] Major Disk Support Changes (not ready, design choice discussion)

LuigiBlood:
This is more in line with Project64's 64DD support. This isn't quite clean just yet.

- Change disk seek to support more formats without conversions
  - Added basic Disk library functions for calculating seek offset.
- System Area heuristics for finding appropriate system formatting data to use as base.
- Support Development Disks.
- Support USA and DEV 64DD IPLs.
- Do read errors for System Area / Disk ID if the first block doesn't work.
- Support D64 disk file format.

Some design choices:
- Added `void* extra` to `file_storage` struct to include file information, in this case, for `extra_storage_disk` struct.
  - I consider this information needs to be understood with very early on under the disk integrity check process and shouldn't be under dd_controller.
  - Considering I'm touching a pretty important backend, I don't expect it to break anything, but maybe save states are impacted? No idea.

**EDIT:** This was removed by @bsmiles32 as it is not its purpose to get format information.

TODO:
- Support configuration for Full Disk Copy or RAM Only Save (only the RAM save area of the disk).
- Disk Seek Times
- Disk Swapping
- Cleanup


LuigiBlood:
It looks like Doshin The Giant 1 and expansion isn't working even without my changes.

LuigiBlood:
I now confirm there were changes in mupen64plus that completely broke some 64DD games.

LuigiBlood:
So the issue is timing, in which case CountPerOp=1 might be really imperative for games to work with 64DD.

Narann:
IIRC, @loganmc10 did the change, maybe he can review this part of the code.

LuigiBlood:
We already spoke about it on Discord. I've done some testing, and while / 50 is a good choice for F-Zero X, it's not for the rest of the games where it must be / 25 for some reason despite CountPerOp being 2 by default. It's honestly confusing timing at this point.
To fix F-Zero X Expansion Kit you need to add CountPerOp=1 to the settings.

loganmc10:
Yeah I trust @LuigiBlood's judgement on this stuff

LuigiBlood:
I have a question about this, is the code of this, so far, good?
I had done some serious changes, like how the disk file is not dependent on a format anymore, using code reverse engineered from libleo (the job was done before the gigaleak, easily verifiable in my 64dd repo), I did a change for file_storage.

LuigiBlood:
@richard42 This is above all your project, I really want to finish this. I've done changes that may break some API stuff like file_storage because mupen64plus just isn't made for 64DD in mind.

bsmiles32:
Sorry it took so long to review your work.
I've made a PR (https://github.com/LuigiBlood/mupen64plus-core/pull/1) to your PR to change things that I felt necessary, most importantly I didn't really like the fact that you had to add the extra field to the storage_backend_interface.

The idea behind storage_backend_interface was to abstract storage accesses from emulation code. Emulation code should only care about reading/writing data like on normal hardware, not care about if I put data in memory, or into a file (or in a memory-mapped file). So adding "format" information at the storage_backend_interface felt wrong to me, that's why I've removed them from the interface and pass them as regular parameters to the dd_controller.

If you want we can go a little further and create a dd_disk struct to regroup all disk related stuff for instance:
```
struct dd_disk {
    void* storage;
    const struct storage_backend_interface* istorage;
    uint16_t lba_phys_table[0x10DC];
    uint8_t format;
    uint8_t development;
    size_t offset_sys;
    size_t offset_id;
};
```
This would maybe help readability.

In my PR I've also fixed some warnings (some pointer conversion, and printf formats).

What do you think ?

LuigiBlood:
That's fine. I wanted to try my hand on doing what I've done for Project64 but I'm finally happy to finally have a response on my work.
At this point what I need to deal with is the save process, because that's really a needed feature.

bsmiles32:
@LuigiBlood Take a look at https://github.com/bsmiles32/mupen64plus-core/tree/diskupdate where I've reworked things to handle disk format differences through a specific backend. So you can access disk->format and other fileds inside the storage save method.
It has not been tested, but it should be good enough for now.

LuigiBlood:
That totally works for me.

What I intend to do after that when loading the save there are two formats:
- \*.ndr/\*.d4r (basically a replacement file like you've supported)
- \*.ram (only RAM Area content, SDK and D64 format only)

The first format I just basically try to load it instead of the main disk file (it is a bit weird but it works), and then the *.ram file I just load the file into the loaded disk file and replace the save data in it.
I would probably add a new configuration that says what format to save as overall.

LuigiBlood:
It crashes upon loading and Visual Studio doesn't feel like loading the PDB file and I don't know why so I'm not sure where is the problem

bsmiles32:
I've forced-pushed a fix to a crash I was having with my previous attempt.
I've tried to test with F-Zero (J) + DD disk and it gives loading errors I wasn't having before ?
Did you encountered such problems ?

bsmiles32:
OK, seems like the issue with F-Zero is related to PI timings (I saw you changed cycles computation) and that's what makes F-Zero not load anymore. Other titles exhibit the same kind of warnings, but mostly run OK given the PI timing change.

Several monthes ago I was investigating the possibility of changing the way PI cycles are counted. I didn't succeed in obtaining accurate cycle counting, but at least it was a shot in the "right" direction. I'll try to gather what I had back then and publish some of that.

LuigiBlood:
05000580 is the DD MSEQ RAM... it's pretty useless to emulate for the most part and could be ignored entirely
`Sector size 208 set different than expected 232` maybe you used like a development disk in which case this kind of error may happen but it's normal

LuigiBlood:
Just to make sure it does actually notify: With the current code, F-Zero X Expansion Kit works fine if F-Zero X (J / U) have CountPerOp=1 in the database.

When it comes to the warnings like `Sector size 208 set different than expected 232` I think those warnings are useless because I changed the way it manages sectors. I may have forgotten to move / delete those.

bsmiles32:
OK I've made a PR on your PR so you can benefit from storage_backend rework.

Regarding F-Zero X, indeed, with CountPerOp=1 it loads fine. The thing is, DD games seems to be way more sensitive to timings and we know that that's not were mupen64plus shines... Anyway, the core issue here is that we need better PI and CPU timings. For PI timings, I have some preliminary work in the pi_timings branch, which is not perfect but should be better than what we currently have. For CPU timings, well, the CountPerOp Hack may or may not work and will always be error-prone, but for now that's all we have (and changing it would require massive effort especially with the dynarecs unfortunately).

For the sector size warnings, I'm not knowledgeable enough to tell if they're expected and/or useful. So I'll trust you on that.

LuigiBlood:
So it does not save yet but it can load disk save files depending on the added configuration of "SaveDiskFormat".
0 = Full Disk Copy (*.save), 1 = RAM Area Only (*.ram).

Tested on both NDD (SDK format) and D64 files. I admit to not really test much for MAME formats because it's genuinely not as practical.

LuigiBlood:
TODO: Saving support, refactor disk format stuff into dd_disk.c

LuigiBlood:
So there was a bit of a cleanup and refactoring that's nice.
Looking at the TODO list:
- Support configuration for Full Disk Copy or RAM Only Save (only the RAM save area of the disk).
  - So the configuration exists, but the saving process is not.
- Disk Seek Times
  - On Project64 I rely on really bad math and uses of timers to then issue a CART interrupt, I don't know how this could be managed on mupen64plus at the moment.
- Disk Swapping
  - Requires API update for this? Can Media Loader have the ability to swap content?
- Cleanup
  - That's being sorta done.

I may just add this question though:
About the ability to give information to plugins on what game is loaded...
To not break compatibility, in Project64, I've generated a N64 ROM header from the data and it takes priority from the N64 ROM (even if it's a combo load, so F-Zero X Expansion Kit disk info would be returned instead of F-Zero X ROM info).

Is it fine if I do exactly this?

bsmiles32:
Thanks for the summary and plan of action.
I think we should split your plan and try to merge stable bits as soon as they're stable.
For instance if you think this PR is stable enough, we should not wait too much to merge it.
This will allow to ease testing and reviewing (smaller PR are easier to handle).

> Looking at the TODO list:
>
>     Support configuration for Full Disk Copy or RAM Only Save (only the RAM save area of the disk).
>         So the configuration exists, but the saving process is not.

What's missing here : 
- uncommenting the call to istorage->save in write_sector
- implementing both saving options in the disk save method
- ?
Saving the full content of disk every time istorage->save is called (e.g. each time write_sector is called) is going to be problematic.
What we need is to extend the istorage->save method to also specify the offset/size of bytes to write. Can be something like :
```
void save(void* storage, unsigned int offset, size_t size)
```
Hopefully this should reduce the amount of data to write at write_sector.


>     Disk Seek Times
>         On Project64 I rely on really bad math and uses of timers to then issue a CART interrupt, I don't know how this could be managed on mupen64plus at the moment.

Not really sure how to proceed here as I don't really know how it works at the hardware level. Does the drive emit a CART interrupt when the head has reached its target position ?
Will require more investigation.

>     Disk Swapping
>         Requires API update for this? Can Media Loader have the ability to swap content?

The media loader allows you to ask the frontend for a new disk filename (just a filename).
Disk swapping is then done as follow:
- unload current disk
- ask frontend for new disk filename via media_loader API
- load the new disk
- resume execution

You may need to add a new key to the core to trigger such "disk swap" action.

>     Cleanup
>         That's being sorta done.
>
> I may just add this question though:
> About the ability to give information to plugins on what game is loaded...
> To not break compatibility, in Project64, I've generated a N64 ROM header from the disk data and it takes priority from the N64 ROM (even if it's a combo load, so F-Zero X Expansion Kit disk info would be returned instead of F-Zero X ROM info).
>
>Is it fine if I do exactly this?

I guess this is a good enough solution.

LuigiBlood:
>I think we should split your plan and try to merge stable bits as soon as they're stable.
For instance if you think this PR is stable enough, we should not wait too much to merge it.
This will allow to ease testing and reviewing (smaller PR are easier to handle).

I suggest doing the saving part properly and then merge this PR, and figure out the rest later (Seek Times, Swapping, and Plugin recognition).

LuigiBlood:
Actually just merge this and we'll think about saving afterwards.

bsmiles32:
OK for merging without the saving part as it needs a bit more work on the storage backend (I'll do the part to allow to save files in chunks in a latter PR).
But before merging I'd like to refactor/clean up the load_dd_disk function (move format specific parts inside disk.c) as it still causes some warnings on my compiler and is very big and difficult to read.
I hope to have that done and merged before the end of the week-end.

LuigiBlood:
Just as a short explanation of load_dd_disk():
- Attempt to load *.save disk file first (if Disk Save Format is Full Copy [0])
- Load the real disk file if it failed
- Checking if the System Data info is valid on LBA 0, 1, 2, 3, 8, 9, 10, 11
  - Checks individual data if it's valid or not.
  - Checks if the format is not D64 that the System Data sector repeats 85 times.
  - This is in case the disk file is actually corrupted as the format does not say what blocks are unreadable or not.
  - If the data is valid, then don't check the next LBA because libleo will not attempt to read them.
    - So if LBA 0 (or 2 if dev disk) is valid, that's good enough.
    - The valid System Data LBA is also kept in memory so that the sector seek can actually return a read error if it ever attempts to read the bad ones.
- If the format is MAME or SDK:
  - Check if Disk ID sector (LBA 14 & 15) repeats 85 times in case it is corrupted.
  - The valid Disk ID LBA is also kept in memory for the same reasons as the System Data.
- If the format is D64:
  - We check if the filesize is correct as it can be calculated using the info in the System Data.
  - If correct, then expand the RAM Area to full capacity and change the System Data info accordingly.
- Input the disk information in dd_disk variables.
  - `offset_ram` is not applicable in MAME format due to the ordering of the blocks themselves.
- Attempt to load *.ram file (if Disk Save Format is RAM Area Only [1]) and replace the data of the RAM area with its contents.

bsmiles32:
Thanks for the detailed explanation !
Few questions/remarks :
- I initially put the "*.save"  extension, more as an easy way to prevent corrupting my dumps while working on the 64DD, but I didn't look what were the proper "standard" formats/extensions. I saw you mentioned *.ndr and *.d6r being extensions proper extension for ndd and d64 savefiles. Shouldn't we support that extensions instead  as they're more standard, or at least already in use with PJ64 ?
- We currently don't put the savefile in the standard mupen64plus save file directory. I'll try to also update that part to look in standard mupen64plus save folder.

I'll keep you updated on my progress.

LuigiBlood:
*.ndr / *.d6r for Full Copy Disk saves, and *.ram for RAM Area Only (only applicable for SDK and D64).

bsmiles32:
Another quick question :
- Do we have to handle byteswapped format ? Do such things exists in existing 64DD dumps ? Not sure and honestly I would prefer to to have to worry about that...

LuigiBlood:
Nope, byteswapped doesn't really exist.

LuigiBlood:
Works for me so far.

LuigiBlood:
Actually a good idea to Generate LBA Table in load_dd_disk because I think swapping disks would have gone quite a bit wrong

bsmiles32:
Merging this as it provides some improvements over what currently exists.

To summarize, this PR:
- support MAME, SDK and D64 format without needing to convert between them. Format complexity is handled in disk.c
- support developement disk
- support USA and DEV 64DD IPLs.
- mimic access errors in known "bad" sectors
- can load Full Disk Copy or RAM only saves

What will be addressed in later PR:
- saving Full Disk Copy or RAM only (needs some prior work on the storage_backend)
- implement disk seek times
- disk swapping
- pass information about what disk is loaded to plugin
- support disk only case (no ROM)
- proper PI timings

PR #764, opened 2020-07-01
Get/Set Refresh Rate

Rosalie241:
This is still a Work-In-Progress, however I opened this PR early because I wanted some feedback/discussion per my planned changes.

I want to be able to set/get refresh rates using the mupen64plus API(for i.e GUIs, like m64p-gui), I also want to be able to 'properly' set a fullscreen mode with the refresh rate aswell,

I've also opened a GLideN64 PR https://github.com/gonetz/GLideN64/pull/2297 because ultimately my goal is to use upstream GLideN64(with GLideNUI) & mupen64plus-core for m64p-gui(or maybe in the future other GUIs), however, to add the ability to set/get refresh rates, the spec needs to be modified to support such a thing, I see 2 ways of doing it that way

1) modify the existing `VidExtFuncSetMode` & `m64p_2d_size` & `VidExt_ListFullscreenModes`
2) create 2 new functions (i.e `VidExtFuncSetMode2`, `int* VidExt_ListRefreshRates(m64p_2d_size input, int *count)`)

I'm not sure where to go from here so feedback is appreciated.


Narann:
I’m not sure to understand: What would be stored in the 32 slots of the array?

tim241:
The refresh rates, each resolution can contain multiple refresh rates.

On July 2, 2020 2:26:05 PM GMT+02:00, Dorian Fevrier <notifications@github.com> wrote:
>I’m not sure to understand: What would be stored in the 32 slots of the
>array?>
>>
>-- >
>You are receiving this because you authored the thread.>
>Reply to this email directly or view it on GitHub:>
>https://github.com/mupen64plus/mupen64plus-core/pull/764#issuecomment-652975182

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Rosalie241:
I don't want to seem impatient but I'd like to get this work done, can anyone please read my proposal and give suggestions/feedback so that I can continue my work?

Narann:
As it's an API modification, I think you can't modify function signature without breaking existing ecosystem.

> The refresh rates, each resolution can contain multiple refresh rates.
I think having multiple similar resolution but with different refresh rate wouldn't be better :

640x480, 60
640x480, 59.9
...

Instead of 640x480, [59.9, 60].

```
typedef struct {
  unsigned int uiWidth;
  unsigned int uiHeight;
  unsigned int refreshRate;
} m64p_2d_size;
```

With `refreshRate` as `unsigned int`, how do you represent 59.9 Hz?

Last question: Who would get this `refreshRate` value? IIRC, SDL can't change the refresh rate, so let's say GlideN64 get the 59.9 value, what should it does with it?

Rosalie241:
> As it's an API modification, I think you can't modify function signature without breaking existing ecosystem.

that's alright, I can extend the API to add i.e `VidExt_ListRefreshRates` to not break everything.

> > The refresh rates, each resolution can contain multiple refresh rates.
> > I think having multiple similar resolution but with different refresh rate wouldn't be better :
> 
> 640x480, 60
> 640x480, 59.9
> ...
> 
> Instead of 640x480, [59.9, 60].
> 
> ```
> typedef struct {
>   unsigned int uiWidth;
>   unsigned int uiHeight;
>   unsigned int refreshRate;
> } m64p_2d_size;
> ```
> 
> With `refreshRate` as `unsigned int`, how do you represent 59.9 Hz?

since I use SDL2 to fill that data, and SDL2 uses `int` for the refresh rate, you can't

https://wiki.libsdl.org/SDL_DisplayMode
https://wiki.libsdl.org/SDL_GetDisplayMode

> Last question: Who would get this `refreshRate` value? IIRC, SDL can't change the refresh rate, so let's say GlideN64 get the 59.9 value, what should it does with it?

it'd work something like this:
1) GLideN64 requests a list of resolutions and refresh rates from the m64p API
2) the mupen core library fills a list with the resolutions and refresh rates (in this case, m64p-gui has it's own video extension functions, which I've modified to use SDL2 to do this)
3) when going fullscreen, GLideN64 sets the video mode using the m64p API with the refresh rate (in this case, my prototype m64p-gui implementation uses SDL2 to set the fullscreen mode)

SDL1 can't change refresh rates, that's right, SDL2 can though

https://wiki.libsdl.org/SDL_SetWindowDisplayMode

Jj0YzL5nvJ:
$ xrandr
```
Screen 0: minimum 320 x 200, current 1920 x 1080, maximum 16384 x 16384
DisplayPort-0 connected primary 1920x1080+0+0 (normal left inverted right x axis y axis) 531mm x 299mm
   1920x1080     60.00 + 144.00*  119.98    99.93    84.90    50.00    59.94  
   1680x1050     59.95  
   1280x1024     75.02    60.02  
   1440x900      59.89  
   1280x960      60.00  
   1280x800      59.81  
   1152x864      75.00  
   1280x720      60.00    50.00    59.94  
   1024x768      75.03    70.07    60.00  
   832x624       74.55  
   800x600       72.19    75.00    60.32    56.25  
   720x576       50.00  
   720x480       60.00    59.94  
   640x480       75.00    66.67    60.00    59.94  
   720x400       70.08  
HDMI-A-0 disconnected (normal left inverted right x axis y axis)
DVI-D-0 disconnected (normal left inverted right x axis y axis)
```
What happens if I connect a 144Hz monitor to `DisplayPort-0`, an 85Hz CRT to` DVI-D-0` and a 30Hz HDTV to `HDMI-A-0`?
Will the settings on each screen be respected?
Exclusive mode, non-exclusive mode or both?
What would happen in case of using variable refresh rate (G-Sync, FreeSync, etc.)?
[HDR10 support...](https://github.com/gonetz/GLideN64/issues/1962)

Rosalie241:
> What happens if I connect a 144Hz monitor to `DisplayPort-0`, an 85Hz CRT to `DVI-D-0` and a 30Hz HDTV to `HDMI-A-0`?
> Will the settings on each screen be respected?

it should just work, that is if SDL2 handles such a case nicely (which I presume it does)

> Exclusive mode, non-exclusive mode or both?

exclusive mode only iirc (again, SDL2 will be doing everything)

> What would happen in case of using variable refresh rate (G-Sync, FreeSync, etc.)?

I have no idea, I don't think that's upto SDL2 to handle, I think that's a video plugins' job (something like opengl's Adaptive Vsync I suppose?)

> [HDR10 support...](https://github.com/gonetz/GLideN64/issues/1962)

SDL2 is gonna handle everything so I don't know.

Jj0YzL5nvJ:
> exclusive mode only iirc (again, SDL2 will be doing everything)

That is exactly the direct path to disaster (multi-monitor support). I prefer the non-exclusive mode. Exclusive mode adds nothing to current support.

Rosalie241:
@Narann I've made an alternative implementation of this PR(not fully finished yet, it's also untested), but would this be more sane?

https://gist.github.com/Rosalie241/4a8efb67d66f30863305bd1d6f1f16bf

richard42:
Looks pretty good to me. The only recommendation I have is to rename the function "VidExt_SetVideoMode2" to something more descriptive like "VidExt_SetVideoModeWithRate"

Rosalie241:
> Looks pretty good to me. The only recommendation I have is to rename the function "VidExt_SetVideoMode2" to something more descriptive like "VidExt_SetVideoModeWithRate"

Alright, updated the patch

Rosalie241:
I forgot about the documentation, so I've updated the patch to include it, hopefully it's clear enough.

richard42:
Okay this looks good. The only thing remaining is to advance the video extension API version number and update the API changelist. This is explained in this wiki page:
https://mupen64plus.org/wiki/index.php?title=Mupen64Plus_v2.0_API_Versioning

Rosalie241:
> Okay this looks good. The only thing remaining is to advance the video extension API version number and update the API changelist. This is explained in this wiki page:
> https://mupen64plus.org/wiki/index.php?title=Mupen64Plus_v2.0_API_Versioning
Alright, updated the patch

Rosalie241:
I slightly modified the implementation to verify that the input `Width` and `Height` of `VidExt_SetVideoModeWithRate` are actually valid, this also allows the code to re-use the same `SDL_DisplayMode` structure

m4xw (approved):
Most nits addressed.
I'd maybe recommend properly calculating .Functions count in the ptr table via sizeof but that would be beyond the scope of this PR.

PR #771, opened 2020-08-24
Makefile: add missing "

Rosalie241:
This patch fixes:
```
/bin/sh: -c: line 0: unexpected EOF while looking for matching `"'
/bin/sh: -c: line 1: syntax error: unexpected end of file
make: *** [Makefile:726: targets] Error 1
```
when running `make`


loganmc10:
Whoops good catch

PR #774, opened 2020-08-26
Allow plugins to have a configuration GUI

Rosalie241:
this patch modifies the spec to allow plugins to expose an optional `PluginConfig` function which can run a configuration GUI (like gliden64's glidenUI)

### DIFF
--- a/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Plugin-API.mediawiki
+++ b/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Plugin-API.mediawiki
@@ -25,6 +25,14 @@ These functions are present in all of the plugins.
 <br />
 {| border="1"
 |Prototype
+|'''<tt>m64p_error PluginConfig(void)</tt>'''
+|-
+|Usage
+|This optional function opens a configuration GUI for the plugin
+|}
+<br />
+{| border="1"
+|Prototype
 |'''<tt>m64p_error PluginShutdown(void)</tt>'''
 |-
 |Requirements
--- a/src/api/m64p_common.h
+++ b/src/api/m64p_common.h
@@ -72,6 +72,17 @@ typedef m64p_error (*ptr_PluginStartup)(m64p_dynlib_handle, void *, void (*)(voi
 EXPORT m64p_error CALL PluginStartup(m64p_dynlib_handle, void *, void (*)(void *, int, const char *));
 #endif
 
+
+/* PluginConfig()
+ *
+ * This optional function opens a configuration GUI for the plugin
+ *
+*/
+typedef m64p_error (*ptr_PluginConfig)(void);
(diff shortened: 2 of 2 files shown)


richard42:
I don't think you will be able to convince me to allow this change to the API.  The old Zilmar spec had a similar function, and I deliberately removed it in the 2.0 re-architecture because it is not a good design.  The plugins need to use the Config API to read and write their parameters. I want all of the GUI code to be centralized in the front-end application, and not scattered in a bunch of different modules.

Rosalie241:
I want i.e something like GlideNUI for other plugins aswell, how should I make an actually usable & reasonably looking Configuration GUI without creating a different GUI per plugin in the front-end? In my opinion it'd make much more sense if the plugins would be allowed to make a front-end (optionally) so that not every GUI would need to create custom configuration GUIs for every different plugin.

Rosalie241:
Another option would be to allow plugins to define how certain things are laid out in the GUI, however this'd require changes to the Config API, and I'm not sure if plugin devs would be very keen on such changes.

Narann:
> how should I make an actually usable & reasonably looking Configuration GUI without creating a different GUI per plugin in the front-end?

It's a freedom the API give you without restricting you on how parameters must be organized and which GUI toolkit to use. API simply expose parameters. It's up to the front end to get/set them and most importantly, organize them to provide the best UX way for your platform.

> In my opinion it'd make much more sense if the plugins would be allowed to make a front-end (optionally) so that not every GUI would need to create custom configuration GUIs for every different plugin.

I politely disagree. This make sense from a desktop front end maintainer perspective, but not from what the core is supposed to do. Current API is simple, and that's the reason mupen64plus can run almost anywhere. What would make more sense is that every plugin provide some documentation on how parameters interact to help front end devs so it takes less time to them to write those GUIs.

I don't think it's up to the API to fix the UI layout problem, specially because this UI layout problem highly depends on the platform. TBH, current API could be even more simple if it wouldn't care about loading/saving config file and let the front end taking care of the parameters.

My 2 cts.

richard42:
Before we re-designed the API, there was different GUI code, for multiple platforms, in all these different modules and it always seemed to be going out of date and it was very difficult to maintain.  One of the design goals of the new API is to put all of the GUI code in one module (the front-end application) so that it can be entirely owned and maintained by a single person or team. Regarding GUI layout for the configuration parameters, and front-end application could have a generic interface which presents the parameters in a table with controls for integers, floats, strings, etc and it will work fine.  For special plugins (such as the input plugin) which can benefit from custom code (ie, for testing joystick events, formatting the configuration strings, showing a graphical layout, etc) the front-end application author can decide to write custom dialogs for those plugins if he or she wishes.

Rosalie241:
Fair enough, I'll have to break the spec then in my own plugins to achieve my personal goal, thank you for the information.

PR #785, opened 2020-10-16
Tag outgoing UDP packets with EF DSCP

loganmc10:
You can read about DSCP/QoS here: https://en.wikipedia.org/wiki/Differentiated_services

This tags outgoing UDP packets with "EF" (Expedited Forwarding). This tells the local router that these packets are "realtime" packets and need to be given priority.

I verified that this is working as expected by looking at the traffic in Wireshark after implementing this change:
EF

The effects of this would depend on the users home router. Most "gaming" routers from ASUS or Netgear support QoS based on DSCP tags. Most SMB routers (for instance I use a Ubiquiti EdgeRouter X) also support this.

Since most home Internet plans are quite light on upload speeds, QoS can be helpful to prioritize real time traffic (voice, gaming).

### DIFF
--- a/src/main/netplay.c
+++ b/src/main/netplay.c
@@ -28,6 +28,10 @@
 #include "netplay.h"
 
 #include <SDL_net.h>
+#if !defined(WIN32)
+#include <sys/socket.h>
+#include <netinet/ip.h>
+#endif
 
 static int l_canFF;
 static int l_netplay_controller;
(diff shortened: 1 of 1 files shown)


loganmc10:
I guess Windows blocks directly changing the IP_TOS field, there is another API that needs to be used on Windows. I'll keep working on this

loganmc10:
It seems like getting this to work on Windows using SDL2_net would be almost impossible. I've updated it to just tag packets in Linux.

I did write a guide for Windows on how to do it manually though: https://github.com/loganmc10/m64p/wiki/Quality-of-Service

I think this is ready to be merged.

PR #788, opened 2020-10-22
Add M64CMD_ROM_SET_SETTINGS

Rosalie241:
This PR allows a front-end to 'modify' game settings until the ROM is closed, this allows for per game settings like what Project64 has

imagine something like this:
* Front-end has emulation running and provides a settings window, settings window provides per game options (i.e plugin selection, countperop, save type, etc)
* Front-end saves those settings to i.e a third party configuration file
* Next time front-end loads emulation, it'll execute `M64CMD_ROM_SET_SETTINGS` after the ROM has been opened & before emulation is launched


loganmc10:
Sorry I must not understand, what is the difference in purpose between CoreSetRomSettings and M64CMD_ROM_SET_SETTINGS.

Wouldn't the front-end just call M64CMD_ROM_SET_SETTINGS before launching the game?

Rosalie241:
Per discussion on Discord yesterday, I've removed the `CoreSetRomSettings` function

richard42 (changes requested):
This looks okay, except for 2 things.  One is that in src/main/rom.c, in open_rom() you set the same 3 parameters twice. I think the second group of 3 lines can be removed.  Also, the FRONTEND_API_VERSION should not go from 2.1.3 up to 3.0.0. There is some explanation in the Mupen64Plus-v2.0-API-Versioning document itself which describes how these version numbers should be changed as the software evolves. I would recommend bumping it up to 2.1.4

Rosalie241:
@richard42 this PR breaks existing front-ends, because `M64CMD_EXECUTE` has been moved +1, so now when existing front-ends call `M64CMD_EXECUTE`, they'll actually call `M64CMD_ROM_SET_SETTINGS`. That's why I changed the version number to `3.0.0`

Rosalie241:
As a workaround to the issue explained above, I've moved `M64CMD_ROM_SET_SETTINGS` to the bottom of `m64p_command` and changed the version to `2.1.4`

PR #789, opened 2020-10-23
Add fixed count factor overclock

carnivoroussociety:
This hack applies a constant for count register calculation (all core), much like the [recently added overclock to the libretro fork](https://github.com/libretro/mupen64plus-libretro-nx/pull/236).

This improves performance in many titles such as GoldenEye, Perfect Dark, Conker's Bad Fur Day, Indiana Jones, Re-Volt.

The overclock is mostly untested apart from the mentioned titles.

### DIFF
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -308,6 +308,7 @@ int main_set_core_defaults(void)
     ConfigSetDefaultString(g_CoreConfig, "SaveSRAMPath", "", "Path to directory where SRAM/EEPROM data (in-game saves) are stored. If this is blank, the default value of ${UserDataPath}/save will be used");
     ConfigSetDefaultString(g_CoreConfig, "SharedDataPath", "", "Path to a directory to search when looking for shared data files");
     ConfigSetDefaultInt(g_CoreConfig, "CountPerOp", 0, "Force number of cycles per emulated instruction");
+    ConfigSetDefaultInt(g_CoreConfig, "EnableOverclock", 0, "Override cycle calculation with fixed count factor (overclock)");
     ConfigSetDefaultBool(g_CoreConfig, "RandomizeInterrupt", 1, "Randomize PI/SI Interrupt Timing");
     ConfigSetDefaultInt(g_CoreConfig, "SiDmaDuration", -1, "Duration of SI DMA (-1: use per game settings)");
     ConfigSetDefaultString(g_CoreConfig, "GbCameraVideoCaptureBackend1", DEFAULT_VIDEO_CAPTURE_BACKEND, "Gameboy Camera Video Capture backend");
@@ -1304,6 +1305,7 @@ m64p_error main_run(void)
     size_t i, k;
     size_t rdram_size;
     uint32_t count_per_op;
+    uint32_t enable_overclock;
     uint32_t emumode;
     uint32_t disable_extra_mem;
     int32_t si_dma_duration;
@@ -1338,6 +1340,7 @@ m64p_error main_run(void)
     //We disable any randomness for netplay
     randomize_interrupt = !netplay_is_init() ? ConfigGetParamBool(g_CoreConfig, "RandomizeInterrupt") : 0;
     count_per_op = ConfigGetParamInt(g_CoreConfig, "CountPerOp");
+    enable_overclock = ConfigGetParamInt(g_CoreConfig, "EnableOverclock");
 
     if (ROM_PARAMS.disableextramem)
         disable_extra_mem = ROM_PARAMS.disableextramem;
@@ -1355,7 +1358,7 @@ m64p_error main_run(void)
         si_dma_duration = ROM_PARAMS.sidmaduration;
 
     //During netplay, player 1 is the source of truth for these settings
-    netplay_sync_settings(&count_per_op, &disable_extra_mem, &si_dma_duration, &emumode, &no_compiled_jump);
+    netplay_sync_settings(&count_per_op, &enable_overclock, &disable_extra_mem, &si_dma_duration, &emumode, &no_compiled_jump);
 
     cheat_add_hacks(&g_cheat_ctx, ROM_PARAMS.cheats);
 
@@ -1573,6 +1576,7 @@ m64p_error main_run(void)
                 g_mem_base,
                 emumode,
                 count_per_op,
+                enable_overclock,
                 no_compiled_jump,
                 randomize_interrupt,
                 g_start_address,
--- a/src/device/device.c
+++ b/src/device/device.c
@@ -83,6 +83,7 @@ void init_device(struct device* dev,
     /* r4300 */
     unsigned int emumode,
     unsigned int count_per_op,
+    unsigned int enable_overclock,
     int no_compiled_jump,
     int randomize_interrupt,
     uint32_t start_address,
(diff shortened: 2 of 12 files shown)


m4xw:
cc @Gillou68310 as discussed
@carnivoroussociety On a side note, you think its worth to expose the actual factor as a option?
Or has there been diminishing benefit with other values?
Also the netplay change needs a protocol version bump (cc @loganmc10 , thoughts on handling that?)

carnivoroussociety:
The factor itself was picked because it was the most compatible, having a lower value showed major improvements in GoldenEye and Perfect Dark, but many other titles crashed before reaching the main menu. Speaking of, I've noticed that Indiana Jones and Conker both don't reach the menu despite using the same constant rate value as the libretro fork, and when using the new dynamic recompiler core, only Indiana Jones is function. Both of these games work in the fork (it uses the new dynamic recompiler core).

m4xw:
Does it actually crash?

carnivoroussociety:
Conker softlocks at the [Nintendo logo](https://user-images.githubusercontent.com/38839485/96997525-45b85f80-1575-11eb-97c4-55268298aac9.png).

loganmc10:
I'd be curious to hear @Gillou68310 or @bsmiles32 thoughts on this.

This is a really odd way to do this to be honest. This sort of emulates an unstable/variable clock rate. Adding 6 to CP0_COUNT every time cp0_update_count is executed is sort of arbitrary (it won't emulate a steady clock rate). I would suspect a lot of games would have stability issues, or issues with timers/clocks.

I think it would be better to just divide ```((*r4300_pc(r4300) - cp0->last_addr) >> 2)``` by some factor, that way you'd get a (mostly) consistent emulated clock rate.

As for the netplay question, yes you just need to increment this: https://github.com/mupen64plus/mupen64plus-core/blob/master/src/main/netplay.h#L29 if there are changes made to the netplay code.

loganmc10:
Actually sorry on the netplay question, this is what needs to be incremented:

https://github.com/mupen64plus/mupen64plus-core/blob/6019724f90bb68d7ec7d3c4eb73329b54dcfcdaa/src/main/version.h#L34

Since the packet format is changing, also https://github.com/mupen64plus/mupen64plus-core/blob/master/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Netplay-API.mediawiki would need to be updated

carnivoroussociety:
> This is a really odd way to do this to be honest. This sort of emulates an unstable/variable clock rate. Adding 6 to CP0_COUNT every time cp0_update_count is executed is sort of arbitrary (it won't emulate a steady clock rate). I would suspect a lot of games would have stability issues, or issues with timers/clocks.
> 
> I think it would be better to just divide `((*r4300_pc(r4300) - cp0->last_addr) >> 2)` by some factor, that way you'd get a (mostly) consistent emulated clock rate.

I agree, it's a terrible hack that ended up working good enough for the more popular games. When I have time tomorrow I'll try out using a shift and addition instead of a constant rate. Thank you for the feedback!

carnivoroussociety:
I've changed the calculation to Logan's suggestion, testing confirms a lot more stable however Conker still softlocks (Indiana Jones is fine). If I increase shift to 4 it works but the speed difference compared to using count_per_op 1 is basically nothing. Interrupter and recompiler both use the same speed ratio.

carnivoroussociety:
> @carnivoroussociety On a side note, you think its worth to expose the actual factor as a option?

EnableOverclock is now used as a overclock factor variable - from a brief test it appears to work quite well.

Setting `EnableOverclock = 1` provides a 50% speed improvement over just using `CountPerOp = 1`, with each increase of EnableOverclock it will reduce count by half every time. I've set a hard limit to 50 as any higher gets diminishing returns with increased instability.

bsmiles32:
Not a huge fan of the current implementation (but this is just me and doesn't necessarily reflect other opinions).

If I understand well, your new computation of count update could be reduced to :
```
count = (delta PC) >> (2+oc_factor);
```
Since I don't really like having to maintain 2 separate way of doing the update count computation, if we were to merge the new computation we could instead do as follow:

```
count = (delta PC) >> 2;
count *= count_per_op;
count >>= oc_factor;
```

This would generalize the current count computation to allow "count_per_op = N / (2^K)", with N = current count_per_op value and K = oc_factor, without much added overhead/maintenance burden.
In such a case I would rather rename the oc_factor to something more meaningful (subjective) such as count_per_op_denom_pot (pot = Power Of Two).

Using a 2^K denominator provides some more flexibility regarding count_per_op, without much cost (single right shift), whereas an integer denominator  would require a division and would not be practical. So that's a good point. Also, setting K=0 will simplify to current behavior so that's another good point.

So maybe with the mentioned changes it would get better chances of getting merged.
What do you think ?

carnivoroussociety:
Thank you for the feedback. The hack works a little different than `count = (delta PC) >> (2+oc_factor);`. From testing I've found that shifting right using the overclock factor is unstable at the same overclock value, compared to shifting once on a loop. GoldenEye doesn't even make it to the logo using an overclock factor of 3, and crashes in-game with a value of 2. By using the loop it'll ensure count won't shift until zero. I could add a 1+ so it'll always be above zero, however that isn't much better than using the fixed count rate method I had originally. Perhaps I misinterpreted what you said and your described method is exposing a flaw in the hack itself.

I branched the stock/overclock update calculation as I wasn't sure if count_factor should be included, but I'll change it so there isn't a branch, thanks again for the feedback.

bsmiles32:
OK, I see my mistake here. You need division by 2^K but rounded up, not truncated.
You could do as follow then to avoid rounding to zero:
```
count = (delta PC) >> 2;
count *= count_per_op;
count += (1 << oc_factor) - 1
count >>= oc_factor;
```

carnivoroussociety:
That works perfectly thank you. I'll redo the adjustments along with renaming the overclock variable to your suggestion as it does make more sense.

carnivoroussociety:
Any progress for this PR? I've incorporated the feedback given by logan and bsmiles, and OC works the same across all cores.

Gillou68310:
The implementation for the new dynarec is not complete rn. I can help fixing it but I got other things I'd like to merge to the new dynarec before we can consider merging this. Sorry for the delay...

loganmc10:
My vote would be to merge this, it's off by default so there is no harm done to the new dynarec. If someone wants to submit a further PR to fix up the new dynarec they could.

Gillou68310:
I didn't look at the current implementation in detail, but what is the goal here? Add more granularity to the counter per ops? Like if counter per ops was actually a fractional number?

carnivoroussociety:
>Like if counter per ops was actually a fractional number?

Yes, so that greater count delta spikes are reduced while not breaking during DMA or critical timing segments in games - unlike the previous hack I implemented (constant count).

src/main/main.c
```diff
@@ -1350,12 +1353,15 @@ m64p_error main_run(void)
     if (count_per_op <= 0)
         count_per_op = ROM_PARAMS.countperop;
 
+    if (count_per_op_denom_pot > 50)
```
richard42, line 1356: I think this is way too big of a number for the power-of-two denominator.  Maybe 8 makes sense? When would you ever need to overclock by more than 256x?

carnivoroussociety: Yeah, you're right - anything over 10 is unnoticeable and more likely to cause issues. I'll set to 8 then.

PR #791, opened 2020-10-26
Allow front-ends to override user paths 

Rosalie241:
This PR introduces `ConfigOverrideUserPaths`, allowing a front-end to override the user paths returned by `ConfigGetUserDataPath` and `ConfigGetUserCachePath`


loganmc10:
Unfortunately I don't think this is the right approach. Shared data is things like mupen64plus.ini, on a traditional Linux install this would be located at /use/share/mupen64plus (not user writable).

User data would be in the users home directory.

For instance with m64p I override the shared data to be in the same directory as the exe, but that's not where I want the user data

Rosalie241:
@loganmc10 do you propose I should modify the plugins to use the `ConfigGetSharedDataFilepath` instead?

Rosalie241:
Another thing I'd like to note is that you can override the path for  `ConfigGetUserConfigPath`, so it seems kinda inconsistent if you can't override the path for `ConfigGetUserDataPath` and `ConfigGetUserCachePath` aswell in my opinion

loganmc10:
I don't think you can count on ConfigGetSharedDataFilepath being writeable, so I wouldn't try to have plugins write there.

Adding an override path for ConfigGetUserDataPath makes the most sense to me.

Rosalie241:
That's a good idea, but considering I have #792 open aswell, I think it'd be better to create a new function, i.e `ConfigOverrideUserPaths(char *DataPath, char *CachePath)` to avoid breaking the front-end API.

richard42:
I kind of don't like having a separate function to supply override values for these two paths, but I agree with you that this is preferable to breaking the front-end API.

Rosalie241:
@richard42 what do you think of this patch in it's current state?

richard42:
Everything looks good except for one minor issue.  If the front-end calls this Override function twice in row with non-null path strings, then during the second call it will leak the memory that it allocated for the path string in the first call. Just a small fix for this in 2 places in the override function and I think it should be good to merge.

Rosalie241:
Good catch, fixed!

PR #796, opened 2020-11-02
Align memory for parallel-rdp

loganmc10:
Port from https://github.com/libretro/mupen64plus-libretro-nx/commit/29ffb30c7f133d3fdff7b80dd0b0f63a9fa9e96e#diff-bf40cfecf6b954a6a6074a40e19efbb0bdc1a4997ec7f014dce41647d8a9038a

@m4xw can you review this and let me know if it looks right? I believe this is sort of step 1 to allowing Parallel RDP to work upstream. If there is anything else that needs to happen in the core (besides a Vulkan interface) please let me know

### DIFF
--- a/src/device/memory/memory.c
+++ b/src/device/memory/memory.c
@@ -42,6 +42,10 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+#ifdef _WIN32
+#include <malloc.h>
+#endif
+
 #ifdef DBG
 enum
 {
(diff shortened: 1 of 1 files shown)


m4xw:
image
This was my original diff (for reference), so identical, other than not using the HAVE_PARALLEL_RDP define (which we discussed together with Gillou and chose to always align the memory since it has virtually no downside).
Lgtm.

PR #798, opened 2020-11-11
Allow customizable save slot keybindings

Rosalie241:


richard42:
This is great; I really like your attention to detail, and the fact that you fix up the code comments along with your changes.  The only other thing that's missing with this change is to add the new config parameters to the Mupen64Plus_Core_Parameters wiki page in the core repository.  If you can send a pull request for this change as well I will merge it too. Thanks!

PR #799, opened 2020-11-15
Add support for netplay input delay.

joeldenning:
This is my first substantive contribution to mupen64plus-core - I am working on a netplay implementation that uses client-side input delay and this change lets that happen. An alternative implementation is to implement input delay on the server side - I'm open to switching to that if you think that that approach is better.

Looking forward to any feedback on this - happy to discuss any of the particulars.


src/main/netplay.c
```diff
@@ -382,7 +384,7 @@ static void netplay_send_input(uint8_t control_id, uint32_t keys)
     UDPpacket *packet = SDLNet_AllocPacket(11);
     packet->data[0] = UDP_SEND_KEY_INFO;
     packet->data[1] = control_id; //player number
-    SDLNet_Write32(l_cin_compats[control_id].netplay_count, &packet->data[2]); // current event count
+    SDLNet_Write32(l_cin_compats[control_id].netplay_count + input_delay, &packet->data[2]); // current event count
```
joeldenning, line 387: When input_delay is a positive int, this change means that the netplay client will not send inputs for the first few frames. In my testing with with m64p's netplay server, this has seemed fine, but I'm not sure if it's something that should be accounted for here.

loganmc10:
I think this would probably be easier to implement on the GUI/server side. For instance in m64p-netplay-server
https://github.com/m64p/m64p-netplay-server/blob/master/udpServer.cpp#L106

```
count = qFromBigEndian<quint32>(&incomingData.data()[2]);
count += input_delay;
```

Will accomplish the exact same thing, you can just add an option to the GUI and pass it to the server. You could even bundle it with your other change, where ```use_client_count``` represents the input delay (if it's 0 it's disabled, if it's a positive number, that is the input delay)

joeldenning:
Fair enough - I'll close this in favor of implementing this server-side

Issue #525, opened 2018-01-29
I need help with the transfer pak in a frontend I'm making


Mastergatto:
Hello everyone, first of all I want to apologize if I have made this ticket here not dealing with a bug, but just asking for help with a frontend I'm currently making (for some years and that isn't public yet) in python. Yeah, in python because it's easier for me as I'm quite a novice with programming, and I must say I have learned something and it was a very fun journey...except when dealing with ctypes. :S

Recently I have decided to try to implement transfer pak support, by wrapping some related [struct/function](https://github.com/mupen64plus/mupen64plus-core/blob/master/src/api/m64p_types.h#L168) like this:
`class m64p_media_loader(Structure):
    cb_data = c_void_p
    cart_rom = CFUNCTYPE(c_char_p, cb_data, c_int)
    cart_ram = CFUNCTYPE(c_char_p, cb_data, c_int)
    _fields_ = [
        ("cb_data", cb_data),
        ("get_gb_cart_rom", POINTER(cart_rom)),
        ("get_gb_cart_ram", POINTER(cart_ram))
    ]`

and the command like this:
`status = self.CoreDoCommand(t.m64p_command.M64CMD_SET_MEDIA_LOADER.value, c_int(sizeof(t.m64p_media_loader)), byref(t.m64p_media_loader))`

At this point if I try to launch the command like this, the core seems to accept the parameters without any issue. Of course, the struct was just initialised and it didn't hold any data. The real challenge is when I have to point to it the filenames of the GB ROM/RAM  (that are, obviously, already present in the [Transferpak] section in the .cfg file), thus I have tried to mess with the struct, shown above, and other things, but either Python refuses to execute the code or a segfault happens. At this point I'm lost and I don't know what to do, I may have also misunderstood the purpose of those functions.

Can you help me? Thanks

bsmiles32:
Not really knowledgeable about the python aspect of your issue, but I can try explain how to use the media_loader interface.

This interface allows the core to ask the frontend filenames to use for some emulated storage (eg GB ROM and GB RAM in this case). This is how it should be used:
1. the frontend registers it's media_loader (cb_data, get_gb_cart_{rom,ram}) using the SET_MEDIA_LOADER command:
https://github.com/mupen64plus/mupen64plus-ui-console/blob/a105ae0c6fd8480a05214f78e1a897f6d970e283/src/main.c#L969
then it let the core run
[...]

PR #803, opened 2020-12-06
storage backend save method nows know about modified byte range

bsmiles32:
This features allows to notify storage backends about the range of modified bytes, and in turn allow them to not have to do a full data write every time save method is called, reducing the saving bandwidth requirement.
This was needed to handle 64DD saves which are several order of magnitude bigger than existing saves.

See https://github.com/bsmiles32/mupen64plus-core/tree/wip_64dd_save for possible usage of this feature.

I'd like to have this merge quickly (in the next few days), so please test and voice your concerns if any.

cc : @richard42, @loganmc10, @fzurita

### DIFF
--- a/src/backends/api/storage_backend.h
+++ b/src/backends/api/storage_backend.h
@@ -37,7 +37,7 @@ struct storage_backend_interface
 
     /* Notify the storage backend that data should be persisted
      */
-    void (*save)(void* storage);
+    void (*save)(void* storage, size_t start, size_t size);
 };
 
 #endif
--- a/src/backends/file_storage.c
+++ b/src/backends/file_storage.c
@@ -35,6 +35,7 @@ int open_file_storage(struct file_storage* fstorage, size_t size, const char* fi
     /* ! Take ownership of filename ! */
     fstorage->filename = filename;
     fstorage->size = size;
+    fstorage->first_access = 1;
 
     /* allocate memory for holding data */
     fstorage->data = malloc(fstorage->size);
@@ -58,6 +59,7 @@ int open_rom_file_storage(struct file_storage* fstorage, const char* filename)
     fstorage->data = NULL;
     fstorage->size = 0;
     fstorage->filename = NULL;
+    fstorage->first_access = 1;
 
     file_status_t err = load_file(filename, (void**)&fstorage->data, &fstorage->size);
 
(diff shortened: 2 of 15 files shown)


bsmiles32:
also cc: @LuigiBlood

fzurita:
I don't see an issue with this.

In Android, I already have to copy the while 64DD because of the Android SAF.

To be truthful, I think it would be better if the media loader API took a memory address rather than a file like the ROM loading API does. That would prevent issues with Android SAF.

bsmiles32:
> I don't see an issue with this.

Good. Thanks!

> To be truthful, I think it would be better if the media loader API took a memory address rather than a file like the ROM loading API does. That would prevent issues with Android SAF.

Could you be more specific with the Android SAF issues (sorry I'm not knowledgeable about Android dev) ?
- If it is because you want memory mapped files, that may be done in the core (for the benefit of all users) via an extra "storage backend". Or at least that was my initial idea. Of course, it's a bit more complicated because some files have a format that is not the same as what is seen by the hardware (eg. byteswapped ROM) which greatly reduce the gains of avoiding the copy to a dedicated buffer.
- If it is because of poor interface design choice (especially regarding string memory allocation/free), then yeah... I may not have made the best decision back then.
But we can still deprecate/change things if enough people want that.

It's kind of hard to draw the line of what should be done in the frontend, and what should be done in the core, and where do we move the complexity and dependencies. It's always a trade-off.

fzurita:
The thing with Android SAF is that it prevents applications from having direct access to user viewable storage using paths. Applications can only use paths to view internal private storage. This means that when a user provides a 64DD disk, I have to copy the file to internal storage first before I can provide a path to the core.

If the media loader API used a file handle or a memory address, I wouldn't have to copy the file to internal storage first.

The way Android SAF works is through URIs. I have to request permission to a user to view a folder or file. Once the permission is granted, I get a URI. I can use the URI to get a file handle or use a java input stream to read the whole file to memory.

bsmiles32:
Thanks for the explanation.
We could use a "HACK" to pass pointers between frontend/core by teaching the core to recognize path looking like "buffer://<address_of_buffer_in_decimal>/<size_of_buffer_in_decimal>". The core would then populate the storage members accordingly. The core should not take ownership of such buffer. We'd still have the problem of how to handle the "save" callback...

In a sense, it's an extension of current API were we currently omit the "protocol" (because only file:// was used). But we could extend the API to support more protocols (and default to file:// if omitted). That would bring some flexibility without breaking existing code.

Well that's just some thinking on top of my head, not necessarily something to rely on.

If you feel like discussing this further maybe you should open an issue in the issue tracker so we can gather more input from frontend writers.

Mastergatto:
>     * If it is because of poor interface design choice (especially regarding string memory allocation/free), then yeah... I may not have made the best decision back then.
>       But we can still deprecate/change things if enough people want that.

Frontend 'maker' chiming in (#525 ), I was hesitant speaking about it, for the reason that the API has already settled in regarding the transfer pak and 64DD games, but I have never found a way to make the media loader work properly with my frontend; because that always result in a segmentation fault, double free or memory corruption, no matter what solution I have tried, that I have found by hopelessly browsing StackOverflow for days, if not weeks. Maybe I'm wrong and this theory is total BS, but I think the issue is that when I try to return the string (as an object) to the core through the media loader struct, Python will always free this returned object, and then when the core attempt to free, on its own, the same object that's already freed....

I haven't found a way to prevent Python from to do so, or to work around it. Or at least, if there is any, it probably involves changing the C code too (or even worse, the API), and unfortunately I'm not knowledgeable enough about the lower level parts of C (memory management and such) to attempt so. 

Egoistically speaking, I'd like to get rid of the media loader, but unfortunately I'm not betting on it, it's just me that isn't good at programming...

bsmiles32:
@Mastergatto yep, I remember your issue with transerpak and strings. The issue is indeed the API where I ask the frontend to allocate a string (with malloc) but transfer the ownership (eg. the responsibility of calling free) to the core. This worked for the console-ui, because there is no cross-language magic happening, but for your Python front-end it would require your to call C malloc, copy your Python string content to the malloc'd buffer and pass it to the core. Not great. It may be indeed more sensible to do the copy inside the core and not transfering ownership between core and frontend like I did. Still we could change that.

fzurita:
I crated this issue:
https://github.com/mupen64plus/mupen64plus-core/issues/804

PR #808, opened 2020-12-10
Fix to MacOSX building

SpiceyWolf:
After some struggle and assistance in the 'issues', this was the necessary changes to make the mac stuff build

### DIFF
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -211,11 +211,12 @@ ifeq ($(OS), OSX)
     ASFLAGS = -f macho -d LEADING_UNDERSCORE
   endif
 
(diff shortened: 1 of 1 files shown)


Jj0YzL5nvJ:
BTW, you don't need those quotes in this case. They are "good practice" to avoid weird behaviours in some shells, keep special characters, keep spaces, keep quotation over quotation, etc.

Narann:
This mean we drop Mac OS X 10.6 Snow Leopard and minimal version is now Mac OS X 10.7 Lion.

Are we OK with that ?

SpiceyWolf:
If you try to build with the libc++ it says minimum required is 10.7, not much of a way around that.

PR #814, opened 2020-12-20
Add SmashRemix0.9.5b to INI

loganmc10:
A new version of Smash Remix is out. It has a new CRC and MD5


data/mupen64plus.ini
```diff
@@ -13895,11 +13895,16 @@ CRC=9A6CF2F5 D5F365EE
 GoodName=SmashRemix0.9.4
 CRC=1B5AAD82 368B88C1
 Status=3
-Players=4
 SaveType=SRAM
+Players=4
```
Narann, line 13899: Is there a reason to move this line ?

loganmc10:
The settings are actually copy/pasted from the readme of the ROM hack. I just copied it this time to make sure it was correct and they had it in that order

PR #815, opened 2020-12-26
Simplify and make the asm script more robust to 3rd party scripts

Jj0YzL5nvJ:


Narann:
Hi,

Code is cleaner, but the exact purpose is still difficult to get.

Can we take this PR as an opportunity to write few lines in the 8 does:

```bash
GAS_LOOP=$(LC_ALL=C grep -a @ASM_DEFINE "$2" | sed 's/@ASM_DEFINE offsetof_struct_//g' | sed 's/\n/ /g')
```

Like this: `# for each .h file, we get every word starting by "offsetof_struct_" and remove the...`.

Because for not being used to compilation process, I (read: everyone) *need* to understand what those hacky code is *exactly* doing, not only the broad view.

Thanks in advance.

Narann:
I think it's better now.

Damn, this script is definitely not easy.

Jj0YzL5nvJ:
> Damn, this script is definitely not easy.

#817

PR #816, opened 2020-12-26
[Disk] Make sure the wrong System Data is checked upon read of the System Area

LuigiBlood:
It only impacted SDK format but there were serious regressions where games just didn't work properly. It seems I didn't do proper testing. Now they do.

Basically when you run Mario Artist Paint Studio, and try to access 2D Paint mode (the first on the left) and it would return Error 23 (Unrecovered Read Error), which is definitely weird. Other games were also impacted in weird ways so that was pretty annoying.


LuigiBlood:
@bsmiles32

PR #817, opened 2020-12-30
Failing to explain the asm scripts in a better way...

Jj0YzL5nvJ:
> Damn, this script is definitely not easy.

That means I suck to explain the workflow...

`asm_defines.o` and `asm_defines.obj` [looks something like this...](https://ci.appveyor.com/project/Narann/mupen64plus-core/builds/31155286/job/qsfs0mn7umnt66ta#L190) and `asm_defines_*.h` [like this...](https://ci.appveyor.com/project/Narann/mupen64plus-core/builds/31155286/job/qsfs0mn7umnt66ta#L300)

### DIFF
--- a/tools/gen_asm_script.sh
+++ b/tools/gen_asm_script.sh
@@ -9,25 +9,30 @@ rm -f "$1/asm_defines_*"
 
 #	Adaptation of 'gen_asm_defines.awk' / 'gen_asm_script.cmd':
 #
-#	1. Display 'asm_defines.o' as a list
-#	2. Look for the '@ASM_DEFINE' pattern
-#	3. Sort for easy reading
-#	4. Discard "1" ('@ASM_DEFINE') pattern
-#	5. Take "2" (offset*) and "3" (0x*) patterns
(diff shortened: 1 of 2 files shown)


tools/gen_asm_script.sh
```diff
@@ -9,25 +9,30 @@ rm -f "$1/asm_defines_*"
...
+#		6. Save current value and repeat step 5 on the next value
+		GAS_OFFSET=${GAS_CURSTR}
 	else
 
-		#	8. Save current "offset"
-		GAS_OFFSET=${GAS_CURSTR}
+#		7. Print current values and repeat steps 5-7 on the next value
+#		   until EOL is reached
+		echo "%define ${GAS_OFFSET} (${GAS_CURSTR})" >> "$NASM_FILE"
+		echo "#define ${GAS_OFFSET} (${GAS_CURSTR})" >> "$GAS_FILE"
```
Narann, line 36: Why use a negate here? What was wrong with the previous condition.

Jj0YzL5nvJ: Just to give consistency to the comments explaining the workflow...

PR #829, opened 2021-01-23
Full 64DD Saving Support

LuigiBlood:
I wanted to attempt an experiment with File Storage saving, such as defining an area to save which would deal with RAM Area only saving, which might be a mess.

@bsmiles32 I don't think that's what you had in mind, I just wanted to attempt something of my own.


LuigiBlood:
Also maybe it's just the unoptimized debug but saving does really take a toll on the emulator. Thankfully the disk isn't constantly written, but it does quite the performance impact.

bsmiles32:
Yeah, I would prefer if we keep the disk format logic (Full vs RAM only) inside disk module because it is disk-specific.
The idea of interfaces/backends is that the emulation logic only cares about interface (e.g. emulation logic only notifies of chunks it has changed) and the core is free to choose appropriate backend to store said changes. For now, we only have a file_storage_backend so this distinction is not so visible (and some network code got inside, even if doing a separate backend would be more appropriate to avoid mixing file/network logic).
For 64DD disks, we put a "format" layer in between emulation and core, so that we can abstract both saving format and saving medium orthogonally.

LuigiBlood:
Not working code but would it be appropriate to do something like that?
```c
static void storage_disk_save(void* storage, size_t start, size_t size)
{
    struct dd_disk* disk = (struct dd_disk*)storage;
    struct file_storage* tstorage = malloc(sizeof(struct file_storage));

    switch (disk->save_format)
    {
    case 0: /* Full disk save */
        disk->istorage->save(disk->storage, start, size);
        break;

    case 1: /* RAM only */
        tstorage->data = disk->storage->data + disk->offset_ram;
        tstorage->size = disk->storage->size - disk->offset_ram;
        tstorage->filename = disk->storage->filename;
        disk->istorage->save(tstorage, start, size);
        break;
    default: /* ignoring */
        break;
    }

    free(tstorage);
}```

bsmiles32:
I made an alternative proposition for saving https://github.com/LuigiBlood/mupen64plus-core/pull/7
The idea was to keep file_storage as is and move format specificity to disk module (and some part of it inside load_dd_disk). I had to add an extra save_storage member to disk struct linking to (part of) full disk storage so that I can have both full disk for reading and a save part. Might be a bit confusing, but it should work in practice.
Could you test to see if its good ?

LuigiBlood:
Well it seems to work fine, it's a bit similar to what I did except it's kept in memory as part of dd_disk. If you're happy with that I'll merge it, I'll just do a couple changes like force full disk saves for MAME format, or maybe not even allow to save at all.

bsmiles32:
> I'll just do a couple changes like force full disk saves for MAME format
Yeah, you mentioned it but I forgot about it, so it's cool if you can do it. I think switching to a Full disk copy for MAME format is better than no saving at all, but either way is fine with me.

From my side it's all good, you can merge my changes to your PR and do the MAME saving handling then we can merge your work.

(If you want we can cherry-pick your bugfix commit right now independently of this PR if you want it in master faster)

LuigiBlood:
There's a huge bug with NDD format right now I get disk read errors when trying to go into 2D Paint mode in Paint Studio.

bsmiles32:
Oups, I'll check later what's wrong with my code.

LuigiBlood:
Gotta note that for some reason, it's doing it fine with D64 format, which is kinda odd.

bsmiles32:
I can't reproduce your error. I can run DMPJ just fine and save both as full disk copy or RAM only. There was a bug for RAM only save the first time, when the file didn't exist, it would fail to load and fallback to no save therefore preventing to save and create the RAM file.
But now it should be OK.

LuigiBlood:
Well I'll merge for now but I need to test more on my end.

LuigiBlood:
...well I'm not sure what happened but it works now, maybe I just suck at using Git

LuigiBlood:
Also the bugfix for D64 system data modification I might do something else than what I originally did, and for MAME format to force Full Disk saves it's gonna be more complex than I thought so I'd rather just disable saves, no one should really use that format anyway.

bsmiles32:
He he !
I proposed a similar fix to the MAME case, but upper in the logic to allow it to load the full save if present. Since 64DD disk format guessing is done based on file size, I added an utility function to get file size early (before save loading and also before scan_and_expand function).
The PR is https://github.com/LuigiBlood/mupen64plus-core/pull/8

LuigiBlood:
Well that's even better, I didn't want to add functionality as I didn't really want to break anything since my way of things are quite not like the project lol

LuigiBlood:
The only problem is that it creates *.ram files when it's supposed to be *.ndr. Otherwise this seems to be fine.

LuigiBlood:
I think this is ready to go? I don't think I see anything to add about saving.

bsmiles32:
Closing this PR and will merge #832 once CI is green.
The new PR is rebased against master and I squashed all the commits (except the one fixing some offset) to have a more readable commit history. I've also removed some duplicated logic for the non appropriate *.ram extension for MAME disks.

PR #834, opened 2021-01-31
refactor lba_end computation to better convey intent

bsmiles32:
cc @LuigiBlood 
Did what you mentionned in the previous PR. Should be equivalent to previous code.


LuigiBlood:
this is good

PR #835, opened 2021-01-31
Change mempak formatting

bsmiles32:
This should better match what the underlying file system expects and is
in accordance with libdragon formatting procedure.
Previously we were writing 0x0003 (=Free page indicator) until the end
of mempak, but this should only be done in pages 5-128 slots.
The remaining content of the mempak is zeroed.

NOTE: this only change the default data for newly created MPK files.

cc: @bryc


Narann:
Nice!

PR #840, opened 2021-02-21
Add Dinosaur Planet to ini

loganmc10:
Sent to me in: https://github.com/loganmc10/m64p/pull/129

I tested the ROM and it works with these settings


fzurita:
Count per op of 1 makes enemies really hard to hit, almost impossible. Count per op of 2 makes them easier to hit, but the frame rate is lower.

loganmc10:
Interesting, I did notice it was almost impossible to hit the first bad guy, I didn't realize that was the cause.

I submitted https://github.com/mupen64plus/mupen64plus-core/pull/841 to remove that setting

PR #843, opened 2021-02-23
[new_dynarec] Trivial perf improvement on initial recompilation

m4xw:


m4xw:
Combats the mess at https://github.com/mupen64plus/mupen64plus-core/blob/master/src/device/r4300/new_dynarec/new_dynarec.c#L7726-L7782

PR #844, opened 2021-02-28
New Database fixes

Mastergatto:
Since there are a lot of fixes, I thought it's best to make two separate commits for (a bit) easier review.

The first one commit is very large and the reason for this is that currently, according to the database, official and licensed games advertise their support for Pak accessories when they actually don't. For example, Super Mario 64 (U) [!] doesn't support any of available accessories for N64, but here on mupen64plus the game appears to advertise Rumble Pak support, which is false. So, to correct this lie, we must specify also what Paks it doesn't support, and the end result is that for all games we add some thousand of new lines to database. 

The second commit just adds new prototypes and betas, and it also specify what Paks they support. There are also minor fixes, correcting good names to reflect more those in No-Intro. I have also moved Japanese OoT entries at the bottom of the list, for which it can make the PR appearing confusing to whose have to review it, and I apologize for that.

I have tested this new version of the database and it hasn't given any errors so far.

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -35,8 +35,19 @@ Mempak=Yes
 Rumble=Yes
 Transferpak=No
 
+[39ABE0C383A44C464B408BE3E1D6766E]
+GoodName=007 - The World is Not Enough (U) (v2) (Beta) [!]
+CRC=7BD65F9E 76B6A083
+Players=4
+SaveType=None
+Status=1
+Biopak=No
+Mempak=Yes
+Rumble=Yes
+Transferpak=No
+
 [12BD3FAFB4E064B6B0C07B7EE156243B]
-GoodName=007 - The World is Not Enough (U) (Beta) [!]
+GoodName=007 - The World is Not Enough (U) (v21) (Beta) [!]
 CRC=A92E0966 341C3912
 Players=4
 SaveType=None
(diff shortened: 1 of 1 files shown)


Jj0YzL5nvJ (commented):
I dislike this "new" logic.

If the Pak is omitted it means it's not supported. It's easier to just add supported devices and initialize anything else as "No" (Yes | No, True | False, 1 | 0).

Narann:
> according to the database, official and licensed games advertise their support for Pak accessories when they actually don't

I suspect this is the weird statement here. Where the database states a game support Pack accessories when it actually don't?

Jj0YzL5nvJ:
> Where the database states a game support **Pak** accessories when it actually don't?

As far as I know, the opposite case is more common. Many games can unlock additional functionality with certain Paks, even old ones by omission.

Example: If you have Perfect Dark and activate Hi-res with Expansion Pak and then you power off the console and switch to OoT, the PD's anti-aliasing is applied to OoT as well. But OoT hangs if you render certain parts of Hyrule Field. The Hi-res mode can be deactivated from any other game that supports Hi-res. But to apply it to OoT, it only recognizes PD's one.

Mastergatto:
> > according to the database, official and licensed games advertise their support for Pak accessories when they actually don't
> 
> I suspect this is the weird statement here. Where the database states a game support Pack accessories when it actually don't?

It's true that the database doesn't explicitly state that, but...bear with me.

For example, this is with current database:
actual

With proposed fixes:
fixes

I suspect that somehow mupen64plus initialize by default the Rumble Pak and the Controller Pak, unless the relevant keys declare expressly the opposite.

Mastergatto:
If there's no interest in merging this PR then we can close it, although I'd really like to see it being merged instead.

PR #847, opened 2021-03-03
Netplay fixes from Android port

loganmc10:
I wanted to make sure these don't get missed:

https://github.com/mupen64plus/mupen64plus-core/commit/07a22eeff3188b1141c87c26a7a7bba807185241
https://github.com/mupen64plus/mupen64plus-core/commit/a00ba8dca007fdf939b8d83e5090869144f6a12a

@fzurita perhaps you could comment on the results of the speed limiter change? Do you think it could be merged into the upstream code?

### DIFF
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -859,8 +859,6 @@ static void apply_speed_limiter(void)
     static unsigned int StartFPSTime = 0;
     static const double defaultSpeedFactor = 100.0;
     unsigned int CurrentFPSTime = SDL_GetTicks();
-    static double sleepTimes[SAMPLE_COUNT];
-    static unsigned int sleepTimesIndex = 0;
 
     // calculate frame duration based upon ROM setting (50/60hz) and mupen64plus speed adjustment
     const double VILimitMilliseconds = 1000.0 / g_dev.vi.expected_refresh_rate;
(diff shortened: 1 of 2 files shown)


fzurita:
Yeah, I think the speed limiter change can be merged. It mainly gets rid of the averaging and limits how long it can try to run faster to catch up if we fall behind momentarily.

loganmc10:
I know that @joeldenning experienced some issues with the speed limiter during netplay when he was doing performance/latency testing for Smash Bros.

I'm curious @joeldenning do you think you'd be able to test this change in a way that would tell you if it improved your netplay performance? I know you showed me some frame rate tracking that you were doing at the time.

loganmc10:
I suspect we may not hear back from Joel, @m4xw and @fzurita are you ok with these changes? I believe you've discussed it a bit on Discord?

m4xw:
fwiw it can be cleaned up more

loganmc10:
Do you have specific suggestions?

m4xw:
For example the boolean equation `if(sleepTime > 0 && sleepTime < maxSleepNeeded*SpeedFactorMultiple && l_MainSpeedLimit)` can be optimize to check for `l_MainSpeedLimit` first, saves some unrequired calculations on the `<` case, iirc fzurita had smth in mind as well, can't remember what he meant tho

loganmc10:
Ok I made the suggested optimization

loganmc10:
Ok I've addressed the issues mentioned, any other feedback? I'm eager to merge a more netplay friendly limiter :)

loganmc10:
Still looking for someone to merge this plz

Issue #842, opened 2021-02-23
Dinosaur planet beta TLB exceptions in new dynarec


fzurita:
The 64 bit ARM new dynarec is causing TLB exceptions in dinosaur planet.
[game crash.Dinosaur Planet (Dec 2000 Beta).zip](https://github.com/mupen64plus/mupen64plus-core/files/6028933/game.crash.Dinosaur.Planet.Dec.2000.Beta.zip)

In the attached save state, it crashes about 30 second in. With the game reporting a TLB exception and providing a backtrace.

The crash doesn't happen on cached interpreter.

Gillou68310:
What about x86/x64/ARM32 new dynarec?

Gillou68310:
Ok I'll test don't worry. Does it crash after 30sec with or without user input?

fzurita:
I just tested arm32 new dynarec, it also crashes, but the game soft locks, it doesn't go into the debug screen.

The game crashes with no user input, it's in a cut scene.

Gillou68310:
Cool it's gonna be easier to track down the issue then.

PR #848, opened 2021-03-04
Fix https://github.com/mupen64plus/mupen64plus-core/issues/842

Gillou68310:


Narann:
I think it deserves a better name, don't you? :D

richard42:
is it ready to go or still WIP?

Gillou68310:
Should be ready to go!

PR #860, opened 2021-06-01
Support Unicode on Windows

loganmc10:
In order for Windows to support Unicode characters, it needs to use the _w versions of the functions (_wfopen and gzopen_w for example).

Right now mupen64plus doesn't support referencing files on Windows that are in a path containing non-Latin characters, this PR fixes that problem.

### DIFF
--- a/src/main/savestates.c
+++ b/src/main/savestates.c
@@ -43,6 +43,7 @@
 #include "device/device.h"
 #include "main/list.h"
 #include "main/main.h"
+#include "osal/files.h"
 #include "osal/preproc.h"
 #include "osd/osd.h"
 #include "plugin/plugin.h"
@@ -200,7 +201,7 @@ static int savestates_load_m64p(struct device* dev, char *filepath)
 
     SDL_LockMutex(savestates_lock);
 
-    f = gzopen(filepath, "rb");
+    f = osal_gzopen(filepath, "rb");
     if(f==NULL)
     {
         main_message(M64MSG_STATUS, OSD_BOTTOM_LEFT, "Could not open state file: %s", filepath);
@@ -1338,7 +1339,7 @@ static int savestates_load_pj64_unc(struct device* dev, char *filepath)
     FILE *f;
 
     /* Open the file. */
-    f = fopen(filepath, "rb");
+    f = osal_file_open(filepath, "rb");
     if (f == NULL)
     {
         main_message(M64MSG_STATUS, OSD_BOTTOM_LEFT, "Could not open state file: %s", filepath);
@@ -1359,7 +1360,7 @@ static int savestates_load_pj64_unc(struct device* dev, char *filepath)
 static savestates_type savestates_detect_type(char *filepath)
 {
     unsigned char magic[4];
-    FILE *f = fopen(filepath, "rb");
+    FILE *f = osal_file_open(filepath, "rb");
     if (f == NULL)
     {
         DebugMessage(M64MSG_STATUS, "Could not open state file %s\n", filepath);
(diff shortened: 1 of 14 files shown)


loganmc10:
I believe this is ready to merge. I had a German-language user who has a Unicode character in their username test it: https://github.com/loganmc10/m64p/issues/157#issuecomment-852171057

@bsmiles32 @richard42 @Gillou68310 @Narann

src/main/savestates.c
```diff
@@ -200,7 +201,13 @@ static int savestates_load_m64p(struct device* dev, char *filepath)
 
     SDL_LockMutex(savestates_lock);
 
+#ifdef _WIN32
+    wchar_t wstr_filepath[PATH_MAX];
+    MultiByteToWideChar(CP_UTF8, 0, filepath, -1, wstr_filepath, PATH_MAX);
+    f = gzopen_w(wstr_filepath, "rb");
+#else
     f = gzopen(filepath, "rb");
+#endif
```
Narann, line 210: Does it make sense to have a `osal_gzopen()` in `osal/files.c`?

@richard42

src/main/savestates.c
```diff
@@ -1477,7 +1484,13 @@ static void savestates_save_m64p_work(struct work_struct *work)
     SDL_LockMutex(savestates_lock);
 
     // Write the state to a GZIP file
+#ifdef _WIN32
+    wchar_t wstr_filepath[PATH_MAX];
+    MultiByteToWideChar(CP_UTF8, 0, save->filepath, -1, wstr_filepath, PATH_MAX);
+    f = gzopen_w(wstr_filepath, "wb");
+#else
     f = gzopen(save->filepath, "wb");
+#endif
```
Narann, line 1493: Same here: Does it make sense to have a `osal_gzopen()` in `osal/files.c`?

@richard42

Narann:
Thanks for the good work!

Looks nice to me, except the few questions to @richard42 (I'm not a Windows dev).

loganmc10:
Yeah I didn't make an osal_gzopen since it only occurred in 1 file, it didn't seem like there was much to be gained by abstracting it.

Narann:
This would centralize `ifdef WIN32` which is cleaner, IMHO.

loganmc10:
Ok I've created the osal_gzopen function

Narann:
Thanks! Looks better to me, I will let @richard42 merge this.

src/osal/files_win32.c
```diff
@@ -152,43 +156,46 @@ const char * osal_get_shared_filepath(const char *filename, const char *firstsea
...
-        return chHomePath;
+        wcscat(chHomePath, L"\\");
+        WideCharToMultiByte(CP_UTF8, 0, chHomePath, -1, outString, MAX_PATH, NULL, NULL);
+        return outString;
     }
     else
     {
-        osal_mkdirp(chHomePath, 0);
-        if (_stat(chHomePath, &fileinfo) == 0)
+        _wmkdir(chHomePath);
```
richard42, line 188: the only problem I have with this PR is this.  the osal_mkdirp() function will create any necessary new folders in the full path, but _wmkdir() will only create one new folder (the last item in the given path).  One solution for this would be to add a new static function in files_win32.c called _wmkdirp() or something like that.

loganmc10: This was probably just an oversight on my part, I don't think there is any reason that this can't be osal_mkdirp, since I updated the internals of that function to use _wmkdir for windows. I'll test it tomorrow

richard42: okay, good. just be careful because osal_mkdirp() might expect utf8 input but I think chHomePath has already been converted to wchar here.

loganmc10: Ok I took care of this, now it's calling ```osal_mkdirp``` (after converting the string to utf-8)

src/osal/files_win32.c
```diff
@@ -91,8 +93,10 @@ int osal_mkdirp(const char *dirpath, int mode)
...
         *lastchar = 0;
 
+    wchar_t w_mypath[PATH_MAX];
+    MultiByteToWideChar(CP_UTF8, 0, mypath, -1, w_mypath, PATH_MAX);
     // Terminate quickly if the path already exists
-    if (_stat(mypath, &fileinfo) == 0 && (fileinfo.st_mode & _S_IFDIR))
+    if (_wstat(w_mypath, &fileinfo) == 0 && (fileinfo.st_mode & _S_IFDIR))
         goto goodexit;
 
     while ((currpath = strpbrk(currpath + 1, OSAL_DIR_SEPARATORS)) != NULL)
```
richard42, line 102: Looking over this function in particular, I think this loop will not work with your wide-char modifications.  It is going through the path components and modifying the original (utf8) string, but you are calling _wstat and _wmkdir on the wide-char converted string.  For this to work properly, the currpath and mypath variables should be operating on the wide-char string, not the original UTF8 string.

loganmc10: Thanks, yeah when I tested this I only did it with an existing directory, I didn't look at this close enough.

I fixed up osal_mkdirp to use wide chars from start to finish. I tested on my PC by settings my config directory to "existing_dir/one/two" and it did create directories "one" and "two" and put the config inside there.

PR #865, opened 2021-06-10
Fix Pokemon Puzzle League saving

loganmc10:
Saving in Pokemon Puzzle League is currently broken. No .fla file is created at all.

It was broken by: https://github.com/mupen64plus/mupen64plus-core/commit/6049618def04148ae8c6ab3146fd87882231df71

This commit restores this line (using the new logic):
https://github.com/mupen64plus/mupen64plus-core/commit/6049618def04148ae8c6ab3146fd87882231df71#diff-af6d71b286c633a07adc8703bbbb9f229f697dcd1c913aafad1a592d61492205L90

This makes Pokemon Puzzle League saving work again.

I checked a few other Flash RAM games (Command & Conquer, Pokemon Stadium, JFG, Paper Mario) and they are all still working normally. @bsmiles32 I'd really like to hear your feedback on this if possible

### DIFF
--- a/src/device/cart/flashram.c
+++ b/src/device/cart/flashram.c
@@ -106,6 +106,7 @@ static void flashram_command(struct flashram* flashram, uint32_t command)
     case 0xe1000000:
         /* set silicon_id mode */
         flashram->mode = FLASHRAM_MODE_READ_SILICON_ID;
+        flashram->status |= 0x01; /* Needed for Pokemon Puzzle League */
         break;
 
     case 0xf0000000:


bsmiles32:
I'm not sure why the program busy flag should be set when entering silicon_id mode. Maybe some kind of timing issue + incorrect usage of flashram ? But anyway, if it fixes the game I'm OK for merging this if you add a comment like "// Needed for Pokemon Puzzle league" so we know what game to check if we change that logic in the future.

loganmc10:
Sure, I've made that change

loganmc10:
I think this should be ok to merge now

Narann:
Thanks ! :)

Issue #738, opened 2020-03-25
libdragon games only work with Pure Interpreter


loganmc10:
Here are 2 example ROMs:
(DOOM Shareware port to N64)
[DOOMSHAREWARE.zip](https://github.com/mupen64plus/mupen64plus-core/files/4378824/DOOMSHAREWARE.zip)

(FlappyBird port to N64)
[FlappyBird.zip](https://github.com/mupen64plus/mupen64plus-core/files/4378829/FlappyBird.zip)

Both of these games were created with the "libdragon" homebrew SDK. They work with the Pure Interpreter, but not the Cached Interpreter.

You need to use CXD4 RSP + Angrylion to play them, or the very latest master of GLideN64

loganmc10:
I don't fully understand how the compare core feature works, but I ran it using FlappyBird using Pure Interpreter, set to "send" mode, and then again using the Cached interpreter, using "recv" mode, and this was the output:

```
Compare #7940983  old_op: 0 op: 0
err:    gpr  addr:80000184	 My: reg[8]=241004e5	 Ref: reg[8]=241004e1	 My: reg[27]=1	 Ref: reg[27]=ffffffff80010000
```

loganmc10:
Output for Donkey Kong 64 (U) [f2]:

```
Compare #8886165  old_op: 0 op: 0
err:    gpr  addr:80000184	 My: reg[8]=5	 Ref: reg[8]=1	 My: reg[26]=ffffffffdeadbeef	 Ref: reg[26]=ffffffff80100000
```

loganmc10:
And here is the backtrace from the new dynarec:

```
#1  0x00007ffff66af8d9 in abort () at /lib64/libc.so.6
#2  0x00007ffff66af7a9 in _nl_load_domain.cold () at /lib64/libc.so.6
#3  0x00007ffff66bea66 in annobin_assert.c_end () at /lib64/libc.so.6
#4  0x00007fffb8348fe6 in output_modrm (mod=3 '\003', rm=0 '\000', ext=255 '\377') at ../../src/device/r4300/new_dynarec/x64/assem_x64.c:712
#5  0x00007fffb8349278 in emit_mov (rs=-1, rt=0) at ../../src/device/r4300/new_dynarec/x64/assem_x64.c:749
[...]

PR #866, opened 2021-06-11
invalidate alternate address on write

loganmc10:
Fixes https://github.com/mupen64plus/mupen64plus-core/issues/738

@Gillou68310 @bsmiles32 @richard42 

I'd appreciate some feedback to ensure I've done the right thing here (I think I have).

libdragon seems to have a habit of accessing memory using the 0x8.... address and the 0xa.... address interchangeably.

In the other calls to ```invalidate_r4300_cached_code```, both addresses are invalidated, but not in ```r4300_write_aligned_(d)word```.

This allows libdragon software like Flappy Bird to function. I didn't see any negative side affects with other games.

### DIFF
--- a/src/device/r4300/r4300_core.c
+++ b/src/device/r4300/r4300_core.c
@@ -360,6 +360,7 @@ int r4300_write_aligned_word(struct r4300_core* r4300, uint32_t address, uint32_
     }
 
     invalidate_r4300_cached_code(r4300, address, 4);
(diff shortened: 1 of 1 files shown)


Gillou68310:
This only fixes cached interpreter right?

loganmc10:
> This only fixes cached interpreter right?

I just tested, it now works on the new dynarec (tested x64), but not the old dynarec. I'm guessing there is something else unrelated to this issue that would need to be fixed on the old dynarec.

This particular code isn't specific to a single interpreter/dynarec.

PR #868, opened 2021-06-11
Update PI DMA alignment as per CEN64

loganmc10:
Taken from: https://github.com/n64dev/cen64/blob/master/pi/controller.c

Tested using: https://github.com/PeterLemon/N64/tree/master/CPUTest/DMAAlignment-PI-cart

Before:
dma_alignment_(rom_--000

After:
dma_alignment_(rom_--000

There are still 2 tests that fail here that pass on CEN64: https://github.com/n64dev/cen64/pull/200. However, I couldn't get those tests to pass without breaking the graphics in AI Shougi 3, so I left it at the point where AI Shougi 3 would work. AI Shougi 3 graphics are messed up in CEN64, so I assume there is still more work to be done on their implementation.

### DIFF
--- a/src/device/rcp/pi/pi_controller.c
+++ b/src/device/rcp/pi/pi_controller.c
@@ -54,8 +54,8 @@ static void dma_pi_read(struct pi_controller* pi)
         return;
 
     uint32_t cart_addr = pi->regs[PI_CART_ADDR_REG] & ~UINT32_C(1);
-    uint32_t dram_addr = pi->regs[PI_DRAM_ADDR_REG] & 0xffffff;
-    uint32_t length = (pi->regs[PI_RD_LEN_REG] & UINT32_C(0x00fffffe)) + 2;
+    uint32_t dram_addr = pi->regs[PI_DRAM_ADDR_REG] & 0xfffffe;
+    uint32_t length = (pi->regs[PI_RD_LEN_REG] & UINT32_C(0x00ffffff)) + 1;
     const uint8_t* dram = (uint8_t*)pi->ri->rdram->dram;
 
     const struct pi_dma_handler* handler = NULL;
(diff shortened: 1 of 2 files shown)


Narann:
Thanks for the hard work. The code is clean, thanks for this!

I can merge this because I suspect no-one will have the required knowledge to validate this and the scope of this flag seems to be limited.

loganmc10:
Yeah it's ready now, I don't plan on making any further changes

loganmc10:
Actually hold on for a little bit. @rasky did some hardware tests to figure out how to get the test ROM and AI Shougi 3 working

Narann:
Haha! :) No worry!

loganmc10:
Ok I'm happy with this now. The test ROM passes at the same level as CEN64, AI Shougi works. I also tested other games I know that have specific PI alignment requirements (like command & conquer and taz express). Everything seems to be working normally.

This work was based on hardware tests done by @rasky:
image

As well as tests previously documented here: https://n64brew.dev/wiki/Peripheral_Interface#Unaligned_DMA_transfer

Narann:
Thanks for the hard work!

I don't know how the compatibility situation is but can you just convert the `//` by `/* */` ?

loganmc10:
ok done

Narann:
Thanks! :)

m4xw:
This caused a regression for 64DD for me.
Reverting pi_controller solves this. No game would boot (stuck on black screen) or it would just enter the IPL.
Didnt check yet which change in particular caused this, just a headsup.

loganmc10:
No problem, I'll fix this and submit a new pr

m4xw:
Thanks, if you read discord, my suspicion rn is the `pi->regs[PI_CART_ADDR_REG]` touching it in some way that doesnt trigger the
 ```
       if ((pi->regs[PI_CART_ADDR_REG] == MM_DD_C2S_BUFFER) ||
            (pi->regs[PI_CART_ADDR_REG] == MM_DD_DS_BUFFER)) {
```
condition for the if(dd) stuff at end of pi event.
Didnt spend time today investigating further tho.

m4xw:
Forgot to add, as far i can tell from the talks with luigi, we can consider the end of pi dma event stuff for DD a hack. So feel free to rework as you see fit.

PR #869, opened 2021-06-11
Update RSP alignment code as per CEN64

loganmc10:
Taken from: https://github.com/n64dev/cen64/blob/master/rsp/interface.c

Tested using: https://github.com/PeterLemon/N64/tree/master/RSPTest/DMAAlignment-SP

Before:
dma_alignment_(sp_- -001

After:
dma_alignment_(sp_- -000

With these changes, we achieve the same compatibility as CEN64


loganmc10:
By the way, this does line up with the documentation:
image

Narann:
Thanks!

PR #872, opened 2021-06-16
Fix address check for DD DMA Interrupt

loganmc10:
Fixes the issue identified in https://github.com/mupen64plus/mupen64plus-core/pull/868#issuecomment-861760293

@m4xw


m4xw (approved):
lgtm

PR #873, opened 2021-06-17
VRU support and Hey You Pikachu audio fix

loganmc10:
https://youtu.be/6_iPf8ZC_3M

* Windows build of m64p with VRU support: https://drive.google.com/file/d/1ZarcKEHs0ZpNhtan0QatjdIOPhxCP3S1/view?usp=sharing
* Linux: https://drive.google.com/file/d/191vE6ZoghRPGHdyUUnClEYaeA0onBz-1/view?usp=sharing

You can see the other changes that were made (to the input plugin) by looking at this branch: https://github.com/loganmc10/m64p/tree/vru

* VRU Support
  - Only 2 games support the VRU: Hey You Pikachu, and Densha de Go! (English translation: https://www.romhacking.net/translations/2947/ also supports the VRU)
  - Might not be perfect. I don't own a physical VRU, so all this work was done based on what zoinkity wrote and what I could get from the programming manual, and from trying to match the behaviour I saw on YouTube playthroughs).
  - This is an HLE implementation. I did not emulate the manner in which the VRU does voice recognition. Voice recognition is handled by the input plugin. I used Vosk (https://github.com/alphacep/vosk-api), so the voice recognition I'm sure will be better than an actual VRU.
  - Added some more functions to the input plugin spec, but they are all optional.
  - To specify that a VRU is plugged into a controller slot, I added a new Pak type, PLUGIN_VRU. I understand that the VRU is not a Pak, but this was the simplest and least intrusive way I could see to set this information.
  - Save states probably won't work. As you'll see in the YouTube video, the game uploads a list of words to the VRU. This would be part of the "state". So if a save state was created after that word list is uploaded to the input plugin, it won't be there when the state is loaded.

* Hey you Pikachu audio fixes
  - The audio in this game has always been choppy. It appears to be a timing issue, the AI DMAs take too long, so there are large pauses between the audio data. Shortening the AI DMA time fixes the issue. I added a new INI parameter: AiDmaModifier. This is a percentage of the DMA time, so a setting of 88 means "AI DMA time should be 88% of the regular time (shortened by 12%)". Until we have accurate timing emulation, this may be a helpful lever in games with audio problems.

References:
https://pastebin.com/6UiErk5h (zoinkity reverse engineering)
https://ultra64.ca/files/documentation/online-manuals/man/pro-man/pro26/index26.8.html

### DIFF
--- a/data/mupen64plus.ini
+++ b/data/mupen64plus.ini
@@ -6294,6 +6294,7 @@ CRC=D3F10E5D 052EA579
 Players=1
 SaveType=Eeprom 4KB
 Rumble=Yes
+AiDmaModifier=88
 
 [DCC316EFFC4928F5B0AE8D273D8024BF]
 GoodName=HiRes CFB Demo (PD)
@@ -11893,6 +11894,7 @@ CRC=3F245305 FC0B74AA
 Players=1
 SaveType=Eeprom 4KB
 Rumble=Yes
+AiDmaModifier=88
 
 [D0AE6C07AC0481EBA5FF9CE798A69574]
 GoodName=Pikachu Genki Dechu (J) [b1]
--- a/doc/emuwiki-api-doc/Mupen64Plus-v2.0-API-Versioning.mediawiki
+++ b/doc/emuwiki-api-doc/Mupen64Plus-v2.0-API-Versioning.mediawiki
@@ -129,3 +129,5 @@ This is the most complicated interface, because it involves 3 components: the vi
 ** added "M64CMD_ROM_SET_SETTINGS" command to allow setting ROM settings for the currently opened ROM until the ROM is closed.
 * '''CONFIG_API_VERSION''' version 2.3.2:
 ** add ConfigOverrideUserPaths() function to allow front-ends to override user paths.
+* '''INPUT_API_VERSION''' version 2.1.1:
+** add optional functions: SendVRUWord(), SetMicState(), ReadVRUResults(), ClearVRUWords(), SetVRUWordMask(). These functions add support for the VRU (Voice Recognition Unit). Also added a new int to the CONTROL struct: Type. Type can be CONT_TYPE_STANDARD (0) or CONT_TYPE_VRU (1).
--- a/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Plugin-API.mediawiki
+++ b/doc/emuwiki-api-doc/Mupen64Plus-v2.0-Plugin-API.mediawiki
@@ -224,6 +224,21 @@ These functions are present in all of the plugins.
 |-
 |<tt>void SDL_KeyUp(int keymod, int keysym);</tt>
 |Pass a SDL_KEYUP-style message to the input plugin
+|-
+|<tt>void SendVRUWord(uint16_t length, uint16_t *word, uint8_t lang);</tt>
+|Load a word/phrase into the VRU dictionary. length specifies the length of the buffer. lang is 0 for English and 1 for Japanese. Japanese words are encoding using S-JIS. English words are encoded using a phonetic alphabet explained here: https://pastebin.com/ajLzRLze
+|-
+|<tt>void SetMicState(int state);</tt>
+|A state of 1 tells the input plugin to enable the microphone. A state of 0 tells the input plugin to disable the microphone
+|-
+|<tt>void ReadVRUResults(uint16_t *error_flags, uint16_t *num_results, uint16_t *mic_level, uint16_t *voice_level, uint16_t *voice_length, uint16_t *matches);</tt>
+|Refer to https://pastebin.com/6UiErk5h and https://ultra64.ca/files/documentation/online-manuals/man/pro-man/pro26/index26.8.html for an explanation of these variables. This function is called after the player has finished speaking. The purpose is for the VRU/input plugin to tell the core if any matches were found against the dictionary that was provided earlier
+|-
+|<tt>void ClearVRUWords(uint8_t length);</tt>
+|Clears and initializes the word/phrase dictionary with a max length as provided by the variable
+|-
+|<tt>void SetVRUWordMask(uint8_t length, uint8_t *mask);</tt>
+|Assign a mask to the words in the dictionary. A bit of 1 means to recognize the word, 0 means to ignore the word. See https://ultra64.ca/files/documentation/online-manuals/man/pro-man/pro26/index26.8.html
 |}
 
 === Remove From Older Input API ===
--- a/projects/msvc/mupen64plus-core.vcxproj
+++ b/projects/msvc/mupen64plus-core.vcxproj
@@ -136,6 +136,7 @@
     <ClCompile Include="..\..\src\device\cart\is_viewer.c" />
     <ClCompile Include="..\..\src\device\cart\sram.c" />
     <ClCompile Include="..\..\src\device\controllers\game_controller.c" />
+    <ClCompile Include="..\..\src\device\controllers\vru_controller.c" />
     <ClCompile Include="..\..\src\device\controllers\paks\biopak.c" />
     <ClCompile Include="..\..\src\device\controllers\paks\mempak.c" />
     <ClCompile Include="..\..\src\device\controllers\paks\rumblepak.c" />
@@ -440,6 +441,7 @@
     <ClInclude Include="..\..\src\device\cart\is_viewer.h" />
     <ClInclude Include="..\..\src\device\cart\sram.h" />
     <ClInclude Include="..\..\src\device\controllers\game_controller.h" />
+    <ClInclude Include="..\..\src\device\controllers\vru_controller.h" />
     <ClInclude Include="..\..\src\device\controllers\paks\biopak.h" />
     <ClInclude Include="..\..\src\device\controllers\paks\mempak.h" />
     <ClInclude Include="..\..\src\device\controllers\paks\rumblepak.h" />
--- a/projects/msvc/mupen64plus-core.vcxproj.filters
+++ b/projects/msvc/mupen64plus-core.vcxproj.filters
@@ -360,6 +360,9 @@
     <ClCompile Include="..\..\src\device\controllers\game_controller.c">
       <Filter>device\controllers</Filter>
     </ClCompile>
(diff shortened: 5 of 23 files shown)


Jj0YzL5nvJ:
Majora's Mask too... with cheats.

https://youtu.be/lFa6GwuG-8I
https://youtu.be/VxK8XqYecSg

loganmc10:
Even though it does work with the NTSC ROM, it looks like it's still using the Japanese VRU, which probably won't work with this PR. The library I'm using for voice recognition (Vosk) doesn't support Japanese, and I don't speak or read Japanese, so I didn't feel I'd be able to test/troubleshoot it in a meaningful way.

loganmc10:
I spent some time testing/debugging Densha de Go. I got the VRU initialization much closer to the communication dump that zoinkity provided. I also got audio recognition working (if you say "depart" when you're leaving for instance you get "1 bonus voice point"). I've updated the Google Drive links in the first post.

loganmc10:
I decided I might as well go all the way with this. I got Japanese support working. I tested Hey You Pikachu, Densha de Go, and the VRU test ROM included in the SDK, they all work. As mentioned, the library I'm using in the input plugin doesn't have Japanese support, so you can't speak Japanese to it, but I did test that language recognition works. So it would work in Japanese if someone wanted to write an input plugin that supported it.

I've updated the .zips in the main post. I'd consider this complete at this point

bsmiles32 (commented):
Congratulation for this work !
I won't have time to do a full review of this work, but still I'll try to give some thought on what can be improved
- The AI DMA modifier : Not a huge fan, but I can understand why you did it. Still, I did some work few month ago on AI timings in https://github.com/bsmiles32/mupen64plus-core/tree/fix_ai_timings, and I was wondering if it could help avoiding the AI DMA modifier hack.
- I'm not sure putting all the HLE VRU logic in the input plugin is the best option. Sure it might be the easiest, but I would prefer if the HLE VRU logic could be built into the core even if it means adding the vosk dependency (it could be made optional). That way, any input plugin can benefit from VRU emulation. Also plugin can't have state, so moving the VRU HLE emulation to the core will allow you to support savestates.
- You should create a device/controllers/vru.c (or device/vru/vru.c) file and put all the vru related logic in there. Also, you should implement the vru as a proper joybus_device instead of using a game controller flavor. For the mouse I didn't felt like it was necessary to have a proper file and joybus_device because except for the paks and ID, the mouse andthe game_controller both behave the same way. The flavor could also be reused for the train controller (see https://github.com/bsmiles32/mupen64plus-core/tree/wip_train_support_rebased which was an old attempt at adding the train controller but was missing the input plugin support...). But the VRU is a very different kind of device, so it deserve a proper file / directory and joybus_device interface. And if in the future we want to extend VRU emulation with LLE it might be better to have a proper vru directory.
- Don't make the VRU a PAK because it is confusing, make it a joybus device.

Well these are my suggestions to improve the current PR.
In any case, very good job !

loganmc10:
Thanks for the feedback:
* For the AI DMA timing, I tried your branch, but unfortunately I didn't notice any improvement in the audio. It's still very choppy (you can hear it in the opening song that plays)
* There are 2 issues I can think of having the HLE VRU logic in the core:
  - Vosk is not an easy dependency to install. I had to extract the .so and .dll files from the Python package and then manually include them in my input plugin. I've looked around at a few alternatives, but it's actually a little hard to find speed recognition libraries that are maintained and have C interfaces as part of their releases. Many are focused on Python.
  - Related to the above, it may actually be easier on other platforms to use a different speech recognition library. For instance, Android has an android.speech.SpeechRecognizer library. It may be easier for the AE project to use that than figure out how to integrate Vosk on Android.
  - The HLE implementation relies on zoinkitys deciphering of the words (https://pastebin.com/JWwSVUS7). My input plugin contains these mappings in a header file. Depending on who you ask, these may be considered ROM assets and shouldn't be included in the core project.

For points 3 and 4 I agree, I can clean that up. I was under the impression that anything that plugged into a controller port was a "game controller" and the flavors were used to specify the different types, but I can put it in it's own file.

In order for the input plugin to specify that a VRU is connected, can I extend the CONTROL struct (https://github.com/mupen64plus/mupen64plus-core/blob/db56853707ca6f4223adf5e7f0b7725f90d3a0a3/src/api/m64p_plugin.h#L139-L143) to include another int for "Type", where 0 would be a regular controller, and 1 could be mouse, 2 could be VRU, etc? This should be backwards compatible if I add it to the end of the struct I believe?

The other option might be to repurpose the RawData int, rename it to Type, and say that 0=Regular Controller, 1=RawData, 2=VRU, 3=Mouse, etc... which should still be backwards compatible.

vilhelmgray:
>     * There are 2 issues I can think of having the HLE VRU logic in the core:
>       
>       * Vosk is not an easy dependency to install. I had to extract the .so and .dll files from the Python package and then manually include them in my input plugin. I've looked around at a few alternatives, but it's actually a little hard to find speech recognition libraries that are maintained and have C interfaces as part of their releases. Many are focused on Python.
>       * Related to the above, it may actually be easier on other platforms to use a different speech recognition library. For instance, Android has an android.speech.SpeechRecognizer library. It may be easier for the AE project to use that than figure out how to integrate Vosk on Android.

I don't know whether it would make sense to integrate the HLE VRU logic into the core, but regarding the difficulty and limitations of Vosk, would [Julius](https://github.com/julius-speech/julius) be a viable alternative? It's a [BSD-licensed C speech recognition library maintained by the prestigious Kyoto University](https://julius.osdn.jp/en_index.php?q=index-en.html) and it recognizes both Japanese and English (thanks to help from the Mozilla foundation) -- actually it can recognize more languages, but I think those are the only two relevant to the Nintendo 64 VRU. Furthermore, Julius is cross-platform and runs on Linux, Windows, Android, Mac, etc, so it shouldn't be a problem to build for those.

It looks like you can just pass audio to the library and it'll return an array of recognized sentences, which you can then grep for the words the game expects. Here's the [documentation to the JuliusLib API](https://www.sp.nitech.ac.jp/~ri/julius-dev/doxygen/julius/4.0/en/index.html) along with a [simple example program showing how to use it](https://www.sp.nitech.ac.jp/~ri/julius-dev/doxygen/julius/4.0/en/julius-simple_8c-source.html).

loganmc10:
There are probably pros and cons to various speech recognition libraries. For instance, the Julius English voice model is 479MB, vs 40MB for Vosk. However, it does look there are Ubuntu packages available for Julius.

Overall this is why I think it would be best to leave the voice recognition up to the input plugin. As I mentioned before, there are platforms like Android that have native speech recognition libraries that could possibly be used.

vilhelmgray:
That's a fair point, although Julius might be a more advanced engine, the downside of a large voice model would make it impractical for platforms where disk space is at a premium. Perhaps a plugin setup would be better after all: platforms that can benefit from more advanced engines can use them as they like, whereas platforms that require something simpler can select such plugins as necessary.

bsmiles32:
In the core you have backends that would allow to choose at compile time ans runtime what voice recognition engine to use. You juste have to create a dedicated backend interface and implement a vosk and/or Julius implementation. Just make sure that no emulation state is kept inside the backend as it can't be savestated

amatho:
> In order for the input plugin to specify that a VRU is connected, can I extend the CONTROL struct (
> 
> https://github.com/mupen64plus/mupen64plus-core/blob/db56853707ca6f4223adf5e7f0b7725f90d3a0a3/src/api/m64p_plugin.h#L139-L143
> ) to include another int for "Type", where 0 would be a regular controller, and 1 could be mouse, 2 could be VRU, etc? This should be backwards compatible if I add it to the end of the struct I believe?

This is unfortunately not backwards compatible, as the `CONTROL` struct is part of an array in `CONTROL_INFO` here:
https://github.com/mupen64plus/mupen64plus-core/blob/db56853707ca6f4223adf5e7f0b7725f90d3a0a3/src/api/m64p_plugin.h#L171-L174
This would increase the size of the array (to 4 * 16 bytes), while older plugins would assume it's shorter (4 * 12 bytes). This is what I tried to convey in my bad explanation in loganmc10/m64p#206 😉.

> The other option might be to repurpose the RawData int, rename it to Type, and say that 0=Regular Controller, 1=RawData, 2=VRU, 3=Mouse, etc... which should still be backwards compatible.

However, this should be backwards compatible.

vilhelmgray:
Because the VRU vocabulary list is small, a future optimization is to train custom voice models tuned specifically to the vocabulary list for the N64 games that use the VRU. I expect this would significantly reduce the size of the language models that are included with mupen64plus by not having to recognize such a large subset of a language's vocabulary, instead only concerning itself with the small number of words the VRU is interested in.

See the ["Training your own model"](https://alphacephei.com/vosk/models) section on the VOSK website, and the ["About Models"](https://julius.osdn.jp/en_index.php?q=index-en.html#about_models) section of the Julius website.

jet082:
If you need someone who can test Japanese, I am fairly fluent.

jet082:
The builds on the google drive are also no longer operational.

richard42:
@loganmc10 are you finished with your work on this, or is this still WIP?

loganmc10:
I'm finished with it, I've just been rebasing it against master. But I know @bsmiles32 had some disagreement about the implementation. I've been including it with my package for quite some time.

There is some disagreement about whether the voice recognition library should like inside the core or the input plugin basically, and that decision would affect the API design, so it's not just a matter of code organization.

bsmiles32:
I'm sorry but I won't have time to work on this project anymore (at least for the coming months/years). So you can disregard my opinion on this and other PR. If you feel like merging it no problem. It would be unfair of me to block other contribution given that I won't be around to help.
Sorry it took me so long to realize that I won't have time for this project anymore.

vilhelmgray:
It might not be so bad to tentatively merge this PR so that mupen64plus has at least some sort of VRU support. I think in the longterm it would be good to transition this VRU functionality into the mupen64plus core -- thus allowing better support for savestates and such as mentioned earlier in this thread -- but that can happen in the future when someone with the motivation and familiarity with the code is available to spearhead such an effort.

MeleeGamer:
The download link doesn't work anymore.

loganmc10:
Simple64 includes VRU support based on this work:

https://github.com/simple64/simple64/releases

PR #876, opened 2021-06-23
IS Viewer support

loganmc10:
IS-Viewer 64 was a physical debugging tool (https://n64squid.com/homebrew/n64-sdk/development-hardware/). CEN64 has had support for a while (https://github.com/n64dev/cen64/blob/master/pi/is_viewer.c). I referenced/copied their implementation for mupen64plus.

When a ROM has IS64 support, the output is printed as a DebugMessage:

image

This is used today by homebrew ROM developers to debug their games.

I don't think this should interfere with any existing games. The address space is inside ```dom1 addr2``` (cart ROM), but I map the IS Viewer before the cart ROM. As I understand it, if the current ROM includes that address space, it'll overwrite the IS Viewer mapping. @bsmiles32

### DIFF
--- a/projects/msvc/mupen64plus-core.vcxproj
+++ b/projects/msvc/mupen64plus-core.vcxproj
@@ -133,6 +133,7 @@
     <ClCompile Include="..\..\src\device\cart\cart_rom.c" />
     <ClCompile Include="..\..\src\device\cart\eeprom.c" />
     <ClCompile Include="..\..\src\device\cart\flashram.c" />
+    <ClCompile Include="..\..\src\device\cart\is_viewer.c" />
     <ClCompile Include="..\..\src\device\cart\sram.c" />
     <ClCompile Include="..\..\src\device\controllers\game_controller.c" />
     <ClCompile Include="..\..\src\device\controllers\paks\biopak.c" />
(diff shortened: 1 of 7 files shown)


mikeryan:
Hi, I'm the original author of the code you borrowed from CEN64. I notice that you didn't include any of the iconv code. In order to interface with retail games, you'll need to convert the text from whatever locale it was originally in (typically EUC-JP) into UTF-8 for it to display correctly in your GUI. I'm not sure if iconv is the best option for mupen, but you'll definitely need something. Best of luck and thanks for porting this over.

loganmc10:
I mostly had homebrews in mind. I assume the only games that use EUC-JP are the OG Zelda debug ROMs.

These messages are actually passed to a front-end, so technically the front-end could probably detect EUC-JP encoding and display it accordingly, without modifying this code.

jkbenaim:
This PR looks good. It makes sense for the core to pass the data to the frontend without conversion. Any changes to the frontend would be done in a separate PR.

loganmc10:
Unless @bsmiles32 or @richard42 have any objections I think this should be pretty safe to merge. The code is basically self contained, so I don't think there is much of a chance for regressions.

richard42:
I'm fine with this.

Narann:
Thanks @richard42 !

PR #878, opened 2021-06-23
Make 4K EEPROM default save type; support carts with no EEPROM

meeq:
Proper fix for workaround in https://github.com/DragonMinded/libdragon/pull/125#pullrequestreview-690939199

@rasky  @sp1187 @loganmc10 

The Joybus response values have been confirmed on real hardware using the following test ROM which dumps the EEPROM status input and output data:  [eepromtest.zip](https://github.com/mupen64plus/mupen64plus-core/files/6705089/eepromtest.zip)

I was not able to run this test ROM on Mupen64Plus; all of my libdragon example ROMs get stuck at "Mupen64Plus Started...":
mupen

I'd appreciate some assistance on getting this to run. I was able to [test this ROM successfully with cen64](https://github.com/n64dev/cen64/pull/203).


loganmc10:
I'll test this in a bit, I think it would be cleaner to just leave the type as a uint16_t and do
```
rx_buf[0] = (uint8_t)(cart->eeprom.type >> 0);
rx_buf[1] = (uint8_t)(cart->eeprom.type >> 8);
rx_buf[2] = (uint8_t)(cart->eeprom.type >> 0);
```

(Just repeat the first byte, which is 0 when present and 0xff when absent)

loganmc10:
The other thing I'll say: the way this is written, it defaults to "no eeprom". This means that any ROM not in the INI won't support eeprom saves. This might be a problem for ROM hacks/new ROMs (like a new version of Smash Remix or something).

I think it would be better to retain the default of 4k eeprom. I don't think we've run across a game that failed because it said it had eeprom when it shouldn't.

loganmc10:
Here is the result of the test ROM:

controller_test-000

I think that looks right, but as I said, I think we should still default to 4k eeprom. It's a fairly safe assumption that if a commercial game is probing that address, it expects the eeprom to be there.

meeq:
I have an updated EEPROM test ROM that exercises a variety of EEPROM probe command formats: [eepromtest.zip](https://github.com/mupen64plus/mupen64plus-core/files/6705828/eepromtest.zip)

This branch has been updated to reflect a more-accurate understanding of how the real hardware handles EEPROM checking: If there is no EEPROM, the `recv` buffer will not be modified.

meeq:
As for the "user-experience question" of whether it's actually a good idea to make this change: it's subjective.

According to [this list of save types](http://micro-64.com/database/gamesave.shtml), 4K EEPROM is the second-most popular save format for N64 titles. There is a lot of sense in the argument that a user-friendly emulator should assume 4K EEPROM if no other save type is specified, since it (almost always) won't hurt and has a decent chance of helping.

My only argument against is that homebrew/mods that want other save types (such as Smash Remix which uses SRAM) will still be a broken experience that requires the user to select the right save type. So it will "just work" some of the time, but silently fail other times.

loganmc10:
```
-    JDT_EEPROM_4K        = 0x8000,  /* 4k EEPROM */
-    JDT_EEPROM_16K       = 0xc000,  /* 16k EEPROM */
+    JDT_EEPROM_4K        = 0x0080,  /* 4k EEPROM */
+   JDT_EEPROM_16K       = 0x00c0,  /* 16k EEPROM */
```
Did you mean to flip the order of these bytes? I don't think that's right.

For the save picking, I would do something like:
```
    uint16_t eeprom_type = JDT_EEPROM_4K;
    switch (ROM_SETTINGS.savetype) {
        case SAVETYPE_EEPROM_16KB:
            eeprom_type = JDT_EEPROM_16K;
            break;
        case SAVETYPE_SRAM:
            eeprom_type = JDT_EEPROM_NONE;
            break;
        case SAVETYPE_FLASH_RAM:
            eeprom_type = JDT_EEPROM_NONE;
            break;
        case SAVETYPE_NONE:
            eeprom_type = JDT_EEPROM_NONE;
            break;
    }
```
This way if the ROM specifically has flash, sram, or none set in the INI, it will set the eeprom type to none. Otherwise, it'll default to 4k because of the change you made to rom.c

Jj0YzL5nvJ (commented):
DK64 demos look better (less sped up), but now doesn't initialize the save file.

meeq:
Sorry, I definitely messed up by flipping the bytes of the EEPROM identifiers. Let me try to figure out how to get my libdragon ROMS running on this branch so I can run the test ROM and verify these changes myself.

meeq:
I have attached the libdragon test suite, which includes EEPROM Joybus command tests: [testrom_emu.zip](https://github.com/mupen64plus/mupen64plus-core/files/6709720/testrom_emu.zip)

Line numbers for `test_eeprom` failures are mapped to this file: [libdragon/tests/test_eeprom.c](https://github.com/DragonMinded/libdragon/blob/9b1755c108bd4327d23eec02fa899525dd3b3250/tests/test_eeprom.c)

There are several other failures unrelated to EEPROM that are out of scope for this PR.

Without the changes in this PR, it is not possible to exercise the "No EEPROM present" code-path since Mupen64Plus will always have a 4K EEPROM (unless 16K EEPROM is specified)

src/api/m64p_types.h
```diff
@@ -224,11 +224,11 @@ typedef enum
...
 typedef enum
 {
-    SAVETYPE_EEPROM_4KB = 0,
-    SAVETYPE_EEPROM_16KB,
+    SAVETYPE_EEPROM_4K = 0,
+    SAVETYPE_EEPROM_16K,
     SAVETYPE_SRAM,
     SAVETYPE_FLASH_RAM,
-    SAVETYPE_CONTROLLER_PACK,
+    SAVETYPE_CONTROLLER_PAK,
```
meeq, line 231: "Correct" spelling is "Pak": https://en.wikipedia.org/wiki/Nintendo_64_accessories#Controller_Pak

src/api/m64p_types.h
```diff
@@ -224,11 +224,11 @@ typedef enum
 
 typedef enum
 {
-    SAVETYPE_EEPROM_4KB = 0,
-    SAVETYPE_EEPROM_16KB,
+    SAVETYPE_EEPROM_4K = 0,
```
meeq, line 227: KB is not accurate; the EEPROM sizes are in kilo*bits*: http://n64devkit.square7.ch/pro-man/pro26/26-06.htm

>There are two types of EEPROM according to their capacities. One is the 4k-EEPROM. The capacity of this type is 4 kilobits (64 words * 64 bits = 512 bytes). The other is the 16k-EEPROM. The capacity of this type is 16 kilobits (256 words * 64 bits = 2048 bytes).

src/device/cart/cart.c
```diff
@@ -45,10 +45,12 @@ static void process_cart_command(void* jbd,
     case JCMD_STATUS: {
         JOYBUS_CHECK_COMMAND_FORMAT(1, 3)
 
-        /* set type and status */
-        rx_buf[0] = (uint8_t)(cart->eeprom.type >> 0);
-        rx_buf[1] = (uint8_t)(cart->eeprom.type >> 8);
-        rx_buf[2] = 0x00;
+        if (cart->eeprom.type) {
```
meeq, line 48: This behavior has been tested on real hardware; if there is no EEPROM, the receive buffer is untouched.

src/api/m64p_types.h
```diff
@@ -224,12 +224,12 @@ typedef enum
 
 typedef enum
 {
-    SAVETYPE_EEPROM_4KB = 0,
```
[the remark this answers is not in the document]
meeq: https://github.com/mupen64plus/mupen64plus-core/pull/878/commits/e005ab71bd4f426a2bd93dabb7e984cff7078220

loganmc10:
This LGTM, we've got a lot of pending PRs, if we don't start merging things we'll end up with conflicts soon enough

Jj0YzL5nvJ (approved):
I have been using this PR as a patch and have not experienced any noticeable issues.

Narann:
Here we go.

turntable97:
I just can't manage to save on a DK64 Widescreen hacked ROM.
How is this possible? Do I need to make a specific setting in RetroArch to make it work?
I'd appreciate any help I can get!

PR #880, opened 2021-07-07
Fix 64DD Write Address Check Regression

LuigiBlood:
The DMA Address check fix only worked for reads but writes did not trigger dd_update_bm, and therefore no interrupts were issued

I made it check the entire buffer space so there's no room for errors: It works.


m4xw (approved):
lgtm

Issue #886, opened 2021-07-17
Road rush 64 black screen on new game with new dynarec


fzurita:
@Gillou68310 It looks Road Rush 64 got broken by recent new dynarec changes. It works fine in interpreter. 

Starting a new game causes a black screen after entering a name.

thekovic:
Same issue as Indiana Jones, it seems https://github.com/loganmc10/m64p/issues/202

Gillou68310:
@fzurita can you confirm that https://github.com/mupen64plus/mupen64plus-core/pull/887 fixes both road rush and indiana?

fzurita:
Road Rush 64 is fixed. @thekovic Can you test Indiana jones with this?

thekovic:
I tested Indiana Jones and I can confirm the issue is fixed

PR #887, opened 2021-07-19
Fix https://github.com/mupen64plus/mupen64plus-core/issues/886

Gillou68310:


fzurita:
I tested Road Rush 64 and it works. I tested Indian jones and didn't see any issues either, but I didn't have the original indiana jones issue described in #886 .

fzurita:
Actually, I tested Indiana Jones before this fix and it's not working, so this fix does take care of that.

Narann:
Thanks all!

Issue #889, opened 2021-07-26
Dinosaur Planet crashes with new dynarec


loganmc10:
See https://github.com/loganmc10/m64p/issues/204

Confirmed to crash when starting a new game on x64. Doesn't crash with cached interpreter

PR #890, opened 2021-07-27
Fix https://github.com/mupen64plus/mupen64plus-core/issues/889

Gillou68310:
@loganmc10 This is basically a no brain fix. The whole function is unreadable and need a major refactoring. I'm not even sure all cases are correctly handled right now...
Anyway this should do it for the time being.

The issue was a missing pop register, leading to stack growing indefinitely.


loganmc10:
> The whole function is unreadable and need a major refactoring

This describes the entire new dynarec to me :)

LGTM

PR #898, opened 2021-09-11
Add RISC-V Support

### LINKED ISSUE #897: CPU type "riscv64" not supported.
Upon attempting to build from git, I get the following error.
```
Makefile:139: *** CPU type "riscv64" not supported.  Please file bug report at 'https://github.com/mupen64plus/mupen64plus-core/issues'.  Stop.
```

archanox:
Resolves #897 for mupen64plus-core

### DIFF
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -135,6 +135,13 @@ ifneq ("$(filter aarch64,$(HOST_CPU))","")
     PIC ?= 1
     NEW_DYNAREC := 1
 endif
+ifneq ("$(filter riscv64,$(HOST_CPU))","")
+    CPU := RISCV64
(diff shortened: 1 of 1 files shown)


Narann:
Looks good to me, but I would love to know how does it run?

archanox:
Without a dynarec, not well. But I'll put a video capture up hopefully today.

archanox:
> Looks good to me, but I would love to know how does it run?

https://youtu.be/5vn0IyB2xqM
Judge for yourself, but I wouldn't deem this very playable with the cached interpreter. But the good news is that the port from MIPS to RISC-V dynarec should be relatively straight forward.

Narann:
Thanks for the video! :)

PR #909, opened 2021-11-11
Fix IS Viewer crash

Rosalie241:
Fixes a segfault in [The Missing Link](https://www.romhacking.net/hacks/5334/) when entering a door.

### DIFF
--- a/src/device/cart/is_viewer.c
+++ b/src/device/cart/is_viewer.c
@@ -33,8 +33,8 @@
 
 void poweron_is_viewer(struct is_viewer* is_viewer)
 {
-    memset(is_viewer->data, 0, 0x1000);
-    memset(is_viewer->output_buffer, 0, 0x200);
+    memset(is_viewer->data, 0, IS_BUFFER_SIZE);
+    memset(is_viewer->output_buffer, 0, IS_BUFFER_SIZE);
     is_viewer->buffer_pos = 0;
 }
 
@@ -55,6 +55,16 @@ void write_is_viewer(void* opaque, uint32_t address, uint32_t value, uint32_t ma
     {
         if (word > 0)
         {
+            /* make sure we don't overflow the buffer */
+            if (is_viewer->buffer_pos + word > IS_BUFFER_SIZE)
+            {
+                /* reset buffer */
+                memset(is_viewer->output_buffer, 0, IS_BUFFER_SIZE);
+                is_viewer->buffer_pos = 0;
+                DebugMessage(M64MSG_WARNING, "IS64: prevented buffer overflow, cleared buffer");
+                return;
+            }
+
             memcpy(&is_viewer->output_buffer[is_viewer->buffer_pos], &is_viewer->data[0x20], word);
             is_viewer->buffer_pos += word;
             char* newline = memchr(is_viewer->output_buffer, '\n', is_viewer->buffer_pos);
--- a/src/device/cart/is_viewer.h
+++ b/src/device/cart/is_viewer.h
@@ -25,10 +25,12 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#define IS_BUFFER_SIZE 0x1000
+
 struct is_viewer
 {
-    char data[0x1000];
-    char output_buffer[0x200];
+    char data[IS_BUFFER_SIZE];
+    char output_buffer[IS_BUFFER_SIZE];
     uint32_t buffer_pos;
 };
 


loganmc10:
I think it would be better to fix the logic behind buffer_pos if that is wrong

Rosalie241:
> I think it would be better to fix the logic behind buffer_pos if that is wrong

Will you do that then? I'm unfamiliar with this code and this seemed like the easiest fix.

loganmc10:
I'm guessing the size of is_viewer->output_buffer just needs to be increased. The current size (0x200) was taken from CEN64:

https://github.com/n64dev/cen64/blob/d6b5b043956af205d5f6d28eb321ca796adfa021/pi/is_viewer.c#L7-L8

and their code says that size was arbitrarily chosen, I would just see what the value of is_viewer->buffer_pos is when the crash occurs, and increase the size of is_viewer->output_buffer accordingly

Rosalie241:
> I'm guessing the size of is_viewer->output_buffer just needs to be increased. The current size (0x200) was taken from CEN64:
> 
> https://github.com/n64dev/cen64/blob/d6b5b043956af205d5f6d28eb321ca796adfa021/pi/is_viewer.c#L7-L8
> 
> and their code says that size was arbitrarily chosen, I would just see what the value of is_viewer->buffer_pos is when the crash occurs, and increase the size of is_viewer->output_buffer accordingly

I added some debug logging and well: 
```
buffer_pos = 7166750
```

This happens because if you look at the previous lines
```
memcpy(&is_viewer->output_buffer[is_viewer->buffer_pos], &is_viewer->data[0x20], word);
```

this overflows and changes the `buffer_pos` variable (because it's an array in a struct), on CEN64 it also overflows because of the static buffer size (`buffer_pos` is `636` on CEN64), but on CEN64 it's allocated with `calloc`, so it doesn't push into the `buffer_pos` variable and doesn't crash there (though it also overflows there).

I'll modify this PR to change the code to make use of `malloc`/`realloc` and never have this happen again.

loganmc10:
Thats fine, you could probably also just change the if statement:
```
if (word > 0)
        {
```
to some thing like
```
if (word > 0 && is_viewer->buffer_pos < 0x200 + word)
        {
```
and maybe add an else that would emit a DebugMessage telling you you would overflow the buffer. Then it's never going to try to memcpy into a bad address space

Rosalie241:
> Thats fine, you could probably also just change the if statement:
> 
> ```
> if (word > 0)
>         {
> ```
> 
> to something like
> 
> ```
> if (word > 0 && is_viewer->buffer_pos < 0x200 - word)
>         {
> ```
> 
> and maybe add an else that would emit a DebugMessage telling you you would overflow the buffer. Then it's never going to try to memcpy into a bad address space

I've updated the PR to dynamically realloc if needed instead, which is a much cleaner solution I think.

loganmc10:
```
if (is_viewer->buffer_pos + word >= is_viewer->buffer_len)
            {
                is_viewer->buffer_len += IS_CHUNK_SZ;
                is_viewer->output_buffer = realloc(is_viewer->output_buffer, is_viewer->buffer_len);
```
How is this guaranteed to solve the problem though? What if buffer_len is still too small? IS_CHUNK_SZ is hardcoded to 0x200

This is such a rarely used feature, I would just figure out how big the buffer needs to be for A Missing Link and increase it from 0x200 to whatever it needs to be.

Rosalie241:
> ```
> if (is_viewer->buffer_pos + word >= is_viewer->buffer_len)
>             {
>                 is_viewer->buffer_len += IS_CHUNK_SZ;
>                 is_viewer->output_buffer = realloc(is_viewer->output_buffer, is_viewer->buffer_len);
> ```
> 
> How is this guaranteed to solve the problem though? What if buffer_len is still too small? IS_CHUNK_SZ is hardcoded to 0x200

fixed it in the latest push.

> This is such a rarely used feature, I would just figure out how big the buffer needs to be for A Missing Link and increase it from 0x200 to whatever it needs to be.

but then in the future we might have the same problem again, I dislike a band-aid solution so I'd rather have a permanent solution that'll keep on working.

loganmc10:
Given that the IS viewer address space size is 0xfff:
https://github.com/mupen64plus/mupen64plus-core/blob/db0f7365f2c2b8bbae523b2503e660d09d4a2a85/src/device/device.c#L154 it probably shouldn't be able to write outside that.

I never tried A Missing Link, so I don't know how much data it tries to write, but I'd be surprised if it was more than 0x1000. I assume it should be safe to just set the output buffer size to 0x1000. If it actually needed more than that, then we've got the address space mapping wrong for the IS Viewer and that should be fixed as well.

Rosalie241:
Good point, I'll do that tomorrow-ish then, thank you

Rosalie241:
Okay, pushed, the patch is a lot more simple right now, though I added an overflow check, and when it'll overflow, it'll reset the buffer & buffer_pos and show a warning.

PR #912, opened 2021-11-30
CI/CD: Implement GitHub Actions and public nightly builds

Jj0YzL5nvJ:
You can see the changes taking effect from my repository in the `ci-tests` branch:
https://github.com/Jj0YzL5nvJ/mupen64plus-core/actions
https://github.com/Jj0YzL5nvJ/mupen64plus-core/releases

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,288 @@
+name: Core
+
+on:
+  push:
+    paths-ignore:
+      - '**/*.md'
+      - '.{gitattributes,gitignore,travis.yml}'
+      - 'appveyor.yml,README'
+  pull_request:
+    paths-ignore:
+      - '**/*.md'
+      - '.{gitattributes,gitignore,travis.yml}'
+      - 'appveyor.yml,README'
+  workflow_dispatch:
+jobs:
+
+  Linux:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - cc: GCC
+          - cc: Clang
+    name: Linux / ${{ matrix.cc }} / x64 | x86
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Get build dependencies and arrange the environment
+        run: |
+          sudo dpkg --add-architecture i386
+          sudo apt-get update
+          sudo apt-get -y install libfreetype6-dev libgl1-mesa-dev libglu1-mesa-dev libpng-dev libsdl1.2-dev libsdl2-dev nasm zlib1g-dev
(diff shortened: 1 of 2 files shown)


Narann:
Thanks a lot for the hard work!

(I comment here for all other PRs)

Code looks OK to me. Here are my concerns, it's open to discussion and those would not prevent any merge :) :

I thinks we should follow [original release name conventions](https://github.com/mupen64plus/mupen64plus-core/releases).

```
mupen64plus-bundle-linux32-2.5.9.tar.gz 1.67 MB
mupen64plus-bundle-linux64-2.5.9.tar.gz 1.65 MB
mupen64plus-bundle-osx-2.5.9.zip 2.56 MB
mupen64plus-bundle-src-2.5.9.tar.gz 2.74 MB
mupen64plus-bundle-win32-2.5.9.zip 2.36 MB
mupen64plus-core-src-2.5.9.tar.gz 1.54 MB
mupen64plus-core-src-2.5.9.tar.gz.asc 
```

In your case this would mean:

```
mupen64plus-core-msvc-x64-g0c39f7f.zip
mupen64plus-core-msvc-x86-g0c39f7f.zip
mupen64plus-core-msys2-x64-g0c39f7f.zip
mupen64plus-core-msys2-x86-g0c39f7f.zip
```

I would also add few archive for potential distribution package managers because we could start to use your config for linux nightly built binaries:

```
mupen64plus-core-src-g0c39f7f.zip
mupen64plus-core-src-g0c39f7f.zip.asc
```

What do you think? Any advise @richard42 ?

Thanks a lot for the hard work, Windows binaries are often asked for and we never took the time to do a proper release.

Keep the good work!

Jj0YzL5nvJ:
I'm fine with the naming convention part.

Regarding the distribution of _useful_ binaries for Linux, a complete redesign will be necessary. Maybe generate an AppImage? In any case, that is out of my range of ability at the moment.
The distribution of signed source code from a tagged github commit is redundant in my opinion and I find its usefulness questionable. I have no idea how to use *.asc key files as checksum (only connections with secret/public keys). I think `sha256` and `sha512` would be useful for Windows binaries, better than nothing to say the least.

I will wait a while to see if more observations accumulate before introducing the corrections, in the meantime I will test how to incorporate the aforementioned checksums.

Jj0YzL5nvJ:
[Prototype...](https://github.com/Jj0YzL5nvJ/mupen64plus-rsp-z64/releases/tag/nightly-build) I have no idea why [yt-dlp](https://github.com/yt-dlp/yt-dlp) uses overcomplicated hashing code.

Narann:
Thanks!

The hash should be to the end: `mupen64plus-rsp-z64-ge4a135f-msys2-x86.zip` -> `mupen64plus-rsp-z64-msys2-x86-ge4a135f.zip`

Jj0YzL5nvJ:
Whoops, that went completely unnoticed =P
Check again.

PR #920, opened 2022-01-24
Allow frontends to know what DD ROM region they should use

Rosalie241:
This adds a function called `set_dd_rom_region` to the media loader, allowing frontends to easily switch 64DD ROM based on DD disk region.

### DIFF
--- a/doc/emuwiki-api-doc/Mupen64Plus-v2.0-headers.mediawiki
+++ b/doc/emuwiki-api-doc/Mupen64Plus-v2.0-headers.mediawiki
@@ -162,6 +162,12 @@
    */
   char* (*get_gb_cart_ram)(void* cb_data, int controller_num);
 
+  /* Allow the frontend to know what DD IPL ROM region file to load
+   * cb_data: points to frontend-defined callback data.
+   * region: a region from m64p_system_type
+   */
+  void (*set_dd_rom_region)(void* cb_data, uint8_t region);
+
   /* Allow the frontend to specify the DD IPL ROM file to load
    * cb_data: points to frontend-defined callback data.
    * Returns a NULL-terminated string owned by the core specifying the DD IPL ROM filename to load
--- a/src/api/m64p_types.h
+++ b/src/api/m64p_types.h
@@ -196,6 +196,12 @@ typedef struct {
    */
   char* (*get_gb_cart_ram)(void* cb_data, int controller_num);
 
+  /* Allow the frontend to know what DD IPL ROM region file to load
+   * cb_data: points to frontend-defined callback data.
+   * region: a region from m64p_system_type
+   */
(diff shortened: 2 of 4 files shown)


DarioSamo:
This PR seems to have introduced a runtime error where `dd_rom_size` can now be uninitialized in the `main_run` function because `load_dd_rom` has been moved inside the branch and executed only if `load_dd_disk` succeeds.

It seems `init_device` relies on `dd_rom_size` being correctly initialized later, so this seems to introduce a random chance of incorrect behavior based on whether `dd_rom_size` happens to be 0 or not.

I haven't actually ran into the error myself, but MSVC's debugger caught this pretty quickly on the Debug configuration and would not allow me to run further until it's fixed.

The obvious solution to me seems to just handle the else case on the branch mentioned above, but I figured I'd report it first to the original submitter before making a PR myself.

Rosalie241:
> This PR seems to have introduced a runtime error where `dd_rom_size` can now be uninitialized in the `main_run` function because `load_dd_rom` has been moved inside the branch and executed only if `load_dd_disk` succeeds.
> 
> It seems `init_device` relies on `dd_rom_size` being correctly initialized later, so this seems to introduce a random chance of incorrect behavior based on whether `dd_rom_size` happens to be 0 or not.

Nice catch, I didn't think about that at all and I think GCC didn't warn (or I overlooked it) about this, so I apologize for introducing such an issue.

> I haven't actually ran into the error myself, but MSVC's debugger caught this pretty quickly on the Debug configuration and would not allow me to run further until it was fixed.
> 
> The obvious solution to me seems to just handle the else case on the branch mentioned above, but I figured I'd report it first to the original submitter before making a PR myself.

Yeah seems like a simple fix, thank you for finding the issue and opening a PR :heart:

DarioSamo:
> > This PR seems to have introduced a runtime error where `dd_rom_size` can now be uninitialized in the `main_run` function because `load_dd_rom` has been moved inside the branch and executed only if `load_dd_disk` succeeds.
> > It seems `init_device` relies on `dd_rom_size` being correctly initialized later, so this seems to introduce a random chance of incorrect behavior based on whether `dd_rom_size` happens to be 0 or not.
> 
> Nice catch, I didn't think about that at all and I think GCC didn't warn (or I overlooked it) about this, so I apologize for introducing such an issue.
> 
> > I haven't actually ran into the error myself, but MSVC's debugger caught this pretty quickly on the Debug configuration and would not allow me to run further until it was fixed.
> > The obvious solution to me seems to just handle the else case on the branch mentioned above, but I figured I'd report it first to the original submitter before making a PR myself.
> 
> Yeah seems like a simple fix, thank you for finding the issue and opening a PR ❤️

Thanks for the prompt reply, I've opened a new PR with the fix.

PR #926, opened 2022-02-13
Add ROM entries for SM64 splitscreen hack

loganmc10:
I'm aware of 2 different ROMs for this hack, unfortunately, I couldn't find any version information for them. I don't know of any special requirements so I just referenced them to the main SM64 rom


loganmc10:
The reason for this is that these are popular ROMs on m64p netplay, the goodname is required for them to be announced properly:
image

PR #946, opened 2022-04-12
Use a predictable seed when using netplay

loganmc10:
I hadn't thought of this for the mempak PR, but I think there is a small possibility of netplay desyncs because of the randomized mempak IDs.

If the mempak save file already exists, it won't be a problem, because netplay will sync P1's mempak with the other players, but if no mempak save file exists, netplay will just let each player create their own file.

Because of this randomization, each mempak will be slightly different. I don't know if this would cause desyncs, because I can't say for sure if a game reads a mempak to create some RNG or something.

This PR uses a constant seed for generating the mempak ID when netplay is active.

@bsmiles32

### DIFF
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -1486,7 +1486,8 @@ m64p_error main_run(void)
     }
 
     /* Seed MPK ID gen using current time */
-    l_mpk_idgen = xoshiro256pp_seed((uint64_t)time(NULL));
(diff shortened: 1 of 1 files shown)


bsmiles32:
Forcing a known seed when in netplay (or TAS) is indeed valuable. LGTM

PR #961, opened 2022-07-11
CI/CD: Update

Jj0YzL5nvJ:


projects/unix/Makefile
```diff
@@ -806,7 +804,7 @@ uninstall:
 	$(RM) "$(DESTDIR)$(SHAREDIR)/mupencheat.txt"
 
 clean:
-	$(RM) -r $(TARGET) $(SONAME) $(OBJDIR) $(SRCDIR)/asm_defines/asm_defines_*
+	$(RM) -r $(TARGET) $(SONAME) _obj $(OBJDIR) $(SRCDIR)/asm_defines/asm_defines_*
```
richard42, line 807: I don't understand why the _obj directory is added here

richard42 (changes requested):

projects/unix/Makefile
Jj0YzL5nvJ: The "_obj" folder will always be a temporary object. The variable "$(OBJDIR)" does not always equate to "_obj" in this build script.
Similar changes were accepted in rsp-z64 and rsp-cxd4 before...

richard42: ok I guess it doesn't hurt anything

PR #963, opened 2022-07-20
Added GoldenEye X and Perfect Dark Plus to INI

Graslu:
GoldenEye X and Perfect Dark Plus entries added to mupen64plus.ini


loganmc10:
For the ```Alternate Title``` field, you can't put arbitrary fields in the INI, or the Core will report this error:

```
Core Warning: ROM Database: Unknown property on line 5932
```

You can add comments though, any line starting with a ```;``` is a comment, so I would do:

```
[67F2E18ED18A5F0EA3FE2D913E3F1786]
GoodName=GoldenEye X
; Alternate Title=GoldenEye X (5e Clouds)
CRC=FDF95D15 618546CA
RefMD5=E03B088B6AC9E0080440EFED07C1E40F
```

Graslu:
Alright I'll do that later, I am once again having issues with GitHub reading my ini file as if I added and deleted 38k lines...

loganmc10:
The Goldeneye roms look like their reference MD5 points to Perfect Dark, which probably isn't right since they have different save types

Graslu:
It should be right, GoldenEye X is a Perfect Dark ROM hack.

loganmc10:
Ah makes sense, I had no idea

Looks good to me @Narann @richard42

PR #977, opened 2022-11-19
Add support for Advanced Homebrew ROM Header save type

Rosalie241:
This adds support for the save type from the advanced homebrew ROM header (see https://n64brew.dev/wiki/ROM_Header)


Narann:
Thanks, #976 should be merged before this one.

PR #980, opened 2022-12-07
Have netplay just use file extension for file syncing

loganmc10:
I still need to test this, but this is my proposed solution for the issues discussed in https://github.com/mupen64plus/mupen64plus-core/pull/979 with netplay

@Rosalie241 @richard42 I'll let you know when I've tested it and it's ready to merge


loganmc10:
Ok I've tested this @richard42 , works fine, ready to merge

PR #981, opened 2022-12-15
Add SAVETYPE_AUTO

Rosalie241:
This works in the same way as the Project64 option 'Use first used save type', this is handy for ROMs that aren't in the RDB (i.e romhacks, patched ROMs) and will allow those games to be able to save properly.


loganmc10:
Do you have an example of a game that this fixes? I'm not sure I see the difference, doesn't it already default to 4k eeprom if no save type is specified? And there are heuristics in there to detect SRAM or Flash usage as well.

Rosalie241:
Seems like you're right, thank you

PR #983, opened 2022-12-21
Introduce M64CMD_DISK_OPEN & M64CMD_DISK_CLOSE

Rosalie241:
This adds `M64CMD_DISK_OPEN` & `M64CMD_DISK_CLOSE` as a way for a front-end to load 64DD disks without loading a cartridge, the way this works is to use the existing medialoader API:
1) front-end calls `M64CMD_DISK_OPEN`
2) mupen64plus uses the `get_dd_disk` function from the medialoader API to retrieve the filename
3) mupen64plus determines if that file is a valid 64DD disk, if so, return `M64ERR_SUCCESS`
4) the front-end calls `M64CMD_EXECUTE`
5) mupen64plus will use the same  `get_dd_disk` function from the medialoader API to retrieve the disk filename & load the disk

I'm just wondering if this idea (+implementation) sounds good to you (before I write documentation). @richard42


richard42:
It looks good to me; go ahead and write the documentation please.

Rosalie241:
> It looks good to me; go ahead and write the documentation please.

Done 💕

PR #990, opened 2023-01-06
Improve 64DD emulation

LuigiBlood:
This adds two new interrupt handlers for the two different kinds of interrupts the 64DD can issue (MECHA and BM), then introduces delays based on clock speed (and CountPerOp, kinda to make sure), to every command, but especially the Seek command, and then changes the nature of BM interrupts to work with delays instead of working as fast as possible based on DMAs and other hacky reads.

The timing used is very approximate and is trying to be close to real hardware, but is not quite there, a lot more finetuning will be done eventually, but libleo is very tolerant.

This fixes Gnat Attack in Mario Artist Paint Studio prototype, as the game tries to read the next stage data before unloading the current stage, but since the 64DD responds way too quick it replaces the data before it actually unloads it and crashes.

### DIFF
--- a/src/device/dd/dd_controller.c
+++ b/src/device/dd/dd_controller.c
@@ -143,6 +143,57 @@ void dd_bm_int_handler(void* opaque)
     dd_update_bm(dd);
 }
 
+void dd_dv_active(void* opaque)
+{
+    struct dd_controller* dd = (struct dd_controller*)opaque;
+    /* make motor active and prep standby */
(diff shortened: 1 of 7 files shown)


LuigiBlood:
I'll implement disk motor management next as it seems I did a bit of a regression in the process about Gnat Attack which just crashes the emulator again. This would fix it in a definitive manner.

LuigiBlood:
The only change left for this is how many ticks is truly a second on mupen so I can properly manage that, perhaps..?

richard42:
are you ready to merge this or is it WIP?

LuigiBlood:
Yes this is usable I believe.

PR #992, opened 2023-01-24
Fix screenshots when compiling with OSD=0

Rosalie241:
When compiling with `OSD=0`, `OnScreenDisplay` is `True` in the configuration file and `video_plugin_render_callback` is called with `0`, then taking a screenshot will fail.

This patch fixes that issue, I do wonder if I should extend this patch to guard every OSD function call with a `#ifdef M64P_OSD` to prevent it having to access the `OnScreenDisplay` option in the configuration file?

### DIFF
--- a/src/main/main.c
+++ b/src/main/main.c
@@ -878,25 +878,31 @@ m64p_error main_reset(int do_hard_reset)
 
 static void video_plugin_render_callback(int bScreenRedrawn)
 {
+#ifdef M64P_OSD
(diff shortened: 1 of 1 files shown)


Narann:
I would have manage to remove the use of `bOSD` everywhere so you have no dead code when compiling with `OSD=0`.

Rosalie241:
> I would have manage to remove the use of `bOSD` everywhere so you have no dead code when compiling with `OSD=0`.

Fixed

PR #993, opened 2023-02-02
Sanitize save filename

Rosalie241:
This patch fixes an issue where i.e on windows, certain characters like `:<>"/\|?*` aren't allowed to be used for filenames, causing mupen64plus to be unable to create a save, this issue mainly occurs when a ROM/ROM hack isn't in the database and it uses the internal ROM name for the filename, I could make this patch windows-only mostly but to make the filenames consistent across platforms I think it makes sense to do it for all platforms.


Narann:
LGTM.

PR #1001, opened 2023-04-07
Interpreter accuracy improvements

Rosalie241:
This (currently draft) PR makes the pure interpreter more accurate, allowing [n64-systemtest](https://github.com/lemmy-64/n64-systemtest) to progress much further than current upstream mupen64plus-core, I've mainly used n64-systemtest, the ares source code and the CPU manual as reference to make these changes.

Here's the output from before my changes:
[before_changes.txt](https://github.com/mupen64plus/mupen64plus-core/files/11180366/before_changes.txt)

and here's the output from after my changes:
[after_changes.txt](https://github.com/mupen64plus/mupen64plus-core/files/11180367/after_changes.txt)

I've used the paraLLEl video plugin and paraLLEl RSP plugin to test these changes.

Some of these changes do need some code cleanup and further accuracy improvements could be made, but I'm just posting it was a draft to see what you (the maintainers) think of these changes.


richard42:
looks promising. most of the tests fail but at least they run with your changes. and it looks like you have made a lot of changes to increase accuracy with the pure interpreter.

loganmc10:
Looks good to me. The only thing I'd say is that we should try to maintain parity between the pure and cached interpreter. So for the newly implemented opcodes (break, dmtc0, etc..) there should be something there for the cached interpreter as well. I'd expect we should be able to get the same results from the test ROM via the pure or cached interpreter.

Rosalie241:
> Looks good to me. The only thing I'd say is that we should try to maintain parity between the pure and cached interpreter. So for the newly implemented opcodes (break, dmtc0, etc..) there should be something there for the cached interpreter as well. I'd expect we should be able to get the same results from the test ROM via the pure or cached interpreter.

I agree, but first I want to work more on the FPU tests, then work on the cached interpreter and then I'll cleanup the code and mark this PR as ready to review.

m4xw:
Starting to look good indeed. you have my review on Discord, looking forward to it!
Can start to yeet changes from my branches too!

m4xw:
I'll be also be looking for debugger cb's on break when u are done

src/device/r4300/fpu.h
```diff
@@ -74,33 +85,188 @@ M64P_FPU_INLINE void set_rounding(uint32_t fcr31)
...
+M64P_FPU_INLINE void cvt_s_w(uint32_t* fcr31, const int32_t* source, float* dest)
+{
+    set_rounding(*fcr31);
+
+    fpu_reset_cause(fcr31);
+    fpu_reset_exceptions();
+
     *dest = (float)*source;
+
+    fpu_check_exceptions(fcr31);
```
loganmc10, line 208: I don't think this should be included. I haven't tested, but checking for exceptions like this must create a fairly severe performance penalty.

I think this kind of thing should be wrapped in compile time defines like:
```
#if CHECK_EXCEPTIONS
...
#endif
```

This kind of thing is really only useful for homebrew development. There is no need to turn mupen64plu into Ares. Ares has its strengths, and mupen64plus has its strengths, but worsening the performance of mupen64plus, without providing any benefit to commercial games, is not playing to mupen64plus's strengths.

Rosalie241: > I don't think this should be included. I haven't tested, but checking for exceptions like this must create a fairly severe performance penalty.

I disagree, the performance seems largely fine in games with my ryzen 3600, yes this will have a slight performance impact but I feel that the added accuracy (at the moment setting the correct flag/cause bits) is worth it overall.

> This kind of thing is really only useful for homebrew development. There is no need to turn mupen64plu into Ares. Ares has its strengths, and mupen64plus has its strengths, but worsening the performance of mupen64plus, without providing any benefit to commercial games, is not playing to mupen64plus's strengths.

I disagree, mupen64plus is a nintendo 64 emulator and should mimick the real hardware behavior as much as possible, and i.e other emulators (like Project64) are working towards getting n64-systemtest to pass most tests aswell, so if we don't improve our accuracy, mupen64plus will lack behind the other emulators, which I feel is a bad thing.

loganmc10: Yes, but many mupen64plus users aren't using a ryzen 3600, they are using raspberry pi's. I'm just saying that this additional level of accuracy should be made optional, so as to not affect performance, since it doesn't affect the emulation of commercial games.

Jj0YzL5nvJ: Personally, I prefer that any resource intensive settings can be switched at will. I like the precise emulation, but also the flexibility of being able to use it on any PC designed for an office (low performance PC).
If that's not possible in the immediate future, I'll have to agree with Logan.

richard42: I agree - changes which have significant performance regressions should be removable with compile-time options. Lots of people run M64+ on low-spec hardware.

Rosalie241: Fixed with https://github.com/mupen64plus/mupen64plus-core/pull/1001/commits/e1f425156921321c164e24cda7ba83b16f736938, now there's a `ACCURATE_FPU=1` option in the Makefile, which is disabled by default.

Jj0YzL5nvJ (changes requested):
`$ git apply ../1001.patch`
```
../1001.patch:3496: trailing whitespace.
        cp0_regs[CP0_CONFIG_REG] = (rrt32 & UINT32_C(0x0000000F)) 
../1001.patch:3559: trailing whitespace.
    { 
../1001.patch:3576: trailing whitespace.
    { 
../1001.patch:3577: trailing whitespace.
        (*fcr31) |= FCR31_CMP_BIT; 
../1001.patch:3580: trailing whitespace.
        return; 
warning: squelched 22 whitespace errors
warning: 27 lines add whitespace errors.
```

Rosalie241:
I think this PR is ready for review now, pure and cached interpreter produce identical results with n64-systemtest:

[pure_interpreter.log](https://github.com/mupen64plus/mupen64plus-core/files/11728504/pure_interpreter.log)
[cached_interpreter.log](https://github.com/mupen64plus/mupen64plus-core/files/11728505/cached_interpreter.log)

I've used this n64-systemtest build:

[n64-systemtest.zip](https://github.com/mupen64plus/mupen64plus-core/files/11728522/n64-systemtest.zip)


There is of course room for further accuracy improvements, which I might do in the future, however I feel like this is a good starting point for now.

loganmc10:
I took a quick look through and it looks good to me, one question:

this new latch variable, should it be part of the save state? If the variable is set, and then they create a save state, and load it, latch won't retain it's value pre-savestate

Rosalie241:
That's a good point, I hadn't even thought about save states and yes, I agree it should be added, I'll try to get it done as soon as I'm able to, thank you!

Rosalie241:
ping @richard42, it's ready for review!

PR #1006, opened 2023-06-18
Implement unused RDRAM address range 0x0800000-0x03EFFFFF

Rosalie241:
Fixes an in-game crash in Paper Mario when hitting a specific tree with a hammer.

You can reproduce it with this save state (just hit B on the tree):  

[paper_mario_crash.zip](https://github.com/mupen64plus/mupen64plus-core/files/11782215/paper_mario_crash.zip)

on the N64, `0x0800000-0x03EFFFFF` are mapped, but reads/writes are ignored, see the n64brew documentation: https://n64brew.dev/wiki/RDRAM_Interface#Memory_addressing

### DIFF
--- a/src/device/device.c
+++ b/src/device/device.c
@@ -138,7 +138,7 @@ void init_device(struct device* dev,
         /* clear mappings */
         { 0x00000000, 0xffffffff, M64P_MEM_NOTHING, { NULL, RW(open_bus) } },
         /* memory map */
-        { A(MM_RDRAM_DRAM, dram_size-1), M64P_MEM_RDRAM, { &dev->rdram, RW(rdram_dram) } },
+        { A(MM_RDRAM_DRAM, 0x3efffff), M64P_MEM_RDRAM, { &dev->rdram, RW(rdram_dram) } },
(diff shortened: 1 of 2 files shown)


loganmc10:
I think this is correct, just a couple of notes:

Since you are directing everything to the RDRAM functions, I would just modify the existing line (replace `dram_size-1`):
```
{ A(MM_RDRAM_DRAM, 0x3efffff), M64P_MEM_RDRAM, { &dev->rdram, RW(rdram_dram) } },
```

In the functions themselves, I believe this:
```
if (address <= rdram->dram_size)
```
should be:
```
if (address < rdram->dram_size)
```

For example, if `rdram->dram_size` is 0x8000, then the RDRAM is 0x0 to 0x8000-1, 0x8000 would represent an address outside the valid RDRAM

Rosalie241:
Yeah I wasn't entirely sure about the `<=` or `<` but upon looking at the code again, what you're saying seems right.

Thank you for your feedback, pushed it!

loganmc10:
LGTM

Narann:
Thanks for your work!

PR #1019, opened 2023-07-27
Fix IS64 Truncating Lines

flagrama:
Process new lines in buffer to prevent empty debug messages without losing data.

Current method wipes out everything after a new line if a new line is not placed at the end of the string. The OoT debug build has many of these so this is very annoying when working with the decomp code of it.

No idea if it would be accurate to how the IS64 actually works, but fixes the broken functionality at least.


Rosalie241:
I can confirm this works fine with n64-systemtest too, so LGTM!

PR #1021, opened 2023-07-29
netplay fixes from simple64

loganmc10:
This is an upstreaming of the netplay code in simple64.

Notably:
* rdram_size needs to be set after `netplay_sync_settings`, so that it has the proper value for `disable_extra_mem`
* `main_speedlimiter_toggle` should be disabled when netplay is active
* A new packet type is added, `UDP_RECEIVE_KEY_INFO_GRATUITOUS`, which is when the server sends input data without being asked by the player. 
* Currently, the system checks for desyncs every second. Now, it is every 10 seconds (done in order to reduce network traffic, it's not really important to detect desyncs that quickly)


loganmc10:
cc @fzurita , I believe mupen64plus-ae is the only other frontend that makes use of this currently.

loganmc10:
Ping @richard42 I think this should be ok to merge

PR #1057, opened 2024-01-10
Fix Big-Endian Incompatibilities

Clownacy:
Tested on a Wii U.


Clownacy:
It turns out that the `set_fpr_pointers` code actually supported big-endian in the past, but was broken in 9644b14f96ddddf34c9cdfedc9d11fc61dc8ffab.

PR #1060, opened 2024-01-24
src: fix the new_dynarec with musl

orbea:
This is needed on other platforms too, but may be implicit in some where including it again is harmless.


orbea:
This is POSIX so it should be widely available.

https://pubs.opengroup.org/onlinepubs/000095399/basedefs/sys/types.h.html

PR #1074, opened 2024-05-02
Update minizip

loganmc10:
Minizip no longers builds for me in the latest versions of GCC

```
/__w/simple64/simple64/mupen64plus-core/subprojects/minizip/zip.c:1249:28: error: assignment to 'const long unsigned int *' from incompatible pointer type 'const uint32_t *' {aka 'const unsigned int *'} [-Wincompatible-pointer-types]
 1249 |         zi->ci.pcrc_32_tab = get_crc_table();
```

I've updated minizip using the latest sources from here:
https://github.com/madler/zlib/tree/master/contrib/minizip

### DIFF
--- a/subprojects/minizip/crypt.h
+++ b/subprojects/minizip/crypt.h
@@ -32,12 +32,12 @@
 /***********************************************************************
  * Return the next byte in the pseudo-random sequence
  */
-static int decrypt_byte(unsigned long* pkeys, const unsigned long* pcrc_32_tab)
-{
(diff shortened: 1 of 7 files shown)


Jj0YzL5nvJ:
I think it's worth mentioning the potential vulnerability this fixes: madler/zlib#868
Personally I was waiting for 1.3.2...

Note: MSYS2 is currently failing due to the missing `vulkan-headers` dependencies, I will send an upgrade for **CI/CD** soon...

PR #1077, opened 2024-05-06
CI/CD: Upgrade, generic scripts, independent schedule, etc

Jj0YzL5nvJ:
Note: Later I will make the relevant changes in the other repositories to use these scripts as a prerequisite...

### DIFF
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -796,7 +796,7 @@ targets:
 	@echo "    VULKAN=0       == Disable vulkan support for the default video extension implementation"
 	@echo "    POSTFIX=name   == String added to the name of the the build (default: '')"
 	@echo "  Install Options:"
-	@echo "    PREFIX=path    == install/uninstall prefix (default: /usr/local/)"
+	@echo "    PREFIX=path    == install/uninstall prefix (default: /usr/local)"
 	@echo "    SHAREDIR=path  == path to install shared data files (default: PREFIX/share/mupen64plus)"
 	@echo "    LIBDIR=path    == path to install core library (default: PREFIX/lib)"
 	@echo "    INCDIR=path    == path to install core header files (default: PREFIX/include/mupen64plus)"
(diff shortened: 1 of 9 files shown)


projects/unix/Makefile
```diff
@@ -833,7 +833,7 @@ uninstall:
 	$(RM) "$(DESTDIR)$(SHAREDIR)/mupencheat.txt"
 
 clean:
-	$(RM) -r $(TARGET) $(SONAME) _obj $(OBJDIR) $(SRCDIR)/asm_defines/asm_defines_*
+	$(RM) -r *mupen64plus*.* $(TARGET) $(SONAME) _obj $(OBJDIR) $(SRCDIR)/asm_defines/asm_defines_*.h
```
Rosalie241, line 836: Why was this line added? `$(TARGET)` already contains the name of the produced library,

Jj0YzL5nvJ: I was expecting for this question =)

Blame it all on `$(POSTFIX)`, as `$(TARGET)` derives from `$(POSTFIX)`, `$(TARGET)` does not clear the libraries produced with different values of `$(POSTFIX)`.
Then this code will catch the wrong "target":
https://github.com/mupen64plus/mupen64plus-core/pull/1077/files#diff-c007ce097627a7dd008dbefa3054a533a8a79a4f9e3a3fe2ea7e04228f226f64R85-R86

Rosalie241: Can't you `make clean` multiple times with each `POSTFIX` then?

Jj0YzL5nvJ: - I could add another `make clean` after the `make install` in the build script.
- I don't want to make the build script aware of the existence of `POSTFIX`, it would only complicate things with `rsp-z64` and `rsp-cxd4` repositories.

See the `make clean` of the said repositories:
https://github.com/mupen64plus/mupen64plus-rsp-z64/blob/ad877e263e6ee78839cbd12cdd150d17177079cf/projects/unix/Makefile#L317
https://github.com/mupen64plus/mupen64plus-rsp-cxd4/blob/0a4e30f56033396e3ba47ec0fdd7acea3522362a/projects/unix/Makefile#L335

The problem with the **Makefile** in `mupen64plus-core` goes beyond `POSTFIX`, the name of the library varies between platforms (`mupen64plus` vs `libmupen64plus`) and in Linux **_the extension is versioned_**.

Adding `*mupen64plus*.*` is a nice life hack that doesn't hurt anyone, but it can be easily solved by implementing the first point.

PR #1078, opened 2024-05-08
CI: Remove duplicate line introduced with two different PRs

metalj:
Basically reverts #1075 since this was fully covered with #1077
OCD killing me.


Jj0YzL5nvJ:
I noticed it too... but my nihilism combined with my laziness and my tendency to postpone things to infinity are stronger. Until something starts not letting me sleep...

#1079 covers this too.

PR #1079, opened 2024-05-08
CI/CD: Rearranged some parts of the logic, many details and more...

Jj0YzL5nvJ:
@Rosalie241

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,26 +32,26 @@ jobs:
       - uses: actions/checkout@v4
       - name: Get build dependencies and arrange the environment
         run: |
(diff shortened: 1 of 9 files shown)


richard42:
is it ready?

Jj0YzL5nvJ:
> is it ready?

In theory yes, but better to wait until Monday. If during that period I don't change anything else, you can assume it's ready.
I will try to test this code in my repositories during that period, but I don't have as much free time as I would like.

Jj0YzL5nvJ:
I will make more changes, I accidentally added unnecessary variables that I used in mupen64plus/mupen64plus-video-glide64mk2#131 diagnostic.

Jj0YzL5nvJ:
Now it's ready, unless someone has something else to point out.

PR #1086, opened 2024-07-05
Backport mupen64plus-libretro-nx dynarec fix for OoTMM randomizer

alice-mkh:
So, currently https://ootmm.com/ works with said libretro core but crashes when transitioning between games with upstream m64p. I narrowed it down to this commit: https://github.com/libretro/mupen64plus-libretro-nx/commit/9cbe0fedf18ee13874ac3361ad013c41806fe4c7

With this change I can confirm it switches without crashes:

[Screencast from 2024-07-05 18-26-14.webm](https://github.com/mupen64plus/mupen64plus-core/assets/1510457/b0db0a13-9d51-4df9-ae49-5f02cd9bd0d2)

Disclaimer: I don't know if it makes sense, this is their commit unchanged (except for code style fixes)

### DIFF
--- a/src/device/r4300/new_dynarec/new_dynarec.c
+++ b/src/device/r4300/new_dynarec/new_dynarec.c
@@ -2338,7 +2338,9 @@ static void tlb_speed_hacks()
 u_int verify_dirty(struct ll_entry * head)
 {
   void *source;
-  if((int)head->start>=0xa4000000&&(int)head->start<0xa4001000) {
+  if((int)head->start>=0xa0000000&&(int)head->start<0xa07fffff) {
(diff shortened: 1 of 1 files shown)


m4xw:
Been wanting to upstream that for ages but @Gillou68310 wanted that to work for all interpreter types too and we never verified that

alice-mkh:
Ah, that is fair.  I don't have the knowledge to do that myself, but at least it's documented now, as I previously tried to find it and there was nothing at all ^^

m4xw:
I did it originally for @LuigiBlood 's DD shenanigans

PR #1093, opened 2024-10-21
build: drop SDL1 support + use pkg-config for SDL2

orbea:
* Drops SDL1 support for the unix Makefile since it is unmaintained upstream and using it will result in a degraded experience especially when it receives so little testing. It also makes the SDL section of the Makefile far more complicated than it needs to be.
* Use pkg-config for SDl2. Using sdl2-config hasn't been required for a long time and using pkg-config is preferable for distros.

As a benefit this avoids `which(1)` as a dependency when `PKG_CONFIG` is explicitly set by the user as is the case with some distros.

### DIFF
--- a/projects/unix/Makefile
+++ b/projects/unix/Makefile
@@ -351,16 +351,18 @@ endif
 
 # test for presence of SDL
 ifeq ($(origin SDL_CFLAGS) $(origin SDL_LDLIBS), undefined undefined)
-  SDL_CONFIG = $(CROSS_COMPILE)sdl2-config
-  ifeq ($(shell which $(SDL_CONFIG) 2>/dev/null),)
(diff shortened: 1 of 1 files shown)


Rosalie241:
> Drops SDL1 support for the unix Makefile since it is unmaintained upstream and using it will result in a degraded experience especially when it receives so little testing. It also makes the SDL section of the Makefile far more complicated than it needs to be.

There's quite a bit of code that can be reworked in vidext.c if it's fine to drop SDL1 support completely, I can get to that if @richard42 is fine with dropping it.

On a semi-related note, I want to get SDL3 support in eventually too, so dropping SDL1 would help with that work.

richard42:
yes, let's move forward with this. Please pull out the SDL1 code from the video extension as well.

PR #1103, opened 2024-12-16
Fix incorrect maximum ROM size

### LINKED ISSUE #1101: The 64MB rom size limit should be 4GiB
The rom size limit defined [here](https://github.com/mupen64plus/mupen64plus-core/blob/e170c409fb006aa38fd02031b5eefab6886ec125/src/device/memory/memory.h#L31) should be extended to 4GiB as specified by the [n64brew documentation](https://n64brew.dev/wiki/FAQ)

Rosalie241:
Fixes #1101 

Tested with B3313, a rom hack which makes the ROM 96 MiB.


Jj0YzL5nvJ:
This is much better than the previous attempt #685

Looking at your latest modifications, there seems to be a lot more "_limits_" to change... I can't be of much help in that regard.
[Add this ROM to your tests](https://github.com/PeterLemon/N64/blob/master/Video/BigBuckBunny/BigBuckBunny.N64.7z), in my case it always `soft freeze` (_my ParaLLEl RDP/RSP forks are too old =P_) on the same part... I don't know if it works with Ares or some other emulator. The ideal would be to test it with EverDrive or something similar.

loganmc10:
I would also make sure to check that Paper Mario still works (and that it can save). I remember it did something where it wrote outside the cart address space, but maybe I'm remembering wrong

GhostlyDark:
> I would also make sure to check that Paper Mario still works (and that it can save). I remember it did something where it wrote outside the cart address space, but maybe I'm remembering wrong

Just tested, saving works fine.

GhostlyDark:
> This is much better than the previous attempt #685
> 
> Looking at your latest modifications, there seems to be a lot more "_limits_" to change... I can't be of much help in that regard. [Add this ROM to your tests](https://github.com/PeterLemon/N64/blob/master/Video/BigBuckBunny/BigBuckBunny.N64.7z), in my case it always `soft freeze` (_my ParaLLEl RDP/RSP forks are too old =P_) on the same part... I don't know if it works with Ares or some other emulator. The ideal would be to test it with EverDrive or something similar.

It does work with Angrylion (fine) and Parallel RDP (glitches), but only with Parallel RSP. Other RSP plugins just freeze on boot.

image


image

Jj0YzL5nvJ:
It pass with this build: https://github.com/Rosalie241/RMG/actions/runs/12344405888

But the same build fails with the CI cores: https://github.com/mupen64plus/mupen64plus-core/actions?query=branch%3Afix-romsize

That ROM seems to be extremely sensitive to compiler optimizations...

Rosalie241:
> It pass with this build: https://github.com/Rosalie241/RMG/actions/runs/12344405888
> 
> But the same build fails with the CI cores: https://github.com/mupen64plus/mupen64plus-core/actions?query=branch%3Afix-romsize
> 
> That ROM seems to be extremely sensitive to compiler optimizations...

It plays back fully for you without issues? on Fedora it hangs eventually for me when built locally sadly, though now that you mention compiler optimizations, I've had issues with that before while working on improving FPU accuracy, which is extremely frustrating.

Jj0YzL5nvJ:
> It plays back fully for you without issues?

Only on Windows and Wine with [this configuration...](https://imgur.com/a/A0qPLWC)

Rosalie241:
> > It plays back fully for you without issues?
> 
> Only on Windows and Wine with [this configuration...](https://imgur.com/a/A0qPLWC)

Thank you, I can confirm it plays fully with paraLLEl RSP and angrylion-rdp-plus on Windows

image

I can also confirm saving in Paper Mario works fine still.

Rosalie241:
Thanks to the help of @GhostlyDark, we managed to find a solution to the unreliable behavior, it was a mistake I made.

@Jj0YzL5nvJ can you test it again with the latest changes? It works for GhostlyDark and me fine now.

Jj0YzL5nvJ (approved):
Excellent, runs perfect. Even with my deprecated ParaLLEl RDP/RSP forks =P

fullbrightkay:
This seems to break loading roms in `init_mem_base` on a linux aarch64 device with only 2 GB of ram I assume due to MB_MAX_SIZE_FULL being bumped way up

Rosalie241:
> This seems to break loading roms in `init_mem_base` on a linux aarch64 device with only 2 GB of ram I assume due to MB_MAX_SIZE_FULL being bumped way up

Interesting, I suppose I'll add a compile option called `ACCURATE_ROM_RANGE` so by default mupen64plus-core will still work on low end systems.

Rosalie241:
I'll make changes once these changes have been reviewed, because I'm not sure what the best course of action is at the moment, cc @richard42

Chlorobyte-but-real:
> mapping and reserving the maximum possible ROM size to RAM seems a bit excessive even for machines with more than 8 GiB of RAM

Just because the process asks for 4 GiB of _memory_ doesn't mean that it will suddenly consume that much _RAM_. Virtual memory is silly like that - modern OSes should assign physical RAM only when the virtual memory actually gets accessed.

That being said this _is_ a concern on 32 bit platforms where you can't just reserve the entire address space worth, or possibly OSes that aren't as smart with their memory management

m4xw:
> > mapping and reserving the maximum possible ROM size to RAM seems a bit excessive even for machines with more than 8 GiB of RAM
> 
> Just because the process asks for 4 GiB of _memory_ doesn't mean that it will suddenly consume that much _RAM_. Virtual memory is silly like that - modern OSes should assign physical RAM only when the virtual memory actually gets accessed.
> 
> That being said this _is_ a concern on 32 bit platforms where you can't just reserve the entire address space worth, or possibly OSes that aren't as smart with their memory management

it wont run without kernel overcommit enabled

Rosalie241:
I've updated the patch to use a separate rom pointer which is dynamically allocated in `open_rom`, this should solve the memory requirements of lower end systems, the compressed memory is also smaller now due to that.

src/device/cart/cart_rom.c
richard42, line 1: The "e" here is a bit strange for a bitmask, because it allows the top three bits but forces the next lowest bit to 0. Did you intend to use 0x7fffffff for a 2GB mask?

src/device/cart/cart_rom.h
richard42, line 1: And likewise, this bitmask only allows a 4x bigger ROM space. Did you mean to do 0xfffffffc? or 0x7ffffffc?

src/device/device.c
richard42, line 1: and lastly, the 0xfbfffff here is a bit odd because it caps the size just under 256MB. Is there a missing "f" in here?

richard42 (commented):
everything looks good here but the 3 issues I brought up.

src/device/cart/cart_rom.h
m4xw: Uncached addr space ends at 0xbfffffff  and before that theres 0xbfc80000 which was some mips specific vector stuff so i dont think it physically cant get that high.

src/device/cart/cart_rom.c
m4xw: The 0xe does include the 0xA cache bit for uncached cart dom 2 (0xB00... range) and works with the cached access, I guess thats the idea behind this.

e = b1110
a = b1010 & E = A
b = b1011 & E = A
c = b1100 & E = C

I guess in this case it avoids the kernel-virtual space? Would need to double check where its used

PR #1123, opened 2025-03-09
Fix integer overflow in write_is_viewer

Giles-one:
Modify it to check each number one by one and whether they sum to overflow.

### DIFF
--- a/src/device/cart/is_viewer.c
+++ b/src/device/cart/is_viewer.c
@@ -55,13 +55,15 @@ void write_is_viewer(void* opaque, uint32_t address, uint32_t value, uint32_t ma
     {
         if (word > 0)
         {
-            /* make sure we don't overflow the buffer */
-            if (is_viewer->buffer_pos + word > IS_BUFFER_SIZE)
+            /* make sure we don't overflow the integer or the buffer  */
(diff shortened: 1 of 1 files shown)


src/device/cart/is_viewer.c
```diff
@@ -55,13 +55,15 @@ void write_is_viewer(void* opaque, uint32_t address, uint32_t value, uint32_t ma
...
+            /* make sure we don't overflow the integer or the buffer  */
+            if (is_viewer->buffer_pos > IS_BUFFER_SIZE 
+                || word > IS_BUFFER_SIZE
+                || is_viewer->buffer_pos + word > IS_BUFFER_SIZE )
             {
                 /* reset buffer */
                 memset(is_viewer->output_buffer, 0, IS_BUFFER_SIZE);
                 is_viewer->buffer_pos = 0;
-                DebugMessage(M64MSG_WARNING, "IS64: prevented buffer overflow, cleared buffer");
+                DebugMessage(M64MSG_WARNING, "IS64: prevented integer overflow, cleared buffer");
```
Rosalie241, line 66: the only nitpick I have is the space between `IS_BUFFER_SIZE` and `)` and the fact you changed this to `integer overflow`, I think the previous log line was better and fits the situation better.

PR #1133, opened 2025-04-23
Only allocate UDP packets once

Rosalie241:
I looked at `netplay.c` and think that allocating UDP packets constantly during emulation is very wasteful and rather expensive.

So this patch makes the netplay code re-use the UDP packets, this is allowed in SDL_net:

> https://wiki.libsdl.org/SDL2_net/SDLNet_AllocPacket
> When done with this packet, you can free it with [SDLNet_FreePacket](https://wiki.libsdl.org/SDL2_net/SDLNet_FreePacket). Packets can be reused multiple times; you don't have to allocate a new one for each piece of data you intend to send.

This should hopefully save a few CPU cycles when people are using mupen64plus with netplay.

### DIFF
--- a/src/main/netplay.c
+++ b/src/main/netplay.c
@@ -50,6 +50,13 @@ static uint8_t l_plugin[4];
 static uint8_t l_buffer_target;
 static uint8_t l_player_lag[4];
 
+//UDP packets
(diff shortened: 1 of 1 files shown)


Rosalie241:
The CI failure is unrelated to any code.

PR #1154, opened 2025-09-26
Fix transfer pak functionality

Rosalie241:
This fixes the GameBoy tower functionality in Pokemon Stadium 2 when using a LLE RSP plugin like static interpreter or parallel-rsp:

Screenshot_20250926_224139

This patch also requires https://github.com/mupen64plus/mupen64plus-core/pull/1153

This patch makes the code match ares's implementation: https://github.com/ares-emulator/ares/blob/master/ares/n64/controller/gamepad/transfer-pak.cpp

### DIFF
--- a/src/device/controllers/paks/transferpak.c
+++ b/src/device/controllers/paks/transferpak.c
@@ -35,9 +35,12 @@
 #include <string.h>
 
 
+#define RESET_MODE_CART_ENABLE_BIT UINT32_C(0x00000001)
+#define RESET_MODE_PAK_ENABLE_BIT UINT32_C(0x00000080)
+
 static uint16_t gb_cart_address(unsigned int bank, uint16_t address)
 {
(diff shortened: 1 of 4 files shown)


Jj0YzL5nvJ:
Does this only fix the original versions or does it also fix the European versions and enhanced editions that use RTC?

In any case, there are surprisingly few changes to match ares's implementation in regard to emulating Game Boy cartridges.

Rosalie241:
> Does this only fix the original versions or does it also fix the European versions and enhanced editions that use RTC?

I haven't tried the European or enhanced editions, you can try with the latest RMG version though and report back if you'd like.

> In any case, there are surprisingly few changes to match ares's implementation in regard to emulating Game Boy cartridges.
yeah, it was pretty complete already

Rosalie241:
I've just tried the European version of Pokemon Stadium 2 and it works fine aswell:

image

PR #1180, opened 2026-06-22
CI/CD: Update 2026.6

Jj0YzL5nvJ:


.github/workflows/scripts/ci_get_deprecated_msys2_mingw_deps.sh
Rosalie241, line 1: I'd not rely on mirrors like this, it's extremely fragile.

Rosalie241 (changes requested):

.github/workflows/scripts/ci_get_deprecated_msys2_mingw_deps.sh
Jj0YzL5nvJ: The only thing that matters is that they're not as fragile as what's left of `MSYS2 MINGW32`.

MSYS2 is slowly killing MINGW32 repos with usage statistics as justification. I don't want to go so far as to create repositories with old packages like `MSYS2 CLANG32` as other people have done... or my own run images.

Rosalie241: > MSYS2 is slowly killing MINGW32 repos with usage statistics as justification.

Maybe it's best to drop those CI builds then, would ease up maintenance and prevent having to do further hacks to get removed msys2 mingw32 packages.

PR #1188, opened 2026-09-05
debugger/dbg_memory: fix building with newer binutils

Tatsh:
binutils switched to using C99 bool at some point and they were transitively including `bfd.h` which defined `TRUE`/`FALSE` as `1`/`0` respectively.

The rest of the file uses `0` and `1` everywhere.


Rosalie241:
Thank you!

